Issue #1, opened 2015-08-24
Create lobid-resources repo fo API 2.0 and show example of new json ld transformation


dr0i:
The API 2.0 hbz catalog transformation will be considerably different than the old one used in 'https://github.com/lobid/lodmill'.
For a start:

- show how to transform N-Triples to compacted and framed JSON-LD
- the outcome should be indexed into elasticsearch
- write a short documentation

acka47:
See also https://github.com/hbz/lobid/issues/172.

acka47:
This may overlap with https://github.com/lobid/lodmill/issues/449.

acka47:
Can't we close this one?

acka47:
Forget my last comment. Just saw that #2 isn't merged. The repo already exists though...

dr0i:
Yeah, no we can't close as the presumptions of the issue were quite bold (see above) and are not fullfilled yet.

acka47:
Thanks for adjusting the title accordingly.

acka47:
For a documentation of test usage of JSON-LD framing see https://github.com/lobid/lodmill/issues/433.

acka47:
The JSON-LD frame is quite easy. This suffices:

```javascript
{
  "@context": "http://lobid.org/context/lobid-resources.json",
  "@type": "http://purl.org/dc/terms/BibliographicResource"
}
```

Unfortunately, the arrays for creators/contributors and subjects aren't sorted according to the source data as we would like them to. For this, we will need to 
1. add the order in N-Triples as RDF list or
2. discard this approach and move on like we intended to first, i.e. directly creating JSON-LD from source data with Metafacture.

jschnasse:
For Api 2.0 , I´d suggest another trick. It would be nice to always get a prefLabel as a side car to all occurrences of @id (in document AND in context) . The usage of prefLabel would give clients a simple, unified and intuitive way to access the index and to mock views on lobid data. It can also be useful for configuring the index (e.g. boosting prefLabel). 
Some examples:
- in case of gnd-ids the preferredName should be provided as prefLabel, same for gnd-subjects
- for skos vocabs, a prefLabel comes together with the uris
- for the @ids in the context you can use tools like [etikett](https://github.com/hbz/etikett)

acka47:
Adding to my last comment. The result of the framing has to be post-processed:

1. the `@graph` array must be removed
2. the blank node `@id`s for the multiVolumeWork/series relation have to be removed.

[...]

PR #2, opened 2015-08-27
Issue 1 create repo and setup initial test

dr0i:
See #1.


fsteeg:
A few points I noticed:

- Formatting in README: two bullet points don't render (should be `*` not `-` for .textile)
- No formatter setup yet, I'd suggest we use our current settings from the lobid repo
- To assess progress towards our goals from #1, it'd be nice to see the resulting JSON in the ES index (e.g. by printing it during the test run, or indexing it into a real ES at localhost and providing a sample query in the README)

fsteeg:
The package name is still the old `org.lobid.lodmill`, should we use `org.lobid.resources`?

You did not adress my last point from my previous comment, right? Or did I miss something?

(To assess progress towards our goals from #1, it'd be nice to see the resulting JSON in the ES index, e.g. by printing it during the test run, or indexing it into a real ES at localhost and providing a sample query in the README)

Maybe the best would be to have some code that indexes the data into a common ES instance (like we do in the lobid-organisations repo), so we have a common result that we can look at and discuss, also with @acka47.

In general, I'd suggest we use this pull request to refine and discuss our format and approach, until we're happy. Since this is the first commit, there is no master that we diverge from, so I see no harm in a long running pull request in this case.

fsteeg:
For consistency with the `lobid-organisations` repo, I'd suggest we:

* use `resources` (plural) as the index name
* use `resource` (singular) as the ES type

This would allow a simple, informative GET request as for organisations:

`curl "http://weywot2.hbz-nrw.de:9200/organisations/organisation/DE-605/_source" | json_pp`

Or, with a JSON plugin in the browser, we can just visit:

http://weywot2.hbz-nrw.de:9200/organisations/organisation/DE-605/_source

The anologue for resources would then look like this:

`curl "http://weywot2.hbz-nrw.de:9200/resources/resource/HT018295975/_source" | json_pp`

Or in the browser:

http://weywot2.hbz-nrw.de:9200/resources/resource/HT018295975/_source

BTW if you wrap the curl command in `@` signs, it gets code formatting in the README.

fsteeg:
Looking at [this resource](http://weywot2.hbz-nrw.de:9200/resources/resource/HT018295975/_source), I think these would be the next steps to make it something like [this organisation](http://weywot2.hbz-nrw.de:9200/organisations/organisation/DE-605/_source):

* `compact` the JSON before storing it in ES (short keys, but `compactArrays=false`)
* remove the `@graph` structure, as we discussed today
* nested structure through framing (the current data is not framed, right?)

dr0i:
Yes, these are next steps. There is [already an issue](https://github.com/lobid/lodmill/issues/449) for it. Let's not pack too much in this ticket.

PR #11, opened 2016-01-28
Issue 1 create repo and setup initial test

dr0i:
This should be part of #2. Copying files from the https://github.com/lobid/lodmill repo and traying to preserve the history of these commits messed everything up. It's not good to do this with git. So I  ended up by creating a new pull request with a partly coherent commit history.


src/main/java/org/lobid/resources/DirReader.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Reads a directory and emits all filenames found.
+ * 
+ * @author Markus Michael Geipel, Pascal Christoph
+ */
+@In(String.class)
+@Out(String.class)
+@Description("Reads a directory and emits all filenames found.")
+public final class DirReader
```
fsteeg: We have `DirReader` in metafacture-core:
https://github.com/hbz/metafacture-core/blob/master/src/main/java/org/culturegraph/mf/stream/source/DirReader.java

src/main/java/org/lobid/resources/MabXmlHandler.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * A MAB XML reader.
+ * 
+ * @author Pascal Christoph (dr0i)
+ * 
+ */
+@Description("A MAB XML reader")
+@In(XmlReceiver.class)
+@Out(StreamReceiver.class)
+public final class MabXmlHandler extends DefaultXmlPipe<StreamReceiver> {
```
fsteeg: We have `AlephMabXmlHandler` in metafacture-core:
https://github.com/hbz/metafacture-core/blob/master/src/main/java/org/culturegraph/mf/stream/converter/xml/AlephMabXmlHandler.java

dr0i: Uh - good catch! Fixed that.

src/main/java/org/lobid/resources/DirReader.java
dr0i: Yes, but they are not the same.

dr0i: ... and no more used anyway, so I removed it. Thx for the pointer!

fsteeg:
What are the XSD files in `src/main/resources/schemata/` about?

They seem to be copies from metafacture-core.

dr0i:
They are needed to populate the new metamorph function `UrlEscaper` to the metafacture framework when executing the morph script.

src/main/resources/schemata/metamorph.xsd
```diff
@@ -0,0 +1,927 @@
...
+          <restriction base="int">
+            <minInclusive value="0"/>
+          </restriction>
+        </simpleType>
+      </attribute>
+    </complexType>
+  </element>
+  <element name="urlescape">
+    <annotation>
+      <documentation>URL escaping according to the guava 15.0 UrlEscapers.urlFragmentEscaper library.
```
fsteeg: Thanks for explaining the purpose of the XSDs. It seems the documentation above is out of date.

In general this adds a lot of duplication for something that looks like it's essentially a one-line bug fix of URL encoding.

Does the code in your `UrlEscaper` basically do the same thing as `URLEncode` in metafacture-core, but better, i.e. handling a special case we need correctly? If it also handles general URL encoding correctly, maybe we should replace the `URLEncode` implementation in our metafacture-core fork with your implementation.

dr0i: Yes, it does things better. But it comes with a google library won't have a chance make it into metafacture-core because of the strict policy there. Even if it'll make it into the metafacture-plugins repo it should have a name on its own, because you never know which one of both commands would be executed if they share the same name, so programms will become non-deterministic.

fsteeg:
You're right about the lib and metafacture-core, but that should not keep us from taking the simple approach now. We can still see how we'll get it into metafacture-core (maybe based on a different, already included lib, etc). And in the meantime, we avoid the duplication and two very similar functions in the system we actually work with.

dr0i:
The simple approach would be leave it as it is. Do nothing. Besides, when integrating in our fork and overwriting a core function, and update of that fork would only be made complicater. So I am with you: do the simple approach, leave it separated.

fsteeg:
The *easy* way is to do nothing. *Simple* is to have one way to encode a URL. :-)

fsteeg:
After some offline conversation I understood the difference between the two functions and that it makes sense at this point to keep them both. Since the `urlencode` function is specific to this repo it also makes sense to keep it here, and not in our fork, in spite of the duplicate XSDs (which need to be updated here if updated in metafacture-core). One issue remains then: the documentation update.

dr0i:
Updated the docu in the xsd.

PR #24, opened 2016-02-12
Add test script as helper to build test resources

dr0i:


jsonDiffAndCopy.sh
```diff
@@ -0,0 +1,11 @@
+#/bin/bash
+# This diffs the jsonld documents using the the cli and copy the jsonld
```
fsteeg: Small typo here (`the` twice).

fsteeg:
I don't quite understand in what context this is used. The comment suggest it's for running JUnit tests. In that case, shouldn't it be part of the test code? Is that what you mean with the TODO in the comment ('TODO Should be changed in the java sources, I guess.')?

dr0i:
It's about rebuilding the test data when new transformation rules changes the data. This could (and should) be done in java, I will think about how to do it there.

dr0i:
If you like implement it with java, would be nice.

acka47:
@fsteeg You have this on the radar? Just asking because this is a lonely PR in the Ready column...

dr0i:
Closed this in favor of #58.

Issue #30, opened 2016-02-15
Fix internal_id to be hbz01 001, not ZDB*


dr0i:
We want to http-get IDs directly, like:

http://gaia.hbz-nrw.de:9200/resources/resource/HT001387709

The resource exists, but has a wrong elasticsearch internal_id, taken form a `sameAs` entry. See http://gaia.hbz-nrw.de:9200/resources/resource/_search?q=HT001387709.

fsteeg:
I've deployed usage of the new index in nwbib to staging.

Here's the test case: http://test.lobid.org/nwbib/HT001387709

dr0i:
Deployed this one test resource into the full index at gaia, see  http://test.lobid.org/nwbib/HT001387709.

fsteeg:
+1

Opened new issue https://github.com/hbz/nwbib/issues/253 (blocks https://github.com/hbz/nwbib/pull/252)

fsteeg:
https://github.com/hbz/lobid-resources/pull/32 is merged and deployed, closing.

See http://test.lobid.org/nwbib/HT001387709 (base feature is not on production yet).

PR #32, opened 2016-02-17
Fix getting internal_id by using improved regex

dr0i:
The regex now disallows ZDB-URIs and URIS ending with "about" to be
internal_ids.

See #30.


fsteeg:
Already deployed to staging index, just merging. See: http://test.lobid.org/nwbib/HT001387709

PR #42, opened 2016-02-26
Make context.json-URL configurable

dr0i:
See hbz/lobid-rdf-to-json#29.

The script building the small test set will use the staging context.
The default context is the "productive" one. That guarantees all our
crontabed scripts will work as expected without adapting them.


dr0i:
Merge after #41.

fsteeg:
+1

PR #45, opened 2016-03-01
Use jitpack to handle lobid-rdf-to-json library

dr0i:
- add jitpack repository to pom
- add lobid-rdf-to-json dependency to pom
- remove git building lobid-rdf-to-json from script

Resolves hbz/lobid-rdf-to-json#32.


fsteeg:
I think @jschnasse would be a better reviewer for this, reassigning.

jschnasse:
+1

PR #53, opened 2016-03-04
Lobid rdf to json 10 change uri schema

dr0i:
Change URIs (description & thing) of the resources

Resolves hbz/lobid-rdf-to-json#10.

Merge after #51.


fsteeg:
+1

PR #55, opened 2016-03-04
Change index name for small test

dr0i:
This way the "small test index" can't get into the way of the productive index.


fsteeg:
+1

PR #58, opened 2016-03-09
24 update test resources

dr0i:
Substitutes #24.
No need for temporary files. No need for scripts. Diffing is done using git.


fsteeg:
+1

PR #61, opened 2016-03-09
Add "git pull" to start script

dr0i:
Sometimes the pulling is done in the crontab commands, sometimes
not. It is the best to always pull explicitly in the script.

### DIFF
--- a/startHbz01ToLobidResources.sh
+++ b/startHbz01ToLobidResources.sh
@@ -40,6 +40,7 @@ fi
 
 function indexFile() {
 	git checkout $BRANCH
+	git pull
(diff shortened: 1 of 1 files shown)


startHbz01ToLobidResources.sh
```diff
@@ -40,6 +40,7 @@ fi
 
 function indexFile() {
 	git checkout $BRANCH
+	git pull
```
fsteeg: As the other day, I'd suggest `git pull origin $BRANCH` to avoid different behaviour in different environments.

fsteeg:
+1

PR #71, opened 2016-04-20
Adjust URIs for RDA elements

### LINKED ISSUE #60: Adjust context and morph to use new namespace for RDA properties
Currently, we have seven properties from the namespace `http://rdvocab.info/Elements` in our data. These are deprecated since the new namespace `http://rdaregistry.info/Elements/u/` is used. Also, the property names differ. E.g. http://rdvocab.info/Elements/frequency now is http://rdaregistry.info/Elements/u/P60538. (Obviously, the proponents of non-natural property names have won. :sob: )

We will have to update

- [x] the morph
- [ ] the test files
- [ ] the context(s)

for these seven properties.

acka47:
Fixes #60.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
 		<dependency>
 			<groupId>com.github.hbz</groupId>
 				<artifactId>lobid-rdf-to-json</artifactId>
-				<version>f0d6b368d093fe43537ebcfb540ec7dfef6dde87</version>
+				<version>ba0bb2f47cdfa470e03cf344ab2db2c24906b3ef</version>
 		</dependency>
 		<dependency>
 			<groupId>org.culturegraph</groupId>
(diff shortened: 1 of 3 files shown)


fsteeg:
Deployed to small test index, see e.g.:
http://lobid.org/resources-small?q=otherTitleInformation:*

However, as we noticed when working on this today, there is no change in the resulting JSON.

Only in the context, which now maps the JSON-LD terms to the new properties:
http://lobid.org/download/contextTmp-staging.json

This shows how the redundant intermediate step of creating n-triples causes needless overhead.

acka47:
+1

> This shows how the redundant intermediate step of creating n-triples causes needless overhead.

In this case where we only replace property URIs this is correct. As this is a rare case I don't think this is a problem.

Issue #33, opened 2016-02-17
Map field 333 to rdau:P60327


acka47:
Requested by edoweb stakeholder, example [HT014812594](http://lobid.org/resource/HT014812594).

[Source](http://lobid.org/resource?id=HT014812594&format=source) (snippet):

```xml
<datafield tag="333" ind1="-" ind2="1">
  <subfield code="a">Statistisches Landesamt Rheinland-Pfalz</subfield>
</datafield>
```

JSON:

```javascript
  "otherTitleInformation": [
    "Statistisches Landesamt Rheinland-Pfalz"
  ]
```

I am not sure whether we should do this. In other entries it seems to introduce some redundancy with field 335, though. Example [HT018166285](http://lobid.org/resource/HT018166285):

[Source](http://lobid.org/resource?id=HT018166285&format=source) (snippet):

```xml
<datafield tag="333" ind1="-" ind2="1">
  <subfield code="a">Euromedica Hannover 2010</subfield>
</datafield>
<datafield tag="335" ind1="-" ind2="1">
  <subfield code="a">Euromedica Hannover 2010 ; 3. - 4. Juni 2010 ; Internationaler medizinischer Kongress</subfield>
</datafield>
```

JSON:

```javascript
  "otherTitleInformation": [
    "Euromedica Hannover 2010 ; 3. - 4. Juni 2010 ; Internationaler medizinischer Kongress",
    "Euromedica Hannover 2010"
  ],
```

From the [MAB documentation]():

```
333       ZU ERGAENZENDE URHEBER ZUM HAUPTSACHTITEL

          Indikator:
          blank = nicht definiert


335       ZUSAETZE ZUM HAUPTSACHTITEL

          Indikator:
          blank = keine Nebeneintragung
          a     = zusaetzliche Nebeneintragung unter dem Zusatz
          b     = zusaetzliche Nebeneintragung mit dem Zusatz
```

/cc @aquast

acka47:
@aquast and me just discussed this and we decided not to remove the field but to map it to a separate property. I think the right one is http://rdaregistry.info/Elements/u/P60327 described as "a statement associated with the title proper of a resource that relates to the identification and/or function of any agents responsible for the creation of, or contributing to the realization of, the intellectual or artistic content of the resource."

acka47:
@jschnasse Is it ok when this is only implemented for data 2.0 or do you need it in the production data?

acka47:
[...]

PR #75, opened 2016-04-26
Use adjusted version of lobid-rdf-to-json. (#33).

acka47:
See https://github.com/hbz/lobid-rdf-to-json/pull/38.


acka47:
The tests won't pass. @jschnasse will do the necessary adjustment to make them pass. After that, I will try to adjust the morph as written in https://github.com/hbz/lobid-resources/issues/33#issuecomment-214718155

fsteeg:
Update code for changes in lobid-rdf-to-json, build now succeeds. Reassigning to @acka47.

PR #83, opened 2016-05-31
Dump surplus in fulltextOnline fields

### LINKED ISSUE #69: Resource HT017411546 not found
Originally reported by @acka47 in https://github.com/hbz/nwbib/issues/308:

```
$ curl -i http://lobid.org/resources/HT017411546
HTTP/1.0 404 Not Found
```

dr0i:
Field 655e1u should contain URIs but sometimes contain also
whitespaces followed by some notes. Thus the content is no URI
but a string which contradicts the index config of ES and crashes
indexing the document.

- dump possible suffix which begins with a whitespace
- add test resource HT017411546
- update tests

Resolves #69.


fsteeg:
+1

Issue #85, opened 2016-06-14
Ensure URIs in id field


dr0i:
There are still some (around 4k) documents dropped when indexing them into elasticsearch because they don't fulfill the expectations (being an ID or an String).
Examples for `org.elasticsearch.index.mapper.MapperParsingException`:

object mapping | value
------------ | ------------- 
sameAs|http://worldcat.org/oclc/ocm00895971^
similar|http://dx.doi.org/10.4128/ 9781606493311
seeAlso|ftp://ppftpuser:welcome@ftp01.penguingroup.com/Booksellers
tableOfContents|www.folkwang-uni.de/fileadmin/medien/Downloads/Bibliothek/Inhaltsverzeichnisse/HT018120416.pdf

The fixes for #80 , #69 are not abstract enough to cover also those field normalizations.
Question is:
What to do?
Could write e.g. a metafacture URI validator, or a URI fixer which tries to fix a URI. The latter would mean to program algorithms which could fix some URIs , but not all. Also possible would be to have just a validator and simply reject field which are not correctly catalogued.

dr0i:
The indexing with the current URI enforcer fixes 90% of the lost records. Still 500 documents are not being indexed, thus there is room of imrovement (also see the helpful comments from @fsteeg ). Will switch the index to production (resulting in +450 documents) while working on this ticket.

dr0i:
@acka47 please also have a look at Q2 in https://github.com/hbz/lobid-resources/pull/87#issuecomment-230430184 . The mentioned "empty strings" are produced when URLs are recognized as broken, see e.g. the `urn` in field `similar` of http://lobid.org/resource/HT004426370. The entry will be ignored, see http://gaia.hbz-nrw.de:9200/resources/resource/HT004426370.

acka47:
I agree with @dr0i to drop the triples in question by now and provide the Verbundgruppe with statistics of incorrect URL input. Maybe they will 

1. fix the entries and/or
2. add a validation to field 655 u

If they react we can provide a list of affected resources.

dr0i:
Deployed to production. With this, the API2.0 data are indexed in the ES version 2.3.3 at quaoar1.

PR #87, opened 2016-06-16
85 ensure uri is in id field

dr0i:
See #85.


fsteeg:
When we discussed the general approach of trying to fix the URL, I had in mind something like we do in lobid-organisations: https://github.com/hbz/lobid-organisations/blob/master/app/transformation/morph-enriched.xml#L19

A dedicated morph function might actually be a good idea, but I find the current implementation hard to follow. For instance, I don't understand the "no scheme" regex. Would it be an option to check what `URI#getScheme()` returns instead (or `URI#isAbsolute()` which you use further below)? And what's the reason for the `enforcedUri.split("%20")[0]` as the final attempt to get a valid URI?

Maybe the URL validation routine from Commons-Validator would make sense here: https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/UrlValidator.html

On replacing spaces with %20: in lobid-organisations the issue was not missing escapes, but pointless, wrong spaces, thus the removal of the spaces in the morph linked above. Don't know how that is for the URLs here.

The dummy URL seems odd to me. That will end up in the data, right? I think I don't really understand why we have to have a valid URL here (I remember you said docs were skipped, why is that again?). I guess that is also the main difference to the lobid-organisations implementation: there, we try to fix the URL, but if that does not work, it simply remains an invalid URL (which to me seems to make more sense than replacing it with a dummy URL).

dr0i:
>like we do in lobid-organisations

This is 20m records in lobdi-resources against 18k in organisations. You can't imagine how creative cataloguers sometimes (say, every 10.000th) are. I have seen  _things_ ...

> I don't understand the "no scheme" regex. Would it be an option to check what URI#getScheme() returns instead (or URI#isAbsolute() which you use further below)

Yes, this seems to be a good idea. I will try this.

> Maybe the URL validation routine from Commons-Validator would make sense here:

I don't think so. Mind that's about a URI not just a URL. There are tons of allowed schemes.

What I have in mind is to steadily improve the function by checking it agains the real data.

>...  pointless, wrong spaces, thus the removal of the spaces 

We have URLs with annotations at them. These could the be splitted an thus a valid URL can be gained. Have a look at the `hasVersion` example in the mother ticket, I just added that.

>  I guess that is also the main difference to the lobid-organisations implementation: there, we try to fix the URL, but if that does not work, it simply remains an invalid URL (which to me seems to make more sense than replacing it with a dummy URL).

Have a look in the mother ticket where it is said 
> they don't fulfill the expectations (being an ID [...]

What you do in lobid-organisations is to feign URIs where no URIs are. In lobid-resources this is not possible, because the data is first serialized as ntriples, where URIs are URIs and strings are strings - the latter indicated by typing the value with `^^<http://www.w3.org/2001/XMLSchema#string>`. These triples are then serialised as json, ensuring that values of type string don't fall into `id` fields.
( Hm , btw, I wonder what'd  happen if you serve the lobid-organisation json-ld in another RDF serialization? Would the json field be enforced as a URI triple and thus throwing an exception? Then you would have one document as json-ld WITH a field containing a "URI" , whilest in other serializations this triple would loss (That loss in persistence should be avoided)).
The dummy URI could be avoided, e.g. the returned value could be empty, which could be checked in the morph and then be ignored. This seems to be a good idea.

fsteeg:
> I don't think so. Mind that's about a URI not just a URL. There are tons of allowed schemes.

The validator supports custom schemes or allowing any scheme, see: https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/UrlValidator.html

> I wonder what'd happen if you serve the lobid-organisation json-ld in another RDF serialization?

Conversion to Turtle was actually what pointed me to the invalid URIs, see https://github.com/hbz/lobid-organisations/issues/146

dr0i:
Now switched to sanitizing URLs instead of the more abstract URI. If URIs msut be sanitized, like DOIs, this is done via morph regex.

src/main/java/org/lobid/resources/UrlSanitizer.java
```diff
@@ -0,0 +1,70 @@
...
+	private static String sanitizeUrl(final String value) {
+		String url = value.trim();
+		// unwise characters (rfc2396) :
+		url = url.replace("\\", "%5C").replace("|", "%7C");
+		if (url.matches(".*#.*#.*")) {// allow only one fragment
+			url = url.substring(0, (url.indexOf("#", url.indexOf("#") + 1)));
+		}
+		url = url.replace(" ", "%20");// space in URI
+		if (!urlValidator.isValid(url)) {
+			if (!urlValidator.isValid(url)) {
```
fsteeg: I think this is an accidental duplicate check (checks the same thing twice, no modification in between).

src/main/java/org/lobid/resources/UrlSanitizer.java
```diff
@@ -0,0 +1,70 @@
...
+				for (String urlSplitter : url.split("%20")) {
+					if (urlValidator.isValid(urlSplitter))
+						return urlSplitter;
+				}
+				if (!urlValidator.isValid(url)) {
```
fsteeg: This seems redundant too, since `url` is not assigned since the last identical check above.

fsteeg:
Added two inline comments about `sanitizeUrl` implementation details above.

To make sure I understand the actual issue: is the problem that during the automatic conversion of N-Triples to JSON-LD, certain fields are sometimes represented as a JSON object (triggered by N-Triple URLs), and sometimes as strings (non-URLs)?

What I don't understand is how returning an empty string fixes that. Why can't we have the invalid URL instead? That would seem better than losing the field content, in particular for fixing the content based on our data (which seems impossible if the invalid URL is gone).

src/main/java/org/lobid/resources/UrlSanitizer.java
dr0i: Good catch!

src/main/java/org/lobid/resources/UrlSanitizer.java
dr0i: Another one!

dr0i:
To Q1: yes. That _was_ the original position. Changed that with https://github.com/hbz/lobid-resources/pull/87/commits/44e6189a5749c3d6deb4bffd3bc6c31d5e027f1c .

To Q2: Triples with empty strings will be ignored, so they wouldn't have lead to the ES 
MapperParsingException. Only with the forementioned commit one _could_ use the broken URL. However, I would advise against it, since it's just not correct by definition and would lead RDF-parser to ignore the triple anyway (in best cases, when not crashing).
How to fix the content? Easy: go to the original source of the data (we have that!). Or: actively scan the logs (https://github.com/hbz/lobid-resources/blob/44e6189a5749c3d6deb4bffd3bc6c31d5e027f1c/src/main/java/org/lobid/resources/PipeEncodeTriples.java#L137). I would also claim that less people would even be aware of the need of the fixing of data if it's (incorrectly) there. They would say: "I see the data. All is good". It's what bring them to make errors in the first place: cataloging and seeing the data sink into the database. The data shouldn't ! Then "uh!" - and data adaption.

dr0i:
Some stats:
`grep MapperParsingException  lobid-gaia.log.2016-07-02 | wc -l`: 168
mostly caused of being `urn` beeing an `id` (while expecting a value, see #88)  and `isPartOf` selfreferencing problems (see #89).

`grep  "No absolute URI could be generated from" 20160702-083002-85-ensureURIsInIdField.startHbz01ToLobidResources.sh | wc -l`:1540

fsteeg:
> Triples with empty strings will be ignored, so they wouldn't have lead to the ES
MapperParsingException.

Ah, I see, thanks.

> Only with the forementioned commit one could use the broken URL.

OK, so we can simply choose what we actually want.

I think this should be part of the functional review of #85 by @acka47.

acka47:
I agree with @dr0i to drop the triples in question by now and provide the Verbundgruppe with statistics of the wrong field inputs. Maybe they will 

1. fix the entries and/or
2. add a validation to field 655 u

If they react we can provide a list of affected resources.

dr0i:
Functional review done.

fsteeg:
+1

PR #90, opened 2016-07-06
Hbz lobid 293 repair items endpoint

dr0i:
See hbz/lobid#293.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,9 +43,19 @@
 			<version>1.47.1</version>
 		</dependency>
 		<dependency>
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId>
+			<version>18.0</version>
+		</dependency>
+		<dependency>
 			<groupId>org.elasticsearch</groupId>
 			<artifactId>elasticsearch</artifactId>
-			<version>1.3.2</version>
+			<version>2.3.3</version>
+		</dependency>
+		<dependency>
+			<groupId>com.sun.jna</groupId>
+			<artifactId>jna</artifactId>
+			<version>3.0.9</version>
 		</dependency>
 		<dependency>
 		  <groupId>com.ning</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -41,11 +41,12 @@
 			<groupId>com.google.gdata</groupId>
 			<artifactId>core</artifactId>
 			<version>1.47.1</version>
-		</dependency>
-		<dependency>
-			<groupId>com.google.guava</groupId>
-			<artifactId>guava</artifactId>
-			<version>18.0</version>
+			<exclusions>
+				<exclusion>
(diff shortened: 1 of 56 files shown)


src/main/java/org/lobid/resources/ElasticsearchIndexer.java
```diff
@@ -125,6 +129,10 @@ public void process(final HashMap<String, String> json) {
 		if (json.containsKey(Properties.PARENT.getName())) {
 			updateRequest.parent(json.get(Properties.PARENT.getName()));
 		}
+
+		if (json.get(Properties.ID.getName()).startsWith("HT018786244")) {
+			System.out.println();
+		}
```
fsteeg: This looks like a debugging leftover.

pom.xml
```diff
@@ -43,9 +43,19 @@
...
+		<dependency>
 			<groupId>org.elasticsearch</groupId>
 			<artifactId>elasticsearch</artifactId>
-			<version>1.3.2</version>
+			<version>2.3.3</version>
+		</dependency>
+		<dependency>
+			<groupId>com.sun.jna</groupId>
+			<artifactId>jna</artifactId>
+			<version>3.0.9</version>
```
fsteeg: Are these two new dependencies (jna, guava) really/still needed? Didn't see any new non-JDK imports below.

dr0i: - guava_18.0 is necessary to use the es client, see http://stackoverflow.com/questions/33544863/java-elasticsearch-client-always-null
- jna is only needed to supress an ES  warning , not important. Going to remove the lib.

src/main/java/org/lobid/resources/ElasticsearchIndexer.java
dr0i: Dough - indeed! Thx

pom.xml
fsteeg: Ah, the issue seems to be different versions of guava pulled in as transitive dependencies by com.google.gdata and elasticsearch (see "Dependency Hierarchy" tab in the Eclipse pom editor). Since we don't depend on guava directly, I think a better solution would be to exclude it in the com.google.gdata dependency.

fsteeg:
I added a line note about the guava dependency, but that does not show up here, so here again:

> guava_18.0 is necessary to use the es client, see http://stackoverflow.com/questions/33544863/java-elasticsearch-client-always-null

Ah, the issue seems to be different versions of guava pulled in as transitive dependencies by com.google.gdata and elasticsearch (see "Dependency Hierarchy" tab in the Eclipse pom editor). Since we don't depend on guava directly, I think a better solution would be to exclude it in the com.google.gdata dependency.

pom.xml
dr0i: Very well - muchas gracias - did that!

fsteeg:
+1

PR #104, opened 2016-09-14
Hbz lobid rdf to json 25 contributor types

dr0i:
Will resolve hbz/lobid-rdf-to-json#25.


dr0i:
This branch is obsolete as it's duplicated with #106 (which has the addition of the newest metafacture-core dependency update).

Issue #112, opened 2016-09-30
Dct:created or dct:modified sometimes NOT subsumed under wdrs:describedBy


dr0i:
See e.g. http://lobid.org/resources/HT018786244

dr0i:
Deployed to production, closing.

PR #113, opened 2016-09-30
Fix dct:modified and dct:created

dr0i:
Subsume these properties always under wdrs:describedby.

Will resolve hbz/lobid-resources#112.


fsteeg:
+1

PR #122, opened 2016-11-17
4 order contributor and creator and subject list

### LINKED ISSUE #4: Ordered lists in JSON-LD
@acka47 wrote in #1:

> Unfortunately, the arrays for creators/contributors and subjects aren't sorted according to the source data as we would like them to. For this, we will need to
> 1. add the order in N-Triples as RDF list or
> 2. discard this approach and move on like we intended to first, i.e. directly creating JSON-LD from source data with Metafacture.

dr0i:
Will resolve #4.


fsteeg:
+1

PR #127, opened 2016-12-01
126 tweak triple encoder uri recognition

### LINKED ISSUE #126: RDF parse exception when a literal starts with "http"
The following exception:

    java.lang.RuntimeException: org.openrdf.rio.RDFParseException: Not a valid (absolute) URI: HTTP

because the title of the book is "HTTP" and the encoder assumes therefore that it's an URI. Thus http://lobid.org/resources/HT013480902 doesn't resolve.

dr0i:
Will resolve #126.


acka47:
+1

PR #144, opened 2016-12-23
Hbz lobid lodmill 730 publication of primary and secondary

dr0i:
Will resolve lobid/lodmill#730.


acka47:
This seems to be already deployed on production. Also, there is one thing that shoudl be changed: Instead of `startDate` we should rather just use `date` as key.

fsteeg:
I switched the indexes this morning, but switched back when I noticed the changes, so should not be deployed any longer.

dr0i:
Would like to merge this first as so many other PRs are branched from this. Would change the "date" with a later commit.

PR #145, opened 2016-12-29
132 remove unnecessarya arrays and objects

### LINKED ISSUE #132: Remove unnecessary JSON objects or add additional information
There are some properties that have an array of one or more JSON objects with only the `id` field as information. One such property is `exemplar`. As many title have a lot of items connected (see e.g. http://lobid.org/resources/HT000786420#!) the already long array gets even more bloated by the object parentheses.

Properties where this is the case: 

- `exemplar`
- `fulltextOnline`
- `hasSupplement`
- `hasVersion`
- `isFormatOf`
- `isPartOf`
- `isPrimaryTopicOf`
- `language`
- `owner`
- `sameAs`
- `seeAlso`
- `similar`
- `source`
- `tableOfContents`
- `webPageArchived`
- `description`

There are two options for dealing with these:

1. Adding additional informato the object, e.g. `type` (for exemplars this would be `Item`) or `label` if available.
2. Removing the object and directly list the URIs in the array.

Add least for `language` we should add the labels. But I doubt that we could add useful information to all of these...

dr0i:
Merge after lobid/lodmill#730.

Resolves #132.


acka47:
+1 (except for the problem mentioned in https://github.com/hbz/lobid-rdf-to-json/pull/68#issuecomment-270338122)

PR #146, opened 2016-12-29
139 separate contributor with two roles

### LINKED ISSUE #139: Separate Contribution objects for person with two roles
From https://github.com/hbz/lobid-resources/issues/79#issuecomment-268818628:

> But I don't think it is ok to list two roles in one `contribution` object. If a person has two roles in a work (e.g. "Sprecher" and "Darsteller") these are actually two contributions with different roles but not one contribution covering two roles.

I posted this question to the Bibframe list and Ray Denenberg just confirmed what I said. Quoting Ray:

> If you want to list two roles, e.g. “illustrator” and “author” for the
> same person, the rule is, declare two separate Contribution resources.  
> Two or more role statements within the same Contribution resource should
> occur only because you want to provide two different representations
> (e.g. two different URIs) for the same role. 

Thus, we should have two separate Contribution objects if the same person has two roles regarding a resource.

dr0i:
Merge after #132.

Resolves #139.


dr0i:
Deployed to staging. We don't have a test example so there is just the one artificially constructed in the test files.

acka47:
+1

PR #147, opened 2016-12-30
Lobid lodmill 715 contributor being creator and special role

dr0i:
Merge after #139.

Resolves lobid/lodmill#715.


dr0i:
lobid/lodmill#715 is rejected. So is this PR.

PR #166, opened 2017-01-13
142bibframe relator role one repo

### LINKED ISSUE #142: Replace bf:relator with bf:role
Ray Denenberg made clear on the Bibframe list ([archive](http://listserv.loc.gov/cgi-bin/wa?A2=ind1612&L=bibframe&T=0&P=2622)) that bf:role will be an object property as bf1:relator. Thus, we can use it instead of bf:relator.

To do:
- Adjust morph
- adjust labels.json

dr0i:
Fixes #142.

This is the first pull request using the incoorporated lobid-rdf-to-json library. It shows, in comparison, how much the workflow is simplified.


dr0i:
Note to self: don't forget to deploy context.

dr0i:
@ChristophEwertowski please have a look at the new simplified workflow: there are now just two changes necessary  to e.g. change a URI (https://github.com/hbz/lobid-resources/pull/166/commits/ad2d2c7fbef779872d9784e09923e3a5796f66c7 for the change in `labels.json` and https://github.com/hbz/lobid-resources/pull/166/commits/dc2d7c02815dc61963a0d588819e82fd759065c2 for the change in the `morph`).

dr0i:
Deployed to staging.

ChristophEwertowski:
As simple as it can get. +1

acka47:
@ChristophEwertowski We need to use the namespace for Biframe 2.0 which is `http://id.loc.gov/ontologies/bibframe/`, thus, the correct property URI is `http://id.loc.gov/ontologies/bibframe/role`.

(Also, I think @dr0i did forget to deploy the context after all? Still "relator" at http://id.loc.gov/ontologies/bibframe/...)

ChristophEwertowski:
@acka47 Looked at code and documentation to understand what's the new procedure. Probably forgot to look at test.lobid.org over it. 
@dr0i Only nt files are changed. Shouldn't the jsonld files also be updated?

acka47:
As the JSON key word stays the same (`role`) the JSON should not change except for the context file.

dr0i:
Yes, only the context changes. But I would want to deploy the context *after* getting the ok, so the productive data is not affected.
So it seems to be useful to deploy a staging context for reviewing. Did this, see:
http://lobid.org/download/context-staging.json

fsteeg:
+1

dr0i:
Deployed to production, even didn't forgot context, see https://github.com/hbz/lobid-resources-web/commit/bfb89e63a5224b273c76dff849fb94cae80b5dbc.
Closing.

PR #181, opened 2017-01-20
Swap bibo/Periodical with bibo/Journal

### LINKED ISSUE #121: Use bibo:Periodical (or similar) instead of bibo:Journal
Realized this [tweeting](https://twitter.com/acka47/status/793066162246606848) about ZDB data in lobid. Many ZDB resources are not Journals in the strict sense but periodicals. That's why we should map MAB `<controlfield tag="052">p||||||||||||||</controlfield>` to `bibo:Periodical` or – as we want to get away from bibo – something equivalent class like `schema:Periodical`.

ChristophEwertowski:
Resolves #121

Switched bibo/Periodical with bibo/Journal because periodical better matches 052.j (zeitschriftenartige Reihe) and 052.p (Zeitschrift).

### DIFF
--- a/src/main/resources/context.json
+++ b/src/main/resources/context.json
@@ -1,0 +1,621 @@
+{
+  "@context" : {
+    "extent" : {
+      "@id" : "http://iflastandards.info/ns/isbd/elements/P1053"
+    },
+    "endDate" : {
+      "@id" : "http://schema.org/endDate",
+      "@container" : "@set"
+    },
+    "altLabel" : {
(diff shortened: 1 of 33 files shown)


dr0i:
+1
Going to deploy it to the staging system at the weekend.
Afterwards we will check if the data is actually correct in the index.

acka47:
+1

dr0i:
Deployed to staging, see e.g. http://test.lobid.org/resources/HT016604323.

PR #184, opened 2017-01-20
Declare "role" a single entry, not an array

### LINKED ISSUE #132: Remove unnecessary JSON objects or add additional information

dr0i:
Resolves #132.

### DIFF
--- a/src/main/resources/labels/labels.json
+++ b/src/main/resources/labels/labels.json
@@ -158,7 +158,6 @@
         "label": "Rolle",
         "name": "role",
         "referenceType": "@id",
-        "container": "@set",
         "multilangLabel": { }
     },
     {
--- a/src/test/resources/jsonld/BT000002852
+++ b/src/test/resources/jsonld/BT000002852
@@ -10,10 +10,10 @@
       "label" : "Freudenberg, Kreis Siegen-Wittgenstein",
       "type" : [ "PlaceOrGeographicName" ]
     },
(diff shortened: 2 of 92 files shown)


acka47:
+1

ChristophEwertowski:
+1

fsteeg:
+1

dr0i:
Deployed to staging, see e.g. http://staging.lobid.org/resources/HT017472134.

acka47:
+1

ChristophEwertowski:
+1

PR #186, opened 2017-01-23
Replace schema.org/publication with /PublicationEvent

### LINKED ISSUE #185: Replace schema.org/publication with schema.org/publicationEvent
publicationEvent fits more than publication. For that reason replace URIs in morph-hbz01-to-lobid and labels.json.

ChristophEwertowski:
Resolves #185

### DIFF
--- a/src/main/resources/context.json
+++ b/src/main/resources/context.json
@@ -91,10 +91,6 @@
       "@id" : "http://purl.org/dc/terms/hasPart",
       "@container" : "@set"
     },
-    "Publication" : {
-      "@type" : "@id",
-      "@id" : "http://schema.org/Publication"
-    },
     "eissn" : {
       "@id" : "http://purl.org/ontology/bibo/eissn",
       "@container" : "@set"
@@ -591,6 +587,10 @@
       "@type" : "@id",
       "@id" : "http://purl.org/ontology/bibo/Series",
       "@container" : "@set"
+    },
+    "PublicationEvent" : {
+      "@type" : "@id",
+      "@id" : "http://schema.org/PublicationEvent"
     },
     "location" : {
       "@id" : "http://schema.org/location",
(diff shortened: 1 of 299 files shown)


acka47:
You overshot the mark here. We just have to replace the type `Publication` as this actually does not exist in schema.org, see http://schema.org/Publication. But http://schema.org/PublicationEvent exists. Regarding the property, there only exists http://schema.org/publication and neither http://schema.org/publicationEvent (which is a class) nor http://schema.org/publicationEvent.

Note that there is a general convention to start RDF properties in lower case and RDF classes in upper case. (Though, there are some exemptions, e.g. the RDA properties...)

acka47:
It's strange that the type information is missing for some resources, e.g. BT000168595 and HT018585406. These are both resources where there is no `publishedBy` data. Is the addition of the type `PublicationEvent` somehow coupled to the existence of `publishedBy`? /cc @dr0i

ChristophEwertowski:
Brought in a new commit.

dr0i:
>  Is the addition of the type PublicationEvent somehow coupled to the existence of publishedBy?

Good catch! Indeed it is. The type is coupled to the existence of `@publisher`. As there are indeed many publications without a publisher I propose to boolean-or it with a `startDate` and with `@placeOfPublication` to be failsafe.

dr0i:
- rebased the branch
- squashed the two commits into one
- added two commits fixing the missing type

dr0i:
Note to myself: after deploying let's review ist again.

dr0i:
Deployed to staging, see e.g. http://test.lobid.org/resources/BT000168595?format=json.

acka47:
+1

fsteeg:
+1

ChristophEwertowski:
+1

PR #202, opened 2017-02-02
Support wildcards in issued range queries

fsteeg:
Will resolve https://github.com/hbz/lobid-resources-web/issues/35

Support via `*`, see: [http://test.lobid.org/resources/search?issued=2010-*](http://test.lobid.org/resources/search?issued=2010-*)

While adding this, I've cleaned up the code and added general tests for building the query, so this is a good changeset and should be merged in principle. However, considering documentation for this brings up an issue we did not discuss yet: since the web UI evolved from NWBib, it still uses the NWBib query parameters (like `issued` here). These are internally mapped to the queryString syntax. For this example:

http://test.lobid.org/resources/search?q=*%20AND%20(%20+publication.startDate:[2010%20TO%20*])

So we would not need many of the parameters, however the queryString is quite verbose as an API.

### DIFF
--- a/web/app/controllers/resources/Application.java
+++ b/web/app/controllers/resources/Application.java
@@ -82,11 +82,6 @@ public class Application extends Controller {
 	/** The number of seconds in one day. */
 	public static final int ONE_DAY = 24 * ONE_HOUR;
 
-	private static final String[] FIELDS =
-			new String[] { "contribution.agent.label", "title", "subject.id", "isbn",
-					"publication.publishedBy", "publication.startDate", "medium.id",
-					"type", "collectedBy.id" };
-
 	/**
 	 * @return The index page.
 	 */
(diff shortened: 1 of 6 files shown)


ChristophEwertowski:
The search works but the limiting of the search not. See [example](http://test.lobid.org/resources/search?issued=2011-NaN*&sort=newest). But since this is another issue, +1.

acka47:
+1 @ChristophEwertowski I don't get your comment on limiting the search. Could you elaborate?

fsteeg:
@acka47: if you try to use the slider under the issued facet at [http://test.lobid.org/resources/search?issued=2010-*](http://test.lobid.org/resources/search?issued=2010-*) it fails.

fsteeg:
Redeployed with fix, see sliders at:

[http://test.lobid.org/resources/search?issued=2010-*](http://test.lobid.org/resources/search?issued=2010-*)
http://test.lobid.org/resources/search?issued=*-2000

acka47:
+1

ChristophEwertowski:
Jup, I meant using the slider to limit the search for publication years.

dr0i:
+1

Issue #189, opened 2017-01-26
Create a data documentation


ChristophEwertowski:
json examples:
* book
* journal article
* book in series
* book in multivolume work
* periodical

acka47:
Sounds good to me as relevant examples. Perhaps we should also add a ZDB resource. The last two (book in series and book in multivolume work) can be covered by one example, see https://github.com/lobid/lodmill/issues/497 with examples http://lobid.org/resources/HT002888677 and http://lobid.org/resources/HT017903191.

acka47:
See also https://wiki1.hbz-nrw.de/pages/viewpage.action?pageId=19136520 for examples of different resource types in the catalogue.

ChristophEwertowski:
Noting examples here.
* book: http://lobid.org/resources/HT010912885 (standard fields with isbn and edition)
* article: http://lobid.org/resources/HT014060796 (containedIn, some subjects)
* book in multivolume work and series: http://lobid.org/resources/HT002888677 (volumeIn, series, multple exemplars)
* periodical: http://lobid.org/resources/HT007216566 (frequency)

dr0i:
Don't know for what you need this exactly but if you want to fast-check changes on the `morph` it makes sense to use resources which are already part of the junit-test set so that the resources are transformed when building the project. If no good test-resource exists yet, add it to the test set.

ChristophEwertowski:
Updated list with examples from the test files so you can see faster if some label/uri has changed and update the documentation:
* article: http://lobid.org/resources/HT013577568
* periodical: http://lobid.org/resources/HT013304490
* book: http://lobid.org/resources/HT013480902
* book in MultiVolumeWork and series: http://lobid.org/resources/HT015865114 but above example is better. For example it is in a real series. Will push example data file next week.

ChristophEwertowski:
Ticket is a sub issue of https://github.com/hbz/lobid/issues/288. 
Made the annotations like the [organisations docu](http://lobid.org/organisations/api/de) in structure. Annotated fields have following points:
* descriptive **name**
* **description** which contains what's in the field and, if directly transferred, the corresponding MAB2-field
* **coverage**/**occurence**
* an **example**

I won't open a pull request since it's already [on the page](http://lobid.org/resources/api). 

Feel free to annotate! (And of course to comment here.)

acka47:
Looking through the annotations, I noticed some problems with the chosen example records:

- None of the agents has an altLabel, thus it isn't documented.
[...]

PR #208, opened 2017-02-03
Add new example as replacement for Buch

ChristophEwertowski:
New example contains also altLabels, corporateBody and is set as a simple monograph. @fsteeg Please set the URL so I can annotate also on the test system.

See #189

### DIFF
--- a/web/app/views/api.scala.html
+++ b/web/app/views/api.scala.html
@@ -37,8 +37,7 @@
 	<dd>
 		<p>Springe zu: <a href="#buch">Buch</a>, <a href="#artikel">Artikel</a>, <a href="#periodikum">Periodikum</a>, <a href="#serienband">Serienband</a>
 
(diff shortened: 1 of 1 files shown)


fsteeg:
Deployed to staging: http://test.lobid.org/resources/api

@ChristophEwertowski you should set up your editor to use tabs, or to display whitespace characters, to avoid formatting issues caused by mixing spaces and tabs, see the diff of your commit here: https://github.com/hbz/lobid-resources/pull/208/commits/28f8212e339dd6e6826b44bb74b4e30f8e6a4843

fsteeg:
@ChristophEwertowski as we discussed offline, you can annotate on staging, then remove the old example (by pushing into this branch again), and then I'll deploy the final version to production.

fsteeg:
+1

PR #219, opened 2017-02-09
Ignore order in tests

dr0i:
Merging the lobid-rdf-to-json repo had introduced two versions of jackson library.
Also, the comparison of JSON-LD in one test wasn't as good as in the merged repo
(concerning e.g. ordered and unordered data) - so this switches the comparison to the better implementation.


fsteeg:
+1

PR #236, opened 2017-02-14
Switch advanced search autosuggest to internal API

fsteeg:
Will resolve https://github.com/hbz/lobid-resources-web/issues/28

Also enable autosuggest for agent and subject search (in addition to title).

Deployed to staging: http://test.lobid.org/resources/advanced

### DIFF
--- a/web/app/controllers/resources/Accept.java
+++ b/web/app/controllers/resources/Accept.java
@@ -19,7 +19,7 @@ public class Accept {
 	}
 
 	enum Format {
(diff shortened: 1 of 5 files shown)


ChristophEwertowski:
I think that at last we agreed to leave the autosuggest in as it is and if there is time or a customer who needs an autosuggest which works in a specific way, it will be updated. +1

dr0i:
+1

Issue #237, opened 2017-02-15
Update context location


fsteeg:
We should update the context location:

- [x] in the data: http://lobid.org/resources/context.jsonld
- [x] remove the Apache redirect for `http://lobid.org/download/context.json`
- [x] write the context to `lobid-resources/web/conf/context.jsonld` directly
- [x] update context documentation in README (how to change context)

dr0i:
Deployed to staging, see http://test.lobid.org/resources/context.jsonld.
(sorry to have the PR as a bundle (see above) and not solely PRs. Will change that in future as discussed offline.)

dr0i:
The apache redirect should be removed after having deployed this to production.

fsteeg:
+1

dr0i:
Removed the apache redirect. Note that `http://lobid.org/download/context.json` will nonetheless resolve because of the redirect for the hole `download` directory, which is needed for other things.
Closing this issue.

PR #238, opened 2017-02-15
Include query metadata in search response

fsteeg:
Will partially resolve https://github.com/hbz/lobid-resources-web/issues/26

Still required: context updates (after #237) and corresponding updates in lobid-organisations.

This uses Hydra vocabulary/structure as suggested by @acka47 in https://github.com/hbz/lobid-resources-web/issues/26#issuecomment-278006380. Note that it does not (yet) implement the actual page views, for which the spec has changed compared to the linked OER Worldmap implementation, see http://www.hydra-cg.com/spec/latest/core/#collections and https://github.com/HydraCG/Specifications/issues/42#issuecomment-151644982.

Deployed to staging: http://test.lobid.org/resources/search?q=nippes&format=json

### DIFF
--- a/web/app/controllers/resources/Application.java
+++ b/web/app/controllers/resources/Application.java
@@ -10,7 +10,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -18,7 +17,6 @@ import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Spliterators;
 import java.util.UUID;
-import java.util.concurrent.Callable;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


acka47:
+1 @fsteeg, will you I open an issue here and in lobid-organisations for adding collection/PartialCollectionView information using hydra? Looking at the [hydra mailing list](https://lists.w3.org/Archives/Public/public-hydra/2015Oct/0163.html) the discussion on this has come to an end and there won't be any changes there.

fsteeg:
@acka47 we can open issues, but I don't think we should prioritize this now, since we don't use these additional fields ourselves, and we can always add them later given we now have the correct structure in place.

ChristophEwertowski:
+1

fsteeg:
For the required context changes after #237 see https://github.com/hbz/lobid-organisations/compare/web-26-api?expand=1#diff-5d4c3e2f27ae916edf73f72eb7aceb25.

fsteeg:
Discussed offline with @acka47: with the new structure and hydra vocab it actually makes sense to include the aggregations in the search responses, as the last key in the root object. That way they are out of the way when looking at a response (unlike the previous approach where they were in the first element of the result array), and we get all relevant data for a search with a single request. So we'll use the same structure as currently deployed for organisations-staging for resources as well, see: 
http://test.lobid.org/organisations/search?q=hbz&format=json

dr0i:
+1

fsteeg:
Merged via https://github.com/hbz/lobid-resources/commit/3433263ec1a4a111b0822d6978c196cc195259aa, closing.

Issue #221, opened 2017-02-13
Puplishers with double angle brackets aren't shown correctly in HTML


acka47:
Similar to https://github.com/hbz/nwbib/issues/104 where @dr0i removed the "crabs" from the data.
Example: http://lobid.org/resources/HT018147158 http://nwbib.de/HT018147158

From the JSON-LD:

```json
"publication":[
   {
      "location":"Köln",
      "publishedBy":"<<Die>> Ehrenfelder, Gemeinnützige Wohnungsgenossenschaft",
      "startDate":"2008",
      "type":[
         "PublicationEvent"
      ]
   }
]
```

dr0i:
Is this a problem of the data or the html frontend?

acka47:
It depends on the perspective. In https://github.com/hbz/nwbib/issues/104#issuecomment-95931936 you argued it should be fixed in the data. For consistency, I suggest to also fix the data in this case.

acka47:
@ChristophEwertowski You can easily see in https://github.com/lobid/lodmill/pull/696/commits/1ebde307432521ebefb34397d7dea384aa89b6a6 how to remove the angle brackets.

dr0i:
Deployed to stage. @fsteeg no UI update is needed, yes?

fsteeg:
Right, looks good: http://stage.lobid.org/resources/HT018147158

PR #245, opened 2017-02-16
Remove link to Aleph catalogue

### LINKED ISSUE #232: Remove link to Aleph OPAC from RDF
Basically, just remove https://github.com/hbz/lobid-resources/blob/28100eff6e40f0847084c28c719ebabb3aa811bd/src/main/resources/morph-hbz01-to-lobid.xml#L205-L209. We don't need it in the RDF and probably not in the frontend. If we want to show it in the frontend, we can add it easily as it is based on hbz ID.

ChristophEwertowski:
Also remove corresponding property isPrimaryTopicOf from labels.json.

Resolves #232.
Merge after #221.


ChristophEwertowski:
The Travis test leads to a failure which says it can't transfer a sesame file and ultimately leads to the situation that mvn clean install can't be executed. Since it works locally on my side: What could be the error, @dr0i?

fsteeg:
@ChristophEwertowski Since only the pull request build failed, not the push build, this might very well be a Travis issue (like a network timeout when getting the sesame lib). I restarted the pull request build.

fsteeg:
Travis seems to have issues, it can't show build details and aborted the build very quickly.

We should restart it later (from https://travis-ci.org/hbz/lobid-resources/pull_requests).

PR #259, opened 2017-02-20
Replace LvContainedIn with BibframePartOf

### LINKED ISSUE #228: Replace lv:containedIn by bf:partOf
We could KEEP the JSON key `containedIn` for this. If not, we will have to adjust the frontend code.

ChristophEwertowski:
As suggested in https://github.com/hbz/lobid-resources/issues/228 the JSON key
isn't updated because of dependencies in the frontend.

Resolves #228


dr0i:
+1

acka47:
Thinking about this, `lv:containedIn` is the better choice as it is specific about being used to link articles too the bibliographic resource it is linked to. We will leave it as is and might add a statement `lv:containedIn rdfs:subPropertyOf bf:partOf .` to the vocab later if we see the need to align with bibframe. Closing.

PR #261, opened 2017-02-20
Update URL for hbz01 MabXmlClobs used by download test-set

dr0i:
Set URL of hbz01 MabXmlClobs source to productive one
The "test" subdomain gives http status code 404.


fsteeg:
+1

Issue #230, opened 2017-02-14
Use bf:supplement for link to supplement


acka47:
...not http://rdaregistry.info/Elements/u/P60281. We have to upadte the morph and the labels.json.

ChristophEwertowski:
Updated morph.xml and labels.json but got ```Failed tests:   testJson(org.lobid.resources.Hbz01MabXml2ElasticsearchLobidTest)``` and a ```WARN [main] (CompareJsonMaps.java:83) - At least this element is missing in actual: [supplement, id]```
Why a key is missing there, I couldn't comprehend. Also the formating of the json is broken again for the output/json/ files. What do you think @dr0i ?

dr0i:
@ChristophEwertowski The pretty printing is enabled in #224 which is yet to be merged into master. The "missing element" may well be a result of a resource which is not transformed from the beginning, in the error log I can see a problem with HT013480902. So this may be also solved with merging #224 first and rebasing that master into this branch.

dr0i:
Deployed to staging. Now update UI.

fsteeg:
It seems `hasSupplement` is not used in the UI, reassigning for functional review.

acka47:
JSON looks ok but the context hasn't been updated yet.

dr0i:
JSON context for staging is deliberately _not_ deployed, see first comment in https://github.com/hbz/lobid-resources/pull/262.

acka47:
+1

ChristophEwertowski:
+1

dr0i:
Deployed to production index.

PR #248, opened 2017-02-17
Replace FRBR with bibframe

dr0i:
This is the reopened branch which was merged and reverted in #224.
Will resolve #204.

PR #262, opened 2017-02-20
Replace rda for bibframe supplement

dr0i:
See #230.

Deployed small test set to staging , see e.g. http://test.lobid.org/resources/HT000009600?format=json.

Note that @fsteeg and me decided offline not to deploy the context (at http://test.lobid.org/resources/context.jsonld) so you have to look in the commit for the changes in context.


ChristophEwertowski:
+1

acka47:
+1

fsteeg:
First, https://github.com/hbz/lobid-resources/pull/248#issuecomment-281306129 applies here too.

Second, I see a change that I think we did a long time ago (array changes in http://lobid.org/resources/HT013480902?format=json, which does not have arrays for agent and title, but the diff shows that, see https://github.com/hbz/lobid-resources/pull/262/files#diff-a028cb038fe59ff798780bcd141e353b).

Then there are the context changes, which I reviewed in https://github.com/hbz/lobid-resources/issues/237#issuecomment-281315619.

For `hasSupplement` -> `supplement`: +1

dr0i:
Re: second:
Only one resource is affected, and it was one old resource that newer got updated because the json-comparator checks if the content is changing only, not syntax. So that test resource just now changed in content (different context location) and that got triggered the rebuild of the resource as one would now expect.

PR #268, opened 2017-02-24
260 free running dewey

dr0i:
Treating a dewey subject the same as the other subjects. As @jschnasse agreed that it doesn't matter if all to treat as a set or a list I choosed a set because it's easier to do.


acka47:
+1

acka47:
+1

acka47:
The changes for `subject` look good. But obviously, for `subjectOrder` there is now always a list with one element, see e.g. https://github.com/hbz/lobid-resources/pull/268/files#diff-0ade426ab68de19984c0705567be0e08R76

dr0i:
Hi @acka47 why reassigning me? I started a review and assigned you because I have an unanswered question.

acka47:
> Hi @acka47 why reassigning me? 

Because I think there went something wrong with the "subjectOrder" data, see https://github.com/hbz/lobid-resources/pull/268#issuecomment-282280012.

dr0i:
Yes, but where do you see this? I cannot as I tried to point out in the coment:

>  but here is not a list with only one entry, yes?

If I search for `subjectOrder` in the diff you provided the first hit shows a `subjectOrder` with multiple gnd-entries, no?

acka47:
Sorry, I didn't understand that had the literals for `subjectOrder` in a list all the time. +1

PR #269, opened 2017-02-24
Improve load time of HTML pages

### LINKED ISSUE #264: Improve load time of HTML pages
Currently, loading a single resource via its URL in the browser – e.g. http://lobid.org/resources/HT002948556 – takes > 5s, which is even longer than loading search results (which isn't that quick either). At some point we need to improve the performance before the launch. Directly getting the JSON is fine though (e.g. http://lobid.org/resources/HT018138118?format=json).

fsteeg:
Will resolve #264


dr0i:
+1

PR #274, opened 2017-02-24
795 replace daia label and rda manuscript

ChristophEwertowski:
Resolves https://github.com/lobid/lodmill/issues/795

(edit @dr0i:  Added the missing resources to the test fle set ~~The deleted files are the one which in the master exist only as json files but not in the ClobsXml from which they are built. Because I had to generate the test data from scratch, they couldn't be generated from the ClobsXml. If the resources are inserted again into the ClobsXml, they will be added again as json files.~~)

Merge after #245.


acka47:
Looks good. But I will take another look as soon as the test set is deployed.

PR #275, opened 2017-02-24
Update UI for new items fields

### LINKED ISSUE #204: Replace FRBR "Item" class + properties with bibframe
FRBR still occurs four times in the JSON-LD context. We have already removed `frbr:Manifestation` (see https://github.com/lobid/lodmill/issues/639) and I suggest to replace the rest with Bibframe:

- http://purl.org/vocab/frbr/core#Item -> http://id.loc.gov/ontologies/bibframe/Item
- http://purl.org/vocab/frbr/core#exemplarOf -> http://id.loc.gov/ontologies/bibframe/itemOf
- http://purl.org/vocab/frbr/core#exemplar -> http://id.loc.gov/ontologies/bibframe/hasItem
- http://purl.org/vocab/frbr/core#owner -> http://id.loc.gov/ontologies/bibframe/heldBy

I guess the mapping above is correct. I had some problems with the definition of `bf:itemOf` and `bf:hasItem`, though, see http://listserv.loc.gov/cgi-bin/wa?A2=ind1702&L=bibframe&T=0&P=1431. We should wait for the response to my mail before we implement this.

fsteeg:
Will resolve #204

### DIFF
--- a/web/app/controllers/resources/Index.java
+++ b/web/app/controllers/resources/Index.java
@@ -56,6 +56,7 @@ public class Index {
 					.map(v -> v.unwrapped().toString()).collect(Collectors.toList());
 	private static final String CLUSTER_NAME =
 			Application.CONFIG.getString("index.cluster.name");
+	private static final String OWNER_ID_FIELD = "heldBy.id";
 
 	private JsonNode result;
 	private long total = 0;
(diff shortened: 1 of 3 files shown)


dr0i:
+1

Issue #273, opened 2017-02-24
Fix Travis timeouts


fsteeg:
The Travis build often times out due to 'no output for 10 minutes'

PR #276, opened 2017-02-28
Increase timeout for Travis

fsteeg:
Will resolve https://github.com/hbz/lobid-resources/issues/273

See build details: https://travis-ci.org/hbz/lobid-resources/builds/206095808

Configured as documented in:
https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

Add a separate compilation step with increased timeout to avoid higher timeout when actually running tests (where timing out would indicate a problem).


dr0i:
+1

PR #300, opened 2017-03-08
285 publications dates periodicals

### LINKED ISSUE #226: Bad publication dates lead to wrong year facet
I used the publication year bar to limit the results to seemingly very old resources: http://lobid.org/resources/search?issued=340-477&sort=oldest. Not surprisingly they aren't as old as that but only badly cataloged. If you look at the json files, you can see that they have startDates in the 3000s and 4000s. They should be displayed after the current year. 
Even better would it be to display all resources up to the current year + 5 years and out of the publication year bar (+5 years to show also publications which aren't published yet but cataloged). Then the publication bar could be used to filter in smaller steps. Publishing it first for discussion.

### LINKED ISSUE #285: Transform 425 b/c for periodicals
This came up at https://github.com/hbz/lobid-resources/issues/239#issuecomment-283661661.

From the MAB docu:

```
425       ERSCHEINUNGSJAHR(E)

          Indikator:
          blank = Erscheinungsjahr(e) in Vorlageform
          a     = Erscheinungsjahr(e) in Ansetzungsform
          b     = Erscheinungsjahr des ersten Bandes in Ansetzungsform
          c     = Erscheinungsjahr des letzten Bandes in Ansetzungsform
          p     = Publikationsdatum bei Tontraegern (P-Datum)

```

From the MAB source of the example:

```xml
<datafield tag="425" ind1="b" ind2="1">
  <subfield code="a">1859</subfield>
</datafield>
<datafield tag="425" ind1="c" ind2="1">
  <subfield code="a">1934</subfield>
</datafield>
```

Expected outcome:

```json
"publication":[
   {
      "endDate":"1934",
      "location":"Berlin",
      "publishedBy":"Weidmann",
      "startDate":"1859",
      "type":[
         "PublicationEvent"
      ],
      "frequency":{
         "id":"http://marc21rdf.info/terms/continuingfre#i",
         "label":"dreimal wöchentlich"
      }
   }  ]
```

ChristophEwertowski:
Resolves #285 , resolves #226.
Also added to exceptions for improving the start- and endDates, thus resolving #226:
* multiple resources with wrong year statements have volumes in the publication year field ([look at the results of this search](http://lobid.org/resources/search?t=periodical&sort=newest)). If I look now at the search, it seems they are not that many. It can be removed if the years are corrected.
* years are only taken into account if they start with an one or two (instead of testing if they consist of four numbers)


ChristophEwertowski:
Multiple resources have an indicator ind2="2" in the MAB2 field 425 and thus aren't registered as publication years because we only search for 425 fields with ind2="1" ([example](http://test.lobid.org/resources/HT008359488), [line in the morph](https://github.com/hbz/lobid-resources/blob/master/src/main/resources/morph-hbz01-to-lobid.xml#L673)). @acka47, @fsteeg: What's the "1" standing for? Can we just add the resources with the "2"?

fsteeg:
Not sure which "1" you mean, but to me it seems `425[abc-p][-1].a` should become `425[abc-p][-12].a`

acka47:
`ind2="2"` means that the information originaly comes from the superordinated resource, in the case of http://test.lobid.org/resources/HT008359488 this is [HT002197545](http://lobid.org/resources/HT002197545). Thus, it makes sense to not use the information from `ind2="2"` for populating `startDate`/`endDate`.

ChristophEwertowski:
Deployed at test system. Examples: http://test.lobid.org/resources/HT000009600 / http://lobid.org/resources/HT000009600 and http://test.lobid.org/resources/HT002619538 / http://lobid.org/resources/HT002619538.

Also for the exceptions: 
* only years are taken into account if they begin with an one or two. See http://test.lobid.org/resources/search?sort=oldest / http://test.lobid.org/resources/search?sort=newest. Resources which aren't affected by this rule are at the end of the list (http://test.lobid.org/resources/search?from=4575&sort=newest). So no 5732s and such-a-likes. Also we have no real publication years which start with a 0 like 0950 for 950 A.D. 
* the exception for 425- s I couldn't test with the small test set.
@acka47 : Good. Then happy reviewing.

dr0i:
Deployed test to http://test.lobid.org/resources/HT002619538.
Check if UI updates are needed and then assign @fsteeg, else reassign me so that I deploy it to production.

acka47:
@fsteeg The `endDate` is not shown yet. It would be great to have it in the "Erschienen" field like "Erschienen: 1955 - 2004"

fsteeg:
@acka47 All the samples linked here seem to have identical start and end dates. I'd suggest that we only show e.g. "Erschienen: 2004" in these cases. Do you have an example with different start and end dates for testing?

ChristophEwertowski:
Also there is http://lobid.org/resources/HT008733617?format=json where there are two startDates of which one is the endDate.

ChristophEwertowski:
Opened the issue #315 because it's a secondary publication with a startDate and endDate which isn't handled yet.

fsteeg:
Opened a new issue, since it seems like further morph changes will be required for correct start and end dates (I don't think the identical start and end dates is a deployment issue, since they are identical on production too, see http://lobid.org/resources/search?q=publication.endDate:*&format=json): https://github.com/hbz/lobid-resources/issues/316

dr0i:
Merged with https://github.com/hbz/lobid-resources/commit/7e0f45b97a27c823fab6c757fccfd1f3ccd01bf3, but didn't closed automagically.
Seems that if base for merging is NOT the master github is not satisfied if branch is merged into master but if it's exeplicitly merged into the base's branch.
However - closing this manually.

dr0i:
Right - just changed the base to master - and voila :)

PR #304, opened 2017-03-09
297 add datatypes to start and enddate

### LINKED ISSUE #297: Add datatypes to startDate, endDate
i.e. add to the `@context`. (I presume we normalize all dates in the transformation.)

```json
"startDate": {
  "@id": "http://schema.org/startDate",
  "@container": "@set",
  "@type": "http://www.w3.org/2001/XMLSchema#gYear"
},
"endDate": {
  "@id": "http://schema.org/endDate",
  "@container": "@set",
  "@type": "http://www.w3.org/2001/XMLSchema#gYear"
}
```

Related to #285.

ChristophEwertowski:
Resolves #297 .

Also adds the same datatype to dateOfBirth and dateOfDeath. Because XML Schema gYear allows only four digit numbers year statements with only three digits are expanded by a leading 0.


dr0i:
Deployed to stage. @fsteeg ready for (possibly needed) UI update.

fsteeg:
I don't think the year changes need UI updates. However, this pull request contains the changes from https://github.com/hbz/lobid-resources/pull/305 (@dr0i: it seems you explicitly merged it), which do require UI updates. But since we have that separate pull request I can +1 this.

PR #323, opened 2017-03-17
Revert "Remove unimportant library "jna""

### LINKED ISSUE #322: Add JNA library to avoid WARN when building
Readding the JNA library will prevent disturbing WARN messages in the logs when maven is build. Decided with @fsteeg that adding this dependency at last brings more advantages since the `WARN` in the logs are an annoyance for especially @acka47 and @ChristophEwertowski and also for me.

dr0i:
This reverts commit 02d5066cc4ff92b3e68f7a1f082ede5892809f7e.

Resolves #322.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,6 +55,11 @@
 			<version>2.3.3</version>
 		</dependency>
 		<dependency>
+			<groupId>com.sun.jna</groupId>
+			<artifactId>jna</artifactId>
+			<version>3.0.9</version>
+		</dependency>
+		<dependency>
 		  <groupId>com.ning</groupId>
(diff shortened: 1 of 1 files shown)


dr0i:
Hm! Interstingly THIS builds while many other branches don't. Maybe their problem has indeed something to do with `jna` because of `Java HotSpot(TM) 64-Bit Server VM warning: You have loaded library /tmp/jna8576409803750626595.tmp which might have disabled stack guard. The VM will try to fix the stack guard now.` (note the /tmp/jna...) (see e.g. https://travis-ci.org/hbz/lobid-resources/builds/212058093 ).

fsteeg:
+1

PR #324, opened 2017-03-17
72 marc relator codes in mabrda repair

ChristophEwertowski:
Test


src/main/resources/labels/labels.json
```diff
@@ -5671,6 +5671,96 @@
         "multilangLabel": { }
     },
     {
+        "uri": "http://purl.org/lobid/lv#contributorOrder",
+        "label": "contributorOrder",
```
dr0i: This is already part of the context-labels, see https://github.com/hbz/lobid-resources/blob/72-MARCRelatorCodesInMABRDARepair/src/main/resources/labels/context-labels.json

dr0i: Let's have a look at the original branch `72-MARCRelatorCodesInMABRDA` and fix that one.

ChristophEwertowski:
Conflicts resolved in original branch 72-MARCRelatorCodesInMABRDA

PR #332, opened 2017-03-21
315 add start date end date secondary publications

### LINKED ISSUE #315: Add endDate to publications with secondaryPublicationType
http://lobid.org/hbz01/HT008733617 is transformed into two startDates of the same publication because there is no endDate for secondary publications. There are two ways to handle this:
* only include the first number of the 619 field
* also add an endDate for secondary publications

ChristophEwertowski:
Resolves #315


dr0i:
Deployed to production, see e.g. http://lobid.org/resources/HT008733617?format=json.

PR #371, opened 2017-04-05
Take right number as startDate from MAB089

### LINKED ISSUE #321: Only take right publication years from volume field
@acka47 said in #285 [comment](https://github.com/hbz/lobid-resources/issues/285#issuecomment-287305353):
> Hmm, http://lobid.org/resources/TT001966831 doesn't look satisfying as the wrong value for startDate is taken from 089 ("2977" instead of "1999"):
>```<datafield` tag="089" ind1="-" ind2="1">
  <subfield code="a">59,2. 1999 = 2952/2977</subfield>
</datafield>```
>There are approximately 100 other resources where the wrong number is taken: http://lobid.org/resources/search?size=100&sort=newest

ChristophEwertowski:
Resolves #321


dr0i:
+1

dr0i:
Deployed to production.

Issue #330, opened 2017-03-21
Implement scroll parameter


acka47:
Subissue of https://github.com/hbz/lobid/issues/288

fsteeg:
Updated title, since there is no scroll parameter implemented in lobid-resources 2.0 yet.

Some links provides by @dr0i in mail correspondence with @hagbeck:

- https://www.elastic.co/guide/en/elasticsearch/guide/current/pagination.html
- http://elasticsearch-users.115913.n3.nabble.com/Is-ES-capable-of-doing-pagination-td3399087.html
- https://github.com/hbz/lobid/issues/176

fsteeg:
For consistency, we should implement this for lobid-organisations too.

fsteeg:
Current implementation deployed to production:
http://lobid.org/resources/search?q=ehrenfeld&format=bulk

We will do the actual functional review for this as part of https://github.com/hbz/lobid/issues/329

For details on the implementation see https://github.com/hbz/lobid-resources/pull/372#issue-219932827

PR #372, opened 2017-04-06
Support bulk requests

fsteeg:
Will resolve https://github.com/hbz/lobid-resources/issues/330.

Sample usage (increasing response size):

`curl "http://localhost:9000/resources/search?q=ehrenfeld&format=bulk" > ehrenfeld.jsonl`
`curl "http://localhost:9000/resources/search?q=dom&format=bulk" > dom.jsonl`
`curl "http://localhost:9000/resources/search?q=*&format=bulk" > resources.jsonl`

Response format for `format=bulk` is JSON lines (see http://jsonlines.org/).

Updates since a specific date can be requested as part of the query (was `scroll=20100101` in 1.x):

`curl "http://localhost:9000/resources/search?q=ehrenfeld+AND+describedBy.dateModified:>20100101&format=bulk" > ehrenfeld-since-2010.jsonl`

Compression will be provided via Apache, as for 1.x.

All of this works with a Play instance running locally (57.8G in about 2:40 hours for `q=*`), also when curling from a different machine on our local network. When deployed on our servers (tested on quaoar1 and weywot3), the `q=*` request stalls after downloading about half of the heap size (about 500M with default activator settings). I'm currently working on setting up an environment to closely mirror my machine on weywot3.

@dr0i I'd suggest that you try to reproduce this on your machine, and if that works too we merge and deploy this, since the implementation does not seem to be the problem. Also many common use cases, like getting the NWBib, holdings of a specific library, etc. require way less data and seem to work fine. We can then work on issues with our server setup for the specific case of very large dumps in a separate issue, without delaying the basic bulk download functionality.

### DIFF
--- a/web/app/controllers/resources/Accept.java
+++ b/web/app/controllers/resources/Accept.java
@@ -20,7 +20,8 @@ public class Accept {
 
 	enum Format {
 		JSON_LD("json(.+)?", "application/json", "application/ld+json"), //
-		HTML("html", "text/html");
+		HTML("html", "text/html"), //
+		BULK("bulk", "application/x-jsonlines");
 
 		String[] types;
 		String queryParamString;
--- a/web/app/controllers/resources/Application.java
+++ b/web/app/controllers/resources/Application.java
@@ -179,6 +179,15 @@ public class Application extends Controller {
 			uuid = UUID.randomUUID().toString();
 			session("uuid", uuid);
 		}
+		String responseFormat = Accept.formatFor(format, request().acceptedTypes());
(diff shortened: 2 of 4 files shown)


dr0i:
Works on my machine. Since it's slow (360kb/s) and compression is enabled via the apache proxy (so not usable when usable when testing directly against the play-app) I cancelled that last test (would take ~40h when dealing with ~60gb ).

As you said it should be safe to deploy this, enabling needed functionality. 

This RAM-issue seems to be strongly related to https://github.com/hbz/lobid/issues/188.

PR #381, opened 2017-04-10
Remove obsolete java option

dr0i:
As the "Permanent Generation" was removed in Java 8, on some machines
maven fails with "Unrecognized VM option 'MaxPermSize=256m'".
It's safe to simply remove this option.


fsteeg:
+1

PR #390, opened 2017-04-13
Remove dct source, rename key for isFormatOf

### LINKED ISSUE #388: Remove dct:source, rename key for dct:isFormatOf
As discussed with @ChristophEwertowski offline today we will proceed as mentioned in https://github.com/hbz/lobid-resources/issues/265#issuecomment-293499700:

> As `dct:source` is used in exactly the same way as `isFormatOf` (see the [current morph, lines 220-227](https://github.com/hbz/lobid-resources/blob/a403180f75e9e9fc07d13f6046de775a0560a8b7/src/main/resources/morph-hbz01-to-lobid.xml#L220-L227)), I propose to discard `dct:source` altogether.
>
> `021` carries the "IDENTIFIKATIONSNUMMER DER PRIMAERFORM". `isFormatOf` – for two reasons – seems to be the better choice anyway. (As we call the publication information for the secondary form `SecondaryPublicationEvent`, we might think about changing the JSON key of `isFormatOf` to `primaryForm` or so.)

ChristophEwertowski:
Resolves #388

### DIFF
--- a/src/main/resources/labels/context-labels.json
+++ b/src/main/resources/labels/context-labels.json
@@ -274,11 +274,11 @@
       "name" : "hasVersion"
    },
    {
-      "name" : "isFormatOf",
+      "name" : "primaryForm",
       "multilangLabel" : {},
       "referenceType" : "@id",
       "uri" : "http://purl.org/dc/terms/isFormatOf",
-      "label" : "Format von"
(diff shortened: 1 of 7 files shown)


dr0i:
+1

Issue #8, opened 2015-12-17
Remove obsolete properties


dr0i:
We have a workaround `lv:nameOfContributingCorporateBody` (introduced with hbz/lobid#38), where we subsume literals of corporate bodies, be they creators or contributors. I think this property will be made obsolet by the new `lv:contributorOrder` , right?

acka47:
I re-neamed this ticket to a more general task: remove obsolete properties _after_ properly implementing API 2.0.

When we are properly done with API 2.0 the following properties should be obsolete and should be removed before going into production:

- [ ] `contributingCorporateBodyLabel`: with #40 corporate body search will work with `?q=corporateBody
- [ ] `contributorLabel`: person search will work with `?q=contributor`, see also #40
- [ ] `subjectlabel`, as a subject search works with ?q=subject.*:Pferd
- [ ] `subjectChain`, provided that there is a way to search for a subject chain by using the `subject` list

acka47:
The question is whether these properties are really obsolete or whether they still contain information that isn't found elsewhere. For example, I just noticed that `subjectlabel` contains information from fields 710/711 that is not – and should not– moved to the `subjectarray. Example [TT002234459](http://lobid.org/hbz01/TT002234459) (snippet):

```xml
<datafield tag="710" ind1="-" ind2="1">
  <subfield code="a">
Wandern, Diez, Lahnstein, Rhein-Lahn-Info, Rhein, Lahn, Wirtschaftsförderung
  </subfield>
</datafield>
```

We might not want to have it in the subject array as it would be mistaken as some part of an RSWK subject chain. (There is still the possibility, though, that we need a field like `subjectChain` or `subjectOrder` for providing required functionality like "Themensuche" in NWBib).

We will have to take this into accoutn before removing properties. One possibility for 710/711 might be to switch back to dc:subject for this when we don't list the strings together with alternate names of GND subjects...

acka47:
Talking to @jschnasse the other day, it became clear that we should keep "subjectOrder" as we can not group multiple subject chains in a `subject` array.

dr0i:
Look at the status quo of API 2.0 I think we may close this ticket. Only `subjectChain` is still used and will be in the future because at the moment `json-ld` doesn't support lists in lists, yes @acka47 ?

acka47:
[...]

PR #391, opened 2017-04-13
Remove redundant and deprecated elements from context

acka47:
As we have bibo:Periodical twice, I remove the one with the wrong JSON key "Journal".


acka47:
I also removed `primaryTopic`.

ChristophEwertowski:
No resource with type "journal" exists anymore ([query](http://lobid.org/resources/search?q=type:Journal)). And primaryTopic should have been removed ([see comment](https://github.com/hbz/lobid-resources/issues/8#issuecomment-279723868)) +1

Issue #393, opened 2017-04-18
Subject facet missing


fsteeg:
http://lobid.org/resources/search

fsteeg:
@dr0i: As a prerequisite to updating the subject facet for the new subject structure, the `subject.componentList.id` field needs to be set to not_analyzed (like `subject.id`).

@acka47: I've hit this issue when updating the API docs too. With the new structure we always have to check two fields (resp. 4, if we include labels) when doing subject queries (`subject.id|label` and `subject.componentList.id|label`). That may be a price we're willing to pay for modeling complex subjects, I just wanted to mention it.

acka47:
Re. subject queries: I am aware of the multiple fields to query. There also is the – probably most important – field `subjectAltLabel`.

acka47:
For the subject facet, we should only query `subject.componentList.label`, though, as this is the field that is used most over the whole dataset.

ChristophEwertowski:
@dr0i : Also the componentList gets lost if one generates the files new. And other changes aren't in the master files (the changed context, output and reverseTest files).

dr0i:
re. "componentList gets lost": @ChristophEwertowski I couldn't reproduce your observation. Rebuilding the master anew after removing all output test files shows the creation of `componentList` just fine. Doing this I discovered that some test files reside in the directory *although* they *don't exist in the hbz01XmlClobs.tar.bz2 archive*. These should be in sync. Filed #402. Maybe you looked in some of these files containing old data (i.e. without `componentList`)?

fsteeg:
@dr0i Opened pull request #403 with field update for facet, and deployed to test:

http://test.lobid.org/resources/search?q=k%C3%B6ln

When the `not_analyzed` setup is deployed to the test index, this should work.

ChristophEwertowski:
@dr0i : Yes, it is true that all these changed files aren't in the XmlClobs but in the json/output folder files.

fsteeg:
Deployed to test: http://test.lobid.org/resources/search

ChristophEwertowski:
+1

acka47:
+1

PR #403, opened 2017-04-20
Update field to use for subject facet

fsteeg:
Will resolve https://github.com/hbz/lobid-resources/issues/393


dr0i:
Deployed to test. Only one subject shows up - is this really all there is? http://test.lobid.org/resources/search

dr0i:
+1

PR #413, opened 2017-04-28
402 missing undetected test files

### LINKED ISSUE #402: Test files in hbz01XmlClobs.tar.bz2 missing undetected
Test files added to `hbz01XmlClobs.tar.bz2` may be removed with a new version of this archive, see b03aa724d61b9511380ca03d661142c0c92431a1 where `HT018781534` is added and 9e18340b8f9d2ac9856737746291bdaa64547c1c where it's removed *while at the same time* the corresponding test file (like ./src/test/resources/jsonld/HT018781534 )  is *still there* (thus calling this *missing undetected*).

This is not good, e.g. because this json test file will never be updated when the data is changed thus reflecting an ever older state thus when e.g. grepping something in the test files it may show up with an outdated data structure, confusing everyone. 

So, go on implement the removing of the test files when the file is removed in the source (`hbz01XmlClobs.tar.bz2`).

dr0i:
Will resolve #402.


fsteeg:
+1

Issue #358, opened 2017-03-31
describedby should be describedBy


fsteeg:
For consistency (general camelCasing, specifically collectedBy), `describedby` should be `describedBy`.

ChristophEwertowski:
Will be set for functional review if other functional reviews were went through.

ChristophEwertowski:
[Example (production)](http://lobid.org/resources/HT017028573?format=json) / [example (test)](http://test.lobid.org/resources/HT017028573?format=json).

fsteeg:
Looks good, +1

I'm a bit confused about the property in general though, `describedBy`, it contains creation and modification dates, and links to itself via `id`? Is this correct @acka47?

acka47:
Yes, it is. `http://lobid.org/resources/{hbzId}#!` identifies the bibliographic resource (the instance/manifestation of a book, the journal etc.) and `http://lobid.org/resources/{hbzId}` identifies the page/file served by lobid describing the bibliographic resource (the "record") . This gives us the possibility to separate statements about the different things. E.g. the creation dates of both necessarily differs from each other.

acka47:
+1

PR #416, opened 2017-05-02
Add items directly to lobid-resources

### LINKED ISSUE #140: Embed item information in resource description
As I argued in today's meeting, I would like to embed item information (owner, call number, possibly geo coordinates of the owner) in the resource description. I also think we can leave out the item URIs altogether and embed them as blank nodes.

dr0i:
This is kind of redundant because there will also be item documents on their
own. But it's convenient to consume one big document instead of having multiple
records distributed forcing a lookup and merge in the process of creating that
big document.

Resolves #140.


dr0i:
@acka47 have a look. It's easy to subsume the *whole* item under the lobid-resource. I think that's also kind of coherent. Is this ok for you or are you determined to stick to onyl have the essential data mentioned in the issue?

acka47:
@dr0i Is it deployed anywhere for review?

acka47:
As already addressed in the meeting today, I generally don't see much problems with including the whole item information as is. Anyhow, a few minor adjustments would be nice. For illustration, here is an example first:

```json
{
   "id":"http://lobid.org/resources/HT000290078#!",
   "hasItem":[
      {
         "callNumber":"Jc F 16826",
         "describedBy":{
            "id":"http://lobid.org/items/HT000290078:DE-290:Jc%20F%2016826",
            "label":"lobid Bestandsressource"
         },
         "heldBy":{
            "id":"http://lobid.org/organisations/DE-290",
            "label":"lobid Organisation"
         },
         "id":"http://lobid.org/items/HT000290078:DE-290:Jc%20F%2016826#!",
         "itemOf":"http://lobid.org/resources/HT000290078#!",
         "label":"lobid Bestandsressource",
         "note":"00000001",
         "type":[
            "Item"
         ]
      }
   ]
}
```

1. My first problem applies to items in general – whether embedded or not (thus, we should probably open a separate issue for this): It occurred to me that the `label` information is currently superfluous and that it would make more sense to add the call number in the field if we retain it or to just get rid of the `label` field for items.
2. What might be confusing and generate problems when dealing with the data is the reference loop `_:Instance bf:hasItem _:Item . _:Item bf:itemOf _:Instance .`
3. Furthermore, it would be nice to remove the `describedBy` information from the embedded item data. 

In conclusion, I'd be happy if the example would rather look like below but if it takes too much effort we can also embed the whole information.

```json
{
   "id":"http://lobid.org/resources/HT000290078#!",
   "hasItem":[
      {
         "callNumber":"Jc F 16826",
         "heldBy":{
            "id":"http://lobid.org/organisations/DE-290",
            "label":"lobid Organisation"
         },
         "id":"http://lobid.org/items/HT000290078:DE-290:Jc%20F%2016826#!",
         "note":"00000001",
         "type":[
            "Item"
         ]
      }
   ]
}
```

dr0i:
Re 1): Kept the `label`, for the idea was to _always_ have one for _every_ `id`. Even if it seems redundant (as it's the same as in `callNumber`) the semantic is different.

Deployed to test, see e.g. http://test.lobid.org/resources/HT015891797?format=json

ChristophEwertowski:
+1

acka47:
+1

src/main/java/org/lobid/resources/RdfModel2ElasticsearchEtikettJsonLd.java
```diff
@@ -87,33 +87,46 @@ public RdfModel2ElasticsearchEtikettJsonLd(final Object jsonLdContext) {
...
 			if (subjectResource.isURIResource()) {
 				if (subjectResource.getURI().startsWith(LOBID_ITEM_URI_PREFIX)) {
 					submodel = extractSubmodel(submodel, subjectResource);
 					toJson(submodel,
 							subjectResource.getURI().toString().replaceAll("#!$", ""));
-					// remove the newly created sub model from the main node
-					copyOfOriginalModel.remove(submodel);
+					// remove some properties from that item which clings to the main node
+					removeProperty(copyOfOriginalModel, subjectResource, "itemOf");
+					removeProperty(copyOfOriginalModel, subjectResource, "describedby");
```
fsteeg: I'm a bit confused, since `describedby` is now `describedBy` (see https://github.com/hbz/lobid-resources/issues/358), yet this seems to work, as there is no `describedBy` in the embedded items. Is it still `describedby` internally somehow?

dr0i: We are here at the triple level, i.e. `describedby` is part of the (correct) URI `http://www.w3.org/2007/05/powder-s#describedby`. That's versus "The JSON field name" which is just an arbitrary label choosed in #358 to be camelCased as `describedBy`. This label, and _only_ this, is still there, see the JSON test files.

fsteeg:
+1

dr0i:

PR #444, opened 2017-05-22
Add some missing languages

### LINKED ISSUE #439: Missing languages
See [example](http://lobid.org/resources/HT013208902?format=json) and [query for all resources with portugese language](http://lobid.org/resources/search?q=language.id%3A%22http%3A%2F%2Fid.loc.gov%2Fvocabulary%2Fiso639-2%2Fpor%22). 
It seems to me that the list of labels for languages in labels.json is a bit short. So it would be good to look after resources with non-covered languages.

ChristophEwertowski:
Resolves #439

### DIFF
--- a/src/main/resources/labels/labels.json
+++ b/src/main/resources/labels/labels.json
@@ -13592,36 +13592,284 @@
       }
    },
    {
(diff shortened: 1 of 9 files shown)


dr0i:

PR #445, opened 2017-05-22
Define resources with MAB 051s as books

ChristophEwertowski:
Resolves hbz/lobid#320


dr0i:

PR #453, opened 2017-05-30
441 Change definitions of hasVersion and fulltextOnline

### LINKED ISSUE #441: Clean up definitions of fulltextOnline, hasVersion and seeAlso
The resource [HT018468645 (json)](http://lobid.org/resources/HT018468645?format=json) has the same url three times in it's bibliographic file. The url leads to a publisher abstract stored by the DNB. Since there a lots and lots of abstracts of this type in our data we should remove them from `fulltextOnline` and `hasVersion` and let them stored at `description`. As far as I remember I've never seen other things than abstracts under this domain but I could be wrong. In this case, however, it is clear because the url is stated as an "Inhaltstext".

ChristophEwertowski:
Resolves #441


dr0i:
+1. Reassigning for solving conflicts.

ChristophEwertowski:
Conflicts solved. Reassigning.

fsteeg:
@dr0i Merged to master. Will this be deployed automatically over the weekend? If not, you might have to revisit this when you're back next week.

dr0i:
Good you asked. In general, it is so:
- master will be deployed for the productive index by daily updating
- a deployment of master to the staging system for rebuilding a full new index *might* be done, but chances are that there is a conflict when pulling , thus ignoring the new commits

Now, there was no conflict so deployment started well. Alas, because of #472, it was not build completely.

dr0i:
One more thing, though:
We shouldn't merge into master while the data is not transformed on basis of these commits. This a) avoids that we have a mixed index (old and new data) when doing daily updates on the full index and b) the master should always reflect the productive index to avoid confusion and unhappiness.
I know this is not CI-like, but as we only build our full index two times a week this is, when dealing with data transformation, also our only time frame to merge into master.

dr0i:
Deployed to production, see http://lobid.org/resources/CT003012479.json. Closed.

Issue #187, opened 2017-01-23
Separate subject chains from another in subject array


acka47:
In https://github.com/hbz/nwbib/issues/363#issuecomment-267920766 I wrote: 

> I just remembered that we still have no way of separating two subject chains in the JSON. See for example http://test.lobid.org/resources/HT013099804.

@dr0i said in [response](https://github.com/hbz/nwbib/issues/363#issuecomment-268176121):

> As I remembered we talked about leaving the `subjectChains` as they are. If not, please make an issue and let us discuss _how_ to encode the subject chains under the `subject`-list.

I wouldn' t have brought this up again but today @larsgsvensson [made me aware over at the DINI KIM Titeldatengruppen Wiki](https://wiki.dnb.de/pages/viewpage.action?pageId=124126254) that there exists a way to express this with [MADS RDF](http://www.loc.gov/standards/mads/rdf/v1.html) which would actually not look bad in our JSON-LD. For the example document http://lobid.org/resources/HT013099804 this would look like this (only writing down the interesting parts):

```json
{
   "@context":"http://lobid.org/download/context.json",
   "id":"http://lobid.org/resources/HT013099804#!",
   "subject":[
      {
         "type":"madsrdf:ComplexSubject",
         "madsrdf:componentList":{
            "@list":[
               {
                  "id":"http://d-nb.info/gnd/4032952-5",
                  "label":"Krefeld",
                  "type":[
                     "PlaceOrGeographicName"
                  ]
               },
               {
                  "id":"http://d-nb.info/gnd/4041156-4",
                  "label":"Naherholungsgebiet",
                  "type":[
                     "SubjectHeading"
                  ]
               },
               {
                  "id":"http://d-nb.info/gnd/4125082-5",
                  "label":"Freizeitverhalten",
                  "type":[
                     "SubjectHeading"
                  ]
               },
               {
                  "id":"http://d-nb.info/gnd/4061619-8",
                  "label":"Umweltbelastung",
                  "type":[
                     "SubjectHeading"
                  ]
               }
            ]
         }
      },
      {
         "type":"madsrdf:ComplexSubject",
         "madsrdf:componentList":{
            "@list":[
               {
                  "id":"http://d-nb.info/gnd/4032952-5",
                  "label":"Krefeld",
                  "type":[
                     "PlaceOrGeographicName"
                  ]
               },
               {
                  "id":"http://d-nb.info/gnd/4041156-4",
                  "label":"Naherholungsgebiet",
[...]

PR #462, opened 2017-06-06
Remove subjectChain as this is already covered

dr0i:
With 'componentList' the chained subjects are already covered.
This 'subjectChain' is superfluous.

Complements #187.


dr0i:

Issue #454, opened 2017-06-01
Provide content negotiation on individual resources 


jschnasse:
At the moment [lobid-api v2](https://lobid.org/resources) only provides JSON-LD and HTML for individual resources.

    curl http://lobid.org/resources/HT018472857 -H"accept:text/html"
    curl http://lobid.org/resources/HT018472857 -H"accept:application/json"

It would be nice to have provided more serializations like `rdf+xml`, `turtle`, `ntriples` ...  You could also consider to [support links for the different formats](https://www.mnot.net/blog/2012/10/29/NO_OPTIONS) and make them discoverable as link headers via `HTTP HEAD`.

fsteeg:
For consistency, this will require the same functionality for lobid-organisations.

fsteeg:
Deployed to stage:

```
curl --header "Accept: application/rdf+xml" http://stage.lobid.org/resources/HT018472857
curl --header "Accept: text/turtle" http://stage.lobid.org/resources/HT018472857
curl --header "Accept: application/n-triples" http://stage.lobid.org/resources/HT018472857
```

http://stage.lobid.org/resources/HT018472857.rdf
http://stage.lobid.org/resources/HT018472857.ttl
http://stage.lobid.org/resources/HT018472857.nt

After another talk with @jschnasse I went back to the original plan to only support this for single resources, not for `/search` requests (unlike what I mentioned in our call today).

Documentation at http://stage.lobid.org/resources/api#jsonld

acka47:
Just took a quick look, will review thoroughly tomorrow. However, we should add namespaces to make the turtle and probably also the RDF/XML better readable.

fsteeg:
I've set up Jena to use its extended built-in namespace prefixes and redeployed:

http://stage.lobid.org/resources/HT018472857.ttl

I don't think we should add additional, custom prefixes for this, since we'd have to maintain them in parallel to our context. For the RDF serializations we should be careful not to add additional maintenance overhead. We should restrict ourselves to what we get for free using the RDF tools.

This leads to the RDF/XML namespaces you mentioned. I guess you mean the attribute values (the elements are all in the `rdf` namespace)? Jena does not change anything about these with the built-in prefixes I enabled. In general, I don't think we should worry about (human) readability of RDF/XML.

jschnasse:
[...]

PR #465, opened 2017-06-07
Support RDF content types in single document GET requests

fsteeg:
Will resolve https://github.com/hbz/lobid-resources/issues/454


dr0i:
+1

PR #469, opened 2017-06-09
Add example how to make statistics about hbz01

### LINKED ISSUE #466: Document instructions for analyzing coverage and content of MAB fields
In https://github.com/hbz/lobid-resources/issues/455#issuecomment-305762786 I wrote:

> However, I am not quite sure how 070 is actually filled. The data is hopefully added automatically. As there are some strange cases, we should probably ask the colleagues how it works.

Generally, I think we should run some analytics on the Aleph XML dump before adding or improving the data transformation. This ticket aims at enabling @ChristophEwertowski and myself to do so on the example of field 070 indicators `-`, `a`, and `b`. 

Results should be:

1. information on the coverage of the fields in the hbz01 data (like "`070-`: 100%, `070a`: 77,3% etc.`
2. a list of different values in each field indicating their occurrence

dr0i:
After "mvn clean install", run like this:
`$ mvn exec:java -Dexec.mainClass="org.lobid.resources.run.MakeStatisticsMabXml" -Dexec.args="src/test/resources/hbz01XmlClobs.tar.bz2"`.
Adapt morph-makeHbz01Statistics.xml till it suits you well.

Resolves #466.

### DIFF
--- a/src/main/java/org/lobid/resources/Stats.java
+++ b/src/main/java/org/lobid/resources/Stats.java
@@ -4,7 +4,6 @@ import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Calendar;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


dr0i:
Added sum of values of fields and counting records which miss a field completely. See morph for examples.

ChristophEwertowski:
I don't get the example in line 14. Shouldn't it lead to the counting of all resources with 001 (basically all)? While trying I only got the 138 from the next "query". A few minutes of dabbling with it lead to no different result.

dr0i:
Yes, the `001??.?` is too broad to identify one and only one resource. Better would be `001-[-1].a`. However, there are other ways to get the number of the processed records, which I am implementing right now. Will assign me and reassign you when it's ready.

dr0i:
Force pushed a newer version where basic logik for generating statistics are put into the program rather than in the morph. Thus, the morph is simplified. And the coupling of morph and the statistic program is eased somewhat.

Fields are separated by a blank line. Now how to interpret this:

    |Field: 070-?.?|  //  field name - look at the morph's "name" (a possible "log: " prefix is removed)
    At least once in a record: 124/124 <=> 100%|  // how often it appeared at least once in relation to the processed records
    Frequency of iterated fields: 
    110 x "1", // so there are 110 records where the field appears one time ...
    14 x "2"| // ... and 14 records where the field appears two times.
    Frequency and value of field:
    16 x "6", // 16 times a value of 6 appears ...
    14 x "61", ... you got it :)
     ....

dr0i:
This is the result - took 80 minutes: http://stats.lobid.org/statistics-hbz01-070.csv.
Note: that's not proper csv, or tsv or whatever. I find this structure to be easy to read, but you might want some other for of serialization. Or is this just fine ?

acka47:
I think it is quite fine for now with one suggestion for improvement: When you have tab-separated values instead of the `x` separation, we could easily copy this into a file for tabular display. For example, instead of `4193434 x "HBZ/Off", ` you'd have `4193434 "HBZ/Off",`

dr0i:
@acka47 substitutes the 'x' with a tabulator as you suggested.

fsteeg:
+1

ChristophEwertowski:
Note for using it with LibreOffice: Remove the tabulator as a dividing element while opening the statistics.csv. Otherwise the description texts are mixed up with the contents of the MAB fields and the quantities of them.

PR #473, opened 2017-06-19
Delete all indices without alias

### LINKED ISSUE #472: Remove old index automatically
At the moment ~~up to at least 4~~ indefinetely indices can exist in parallel , which is more than we have space on the hd. So there should just always be ~~one index~~ indices with an alias while a new one is build (which will, when build, result in having (normally) two indices: 'resources' and 'resources-staging'. If other alias' exist these will not be removed!).

dr0i:
The method "aliases(String name)" is called two times. While the first call,
used by renaming aliases, is working fine, the second call (deleting old
indices) is not working. The logs showed:
"[WARNING] thread Thread[elasticsearch[Electron][scheduler]"
resulting in an java.lang.IllegalThreadStateException.
DEBUG level shows that NIO is involved:
("Selector.select() returned prematurely because the I/O thread has been interrupted. Use shutdown() to shut the NioSelector down).
However, checking if an alias is set for a given index is what is needed.
This can also be achieved by:
"getMetaData().hasAliases(new String[] { "*" }, new String[] { indexToDelete })"
where two string arrays are the parameters. There is also a method
"hasAlias(String name)" not needing that array, but then this method don't allow
wildcards so ".hasAliases()" is more convenient.

Resolves #472.

### DIFF
--- a/src/main/java/org/lobid/resources/ElasticsearchIndexer.java
+++ b/src/main/java/org/lobid/resources/ElasticsearchIndexer.java
@@ -297,11 +297,16 @@ public class ElasticsearchIndexer
 		if (allIndices.size() >= 3) {
 			final List<String> list = new ArrayList<>(allIndices);
 			list.remove(name);
-			for (String indexToDelete : list.subList(0, list.size() - 2)) {
-				if (aliases(indexToDelete).isEmpty()) {
-					LOG.info("Deleting index: " + indexToDelete);
+			for (String indexToDelete : list) {
+				boolean hasAlias = client.admin().cluster()
+						.state(Requests.clusterStateRequest().nodes(true)
+								.indices(indexToDelete))
+						.actionGet().getState().getMetaData()
+						.hasAliases(new String[] { "*" }, new String[] { indexToDelete });
+				if (!hasAlias) {
 					client.admin().indices().delete(new DeleteIndexRequest(indexToDelete))
 							.actionGet();
+					LOG.info("Deleting index: " + indexToDelete);
 				}
 			}
 		}


fsteeg:
Thanks for the explanation on the implementation details, sounds reasonable.

One issue however: I think this will delete more indexes than the previous version intended to delete. `deleteOldIndices` is called after `removeOldAliases`, so removing all unaliased indexes in `deleteOldIndices` will only retain the latest index, hence `list.subList(0, list.size() - 2)` in the current code. Or did you intend to delete more?

dr0i:
Yes and no: it wil delete more indices *if* they have no alias. If you would give every index an alias , it would lead to have even more retained indices (in comparison to the old (working!) routine in Api1.0).

`list.subList(0, list.size() - 2)` ensures to have always two indices at the end, regardless if there was an alias set or not. Just to be double safe.

`removeOldAliases` however doesn't remove *all* old aliases, but only those which matches the newly set alias - so it's guaranteed to have an unambigious alias (one-to-one relation of index<->alias).

If you would want to make sure that an index will never be removed automatically, just give it an alias (and make sure to not set that alias again, because  -> `removeOldAliases` ).

fsteeg:
+1, sounds good.

Just to make sure there is no misunderstanding: you removed `list.subList(0, list.size() - 2)` right? Also it was combined with checking the alias, so the old intended result, in my understanding, was to remove all non-aliased indexes, except for the latest unaliased index (the one that used to be the production index before the current switch). If the aliases are set up in a way that avoids loosing the index we were just switching away from (which we might want to restore if we find an issue with the new one), then everything is good.

dr0i:
Uh, you are right I removed `list.subList(0, list.size() - 2)`. So no double-safety.

> except for the latest unaliased index (the one that used to be the production index before the current switch).

Ok, may be so. But as you assume correctly, our aliasing is set up in such a way that this is not happening anyway, because we deliberately name the alias of the new index differently from that of the productive one.

When we would want to implement automatically switching the productive index that old routine should be reimplemented again, I think.

dr0i:

PR #474, opened 2017-06-20
Separate superordinate componentList

### LINKED ISSUE #450: Subject chains of resource and superordiante resource should be separated (2.0)
A customer for edoweb recognised that in [HT015660188](http://lobid.org/resources/HT015660188) subjects are mingled into one componentList, regardless if they originate from the resource or the superordinate resource. A solution for this problem would be splitting the macro `subjectRswkId` into two macros (ind2="1"/ind2="2"). It concerns lobid 1.0 but it should be fixed in 2.0, too.

dr0i:
Resolves #450.


dr0i:
Deployed to production, see e.g. http://lobid.org/resources/HT004381366.json. Closed.

PR #476, opened 2017-06-20
Remove creation of statistics when transforming hbz01

### LINKED ISSUE #475: Remove creation of statistics when transforming hbz01
It is not longer necessary to make always a statistic about nwbib resources which are lacking the 078n (see hbz/nwbib/issues/179). Also, with resolving #466 the statistic file is getting bigger. So, we should remove the making of a statistic file when doing regular hbz01 transformations.

dr0i:
Resolves #475.


dr0i:
As discussed today and as this is a minor PR removing rather than adding something I am going to merge that into master without review.
No problems anticipated here.

Issue #480, opened 2017-06-21
Missing resource in v2


jschnasse:
http://lobid.org/resources/HT016987148
http://lobid.org/resource/HT016987148
http://193.30.112.134/F/?func=find-c&ccl_term=IDN%3DHT016987148

dr0i:
Indexed the resource manually after updating the morph in #481.
This ticket spawned #482 which promises to capture droppings more closely so that a fix of these documents will be done proactively.
Closing this issue.

PR #481, opened 2017-06-22
Reset subject URI anew for inCollection

dr0i:
Ensure to have the correct subject URI when adding data.
The issue arised because of an incorrect data transformation resulting in a
"inCollection.inCollection" structure. Elasticsearch logs:
"MapperParsingException[failed to parse [inCollection.inCollection]]; nested: IllegalArgumentException[unknown property [id]]" and then drops the resource.

See #480.


src/main/resources/morph-hbz01-to-lobid.xml
```diff
@@ -2126,6 +2126,7 @@
 		<!-- ########## classification -->
 		<!-- ####################### -->
 		<!--nwbib -->
+		<call-macro name="setBackRdfSubject" field="078n[-1].a"/>
```
ChristophEwertowski: Should be macro "setRdfSubject". It was set back four lines prior.

dr0i:
As this is a bug fix rather than a change of data structure it can be savely merged into master. Also, it was reviewed by @ChristophEwertowski .
To see the full effect a full indexing is needed.
At least, most of the missing documents reported in #483 could be indexed using this commit.
Re travis failing: that's a timeout. Travis seems to be not happy if a branch is too often forc pushed. However, locally and also at the server the build is successful.
Closing.

PR #489, opened 2017-06-27
Add language labels

### LINKED ISSUE #439: Missing languages

ChristophEwertowski:
Added language labels if at least 10 resources had that language.

Resolves #439

Assigning it to dr0i for deployment because the scale of the changes isn't visible in the test set. After the deployment I will check if the language label of "http://id.loc.gov/vocabulary/iso639-2/crp" is shown correctly.

### DIFF
--- a/src/main/resources/labels/labels.json
+++ b/src/main/resources/labels/labels.json
@@ -13592,8 +13592,32 @@
       }
    },
    {
+      "uri" : "http://id.loc.gov/vocabulary/iso639-2/abk",
+      "label" : "Abchasisch"
+   },
+   {
       "uri" : "http://id.loc.gov/vocabulary/iso639-2/afr",
       "label" : "Afrikaans"
+   },
+   {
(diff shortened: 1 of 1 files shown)


dr0i:
Deployed to staging. @ChristophEwertowski you can test using http://stage.lobid.org/...

ChristophEwertowski:
Looks good. Of 20.1 million hits there are only 11445 left which have a language id but no correct label (under a per mill, [query](http://stage.lobid.org/resources/search?q=language.label:%22http://id.loc.gov/vocabulary%22)). That seems good enough for me that other languages will be added at the wishes of the users.

dr0i:

PR #492, opened 2017-06-28
88 wrong contents in urn object

### LINKED ISSUE #88: URN being an URL => MapperParsingException => drop document
An`urn` beeing an `id` (while expecting a value) breaks ES document parsing with `MapperParsingException`. Thus dropping the doc in whole.
55 docs affected as of lobid-gaia.log.2016-07-02:
`grep MapperParsingException  lobid-gaia.log.2016-07-02 | grep urn | wc -l` : 55

Example: http://lobid.org/resources/HT018956414

ChristophEwertowski:
Resolves #88


ChristophEwertowski:
I had the phenomenon that the jsonld file of the bibliographic description was only added after changing the morph. The commit which should do it only added output and exemplary files and updated the file of all triples ([commit](https://github.com/hbz/lobid-resources/pull/492/commits/c3d0f00cabebe07be306649c9ef92f937b340749)).

dr0i:
Nice - this catches some more docs which where dropped, thus not indexed!

dr0i:

PR #505, opened 2017-07-05
Remove subjectChain from context

ChristophEwertowski:
See #187

Removes the leftover `subjectChain` from the context.


dr0i:

Issue #427, opened 2017-05-11
Use one property for relations to superordinated resources 


ChristophEwertowski:
As discussed today ```volumeIn.multiVolume``` contains similar information like ```inSeries``` (numbering, the id of the superordinated resource and a relation type). So they should be put together. 
[Example for volumeIn and inSeries](http://lobid.org/resources/HT002888677?format=json):

```
"volumeIn" : [ {
    "multiVolumeWork" : [ {
      "id" : "http://lobid.org/resources/HT002888668#!",
      "label" : "lobid Ressource"
    } ],
    "numbering" : "2",
    "type" : [ "MultiVolumeWorkRelation" ]
  } ]
```
and 
```
"inSeries" : [ {
    "numbering" : "73",
    "series" : [ {
      "id" : "http://lobid.org/resources/HT001236717#!",
      "label" : "lobid Ressource"
    } ],
    "type" : [ "SeriesRelation" ]
  } ],
```
should be something like:
```
"isPartOf" : [ {
    "numbering" : "2",
    "type" : "isPartOfRelation",
    "superordinated" : {
      "id" : "http://lobid.org/resources/HT002888668#!",
      "label" : "lobid Ressource"
    }
}, {
    "numbering" : "73",
    "type" : "isPartOfRelation",
    "superordinated" : {
      "id" : "http://lobid.org/resources/HT001236717#!",
      "label" : "lobid Ressource"
    }
} ]
```
If someone can think of a better term than ```isPartOf``` then we take that. Also there are cases where a journal volume is part of a periodical which at the moment can't be covered ([example](http://lobid.org/resources/HT015296862)).
There are also resources which don't have an id for their superordinated resource, see [example](http://lobid.org/resources/HT015894164?format=json).

fsteeg:
If we have that `type` field, we might want to use different types here to get the convenience of having that information without accessing the superordination, to display different labels, as we currently do, i.e. something like:

```
"isPartOf" : [ {
    "numbering" : "2",
    "type" : "inMultiVolumeWork",
    "superordinated" : {
      "id" : "http://lobid.org/resources/HT002888668#!",
      "label" : "lobid Ressource"
    }
}, {
    "numbering" : "73",
    "type" : "inSeries",
    "superordinated" : {
      "id" : "http://lobid.org/resources/HT001236717#!",
      "label" : "lobid Ressource"
    }
} ]

```

acka47:
[...]

PR #506, opened 2017-07-05
Remove volumeIn from context

ChristophEwertowski:
See #427 

(For example https://github.com/hbz/lobid-resources/issues/427#issuecomment-307040729)


dr0i:

PR #533, opened 2017-07-26
Add publication type statistic, also add HT018722247 and HT019072754

### LINKED ISSUE #531: Adding publication types "Statistics" and "Standards"
An edoweb customer wants the publication types "Statistics" and "Standards". From the Jira ticket:

> Bitte Aleph 051 mn bzw. sn unter Publikationstyp "Leitlinien / Normschriften" zusammenfassen, 051 ms bzw. ss unter "Statistiken"

I have to think / read about the most precise English term. Of course we should type resources with other "Erscheinungsformen" also as the same type.

ChristophEwertowski:
Resolves #531

HT018722247: Changed resource to statistic (MAB 051).
~~HT019072754: Changed resource to norm (MAB 051).~~ Updated to include HT001039253 instead which is already a "standard" typed resource. No resource with type `Standard` was found in the collection.
Also sorted context and morph a bit.


ChristophEwertowski:
It fails because of 
```
[error] Test tests.IndexIntegrationTest.testResultCount[publication.location:Berlin] failed: expected:<1[3]> but was:<1[5]>, took 0.02 sec
[error] Test tests.IndexIntegrationTest.testResultCount[publication.location:Berlin AND publication.startDate:[1992 TO 2017]] failed: expected:<1[1]> but was:<1[2]>, took 0.014 sec
```
@dr0i : We already had this kind of problem but I forgot the reason for it.

ChristophEwertowski:
I suppose the problem is the new publication type. Do you know how to adjust the elasticsearch @fsteeg ?

fsteeg:
It sounds like the resources you added increased the result count for the test queries.

If these make sense, you can simply increase the expected result counts in the tests.

PR #534, opened 2017-07-27
Add "RPB-Raumsystematik" as a source, remove it from other RPB subjects

### LINKED ISSUE #517: Add new source for Raumsystematik (rpbr)
At the moment URLs from the Raumsystematik of the "Rheinland-Pfälzische Bibliographie" aren't seperated from other entries of the Bibliography. Now the customer wants that the Raumsystematik in edoweb isn't shown. So we should add a new source so that in edoweb the other two sources ("Systematik der Rheinland-Pfälzischen Bibliographie" and the lbz-notation "RPB-Sachgruppen und Zusätze") are used and the Raumsystematik is ignored. According to the customer the notation of the Raumsystematik starts with "rpbr".
(Maybe we will also get the labels.)

ChristophEwertowski:
Resolves #517

### DIFF
--- a/src/main/resources/labels/context-labels.json
+++ b/src/main/resources/labels/context-labels.json
@@ -339,6 +339,14 @@
       "name" : "dateModified"
    },
    {
+      "container" : "@set",
+      "label" : "Notation",
+      "multilangLabel" : {},
+      "name": "notation",
+      "referenceType" : "String",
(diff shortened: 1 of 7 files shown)


dr0i:
Deployed to production, see e.g. http://lobid.org/resources/HT019149667.json. Closed.

Issue #498, opened 2017-06-29
Link to download page in data set description 


acka47:
We should somehow link to https://datahub.io/dataset/hbz_unioncatalog or directly http://lobid.org/download/dumps/DE-605/mabxml/. As distribution? Or something like "derived from"?

PR #535, opened 2017-07-27
Add link to source data in dataset description

acka47:
Fixes https://github.com/hbz/lobid-resources/issues/498


acka47:
It probably makes sense to also show this in the HTML...

PR #543, opened 2017-08-09
Change default for corporations with MAB 2?[02468][-a].9 to issuing body

ChristophEwertowski:
The label for issuing body is "Herausgeber/in" which fits better than "Autor/in".

Resolves https://github.com/hbz/lobid-resources-web/issues/37


ChristophEwertowski:
Forgot to closing the pull request. Doing it now.

PR #557, opened 2017-09-01
552 add living statistics not years

### LINKED ISSUE #552: Add living statistics which don't consist of year specifications
For example in [HT017260117](http://lobid.org/hbz01/HT017260117) MAB `902-1.d` or [HT018924091](http://lobid.org/hbz01/HT018924091) in `100-1.d`. Maybe adding it as `note`.

ChristophEwertowski:
Resolves #552 

Also adds not pushed commit which deals with `dateOfBirth` and `dateOfDeath` BC.


dr0i:
@ChristophEwertowski not sure if it's a good idea to prefix a non-4-digit number with so many 0 that you enfore a 4 digit number. Why was this necessary? I wouldn't bother. If elasticsearch _can_ be range-queried (say: "from: -0340 to: 0020") I think we are on the save side. Can you confirm this works?

ChristophEwertowski:
I stumbled over a resource with an agent with a date with a leading zero. Apparently I didn't let a statistics be made because then I would have found out that there are very few cases like this. 

At the moment we have a range query for publication dates (`q=issued:year-year`) which doesn't work like it should. For single fields you can search like this: `http://lobid.org/resources/search?q=contribution.agent.dateOfDeath:[* TO 200]`. This has it problems too, as this example only compares the first three digits and not the whole number. So resources from the 1900s are also included. For negative numbers it doesn't work either: `http://test.lobid.org/resources/search?q=contribution.agent.dateOfDeath:[* TO 0000]` doesn't include the example with `Pindaros` who was born "-0522". Also finding a specific resource with a negative date only works by putting the number into paraphrases. I will open tickets.

dr0i:
>  I will open tickets.

@ChristophEwertowski have you?

ChristophEwertowski:
I didn't until now: https://github.com/hbz/lobid-resources/issues/562. Thanks for reminding me.

dr0i:
@ChristophEwertowski Thx for opening :)

PR #551, opened 2017-08-17
Add dateOfBirth and dateOfDeath from 100s

ChristophEwertowski:
Resolves #528

PR #571, opened 2017-09-26
Remove mapping of MAB 551 to "PublishedScore"

acka47:
As MAB field 551 holds the "VERLAGS-, PRODUKTIONS- UND BESTELLNUMMER VON MUSIKALIEN UND TONTRAEGERN" (see [MAB documentation](http://www.dnb.de/SharedDocs/Downloads/DE/DNB/standardisierung/mabTitelBibliographischeDaten2001.txt?__blob=publicationFile)) resources with content in this field can be either PublishedScores or AV documents and thus you shouldn't automatically add the publication type "PublishedScore". See also https://jira.hbz-nrw.de/browse/FRL-275.

### DIFF
--- a/src/main/resources/morph-hbz01-to-lobid.xml
+++ b/src/main/resources/morph-hbz01-to-lobid.xml
@@ -737,7 +737,7 @@
 		<data source="334-1.a" name="@rdftype">
 			<regexp match="Musikdruck" format="http://purl.org/ontology/mo/PublishedScore"/>
 		</data>
-		<data source="541[-ab]?.?|551??.?" name="@rdftype">
+		<data source="541[-ab]?.?" name="@rdftype">
 			<regexp match=".*" format="http://purl.org/ontology/mo/PublishedScore"/>
 		</data>
 		<!-- amtliche Druckschrift -->
--- a/src/test/resources/hbz01.es.nt
+++ b/src/test/resources/hbz01.es.nt
@@ -7191,7 +7191,7 @@
 <http://lobid.org/resources/HT015090208#!> <http://purl.org/ontology/bibo/edition> "DVD-Fassung"^^<http://www.w3.org/2001/XMLSchema#string> .
 <http://lobid.org/resources/HT015090208#!> <http://schema.org/publication> _:Bb16 .
 <http://lobid.org/resources/HT015090208#!> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://purl.org/dc/terms/BibliographicResource> .
-<http://lobid.org/resources/HT015090208#!> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://purl.org/ontology/mo/PublishedScore> .
+<http://lobid.org/resources/HT015090208#!> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://purl.org/lobid/lv#Miscellaneous> .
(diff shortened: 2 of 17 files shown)


statistics.csv
```diff
@@ -0,0 +1,145 @@
+|*field*|*frequency*| facultative: different values sorted by occurrence , separated with commata*|
```
dr0i: Uh - this file wasn't supposed to be commited, yes?

acka47: hm, no.

acka47: I'll remove it

dr0i:
No need to revision an unwanted file, so It would be cleaner to rebase 92df5fa3d890ed00bff9195a2ac6603bba41186e with `git rebase -i 52cc315ab6a977e43431aabf0fe254b5c8377ed7`, delete `statistics.csv` and commit that, then force-push.

acka47:
I don't get it. Why is it so important to remove the file from history? Don't we use git to be able to make mistakes? Why should  hide them?

dr0i:
It doesn't show an "evolution" or improvement of something. It's just a mistake of a revision of a file which should not have revisioned in the first place.

dr0i:
As discussed on telephone with @acka47 we could discuss our git policy.

A note:
- status quo was: 3 commits
- I amended the commit with wrongly added file, force pushed. Now 2 commits.
- we could have also used `git revert` but would then have an extra commit. So there would be 4 commits.

In general I acknowledge that sometimes `git rebase` can be:
- hard to do when hidden deep in the commit chain (because of (subsequent) conflicts)
- annoying because the PR discusses something that doesn't show up in the commit history
- people working on the branch don't have resetted their branch so this will give conflicts

This PR here IMHO is a paradigmatically good example when it makes sense to use `git rebase` even to an already opened PR (i.e. pusblihed branch).

acka47:
Generally, I have the feeling that we too often force push to GitHub into a branch that already has an open PR. But rewriting public history should only happen in very limited exceptions. I think one should rather make a new commit to fix something than using `--amend` and `push -f`. I don't see a problem of having too many commits. Let's discuss this in person next Thursday or so.

dr0i:
Uh - oh.
"Deeper" working with this branch I discovered two more "issues":
1. the branch name doesn't follow our convention (it should start with `571-`, not `551`). As a result, when deploying by merging and pushing, I would wrote "Resolves #551" which is a totally different issue. That different issue would then be closed.)
2. To start with, it is good common sense to create commits which are content with themselves. That is, your first commit adapting the morph should also show the outcome of that adapting by adding the test results into that commit. Now we have two commits which better would be one. So I will squash them to one and force push them here.

acka47:
Re. 1.) You are right. There is no corresponding issue and I shouldn't have started the branch name with a number (it is the field "551" in this case) which makes it confusing.

Re 2.), this is exactly the behaviour I don't understand. You will find several other cases in different repos where the tests were updated after the morph and we haven't had any problems with them yet. You are probably right that this is bad practice but nonetheless I don't see why there is a need to rewrite the public history here.

dr0i:
"How to change *bad practices*?" is the question then.

acka47:
...one of the bad practices being force pushing into a pull request on GitHub because someone has not followed good practice...

dr0i:
I just wanted to help. Why, you may be right.
Can you please open a new Pull Request since this PR is bad practice?
I will close this one.

acka47:
You don't have to be offended. However, this may be one of the ways to deal with such things. Although it means more work we will have a cleaner commit history...

dr0i:
> Although it means more work 

That depends on the perpective taken:
for you it means more work, for me it's much lesser.

Issue #579, opened 2017-10-02
Accelarate geodata lookups


dr0i:
Enrichment with geodata (#130) has decelarated tranformations considerably, around +12h (<=> ~+40%). This should be improved.

dr0i:
Ack re improvement: now it sums up to +1h.

dr0i:
> From 295,459 resources with content in the coverage field only 17,950 are missing the spatial field. 
(Source: https://github.com/hbz/lobid-resources/issues/130#issuecomment-332170896)

This has changed due to this issue to +30k <=> factor 3 => 18k + 30k = 48k <=> 25%.
The cause:
The 30k were coming from a fallback-API-query where the first hit is taken regardless if its score would be high enough. Now the `score` is more often taken into acount for these fallback queries. As a result the accuracy of the spatial data has improved while the quantity of the enrichment has decreased.

PR #580, opened 2017-10-02
Avoid unnecessary fallback-lookups

dr0i:
When a fallback geodata-lookup was done once for a query the lookup
should not be done for the same query again being in the same transformtion
session, because when the query fails at the first time it will most probably
fail subsequently.

This does not only prevent repeating lookups of failed queries but also of
successfully queries which were stored in the geodata-elasticsearch-index
but which scores not enough so that the fallback lookup would be done
repeatedly.

See #579.


fsteeg:
+1

PR #581, opened 2017-10-02
Add dewey labels and TT050427581, remove wrong notations and URLs

### LINKED ISSUE #400: Show complete DDC
Only the first class is shown but not the subClasses, see http://lobid.org/resources/HT019301219?format=json. The reason is a wrong regular expression.

[Source](http://lobid.org/hbz01/HT019301219) (snippet):

```xml
<datafield tag="700" ind1="b" ind2="1">
   <subfield code="a">914.3</subfield>
</datafield>
```

Current JSON-LD: 

```json
{
   "subject":[
      {
         "id":"http://dewey.info/class/914/",
         "label":"Geografie Europas und Reisen in Europa",
         "source":{
            "id":"http://d-nb.info/gnd/4149423-4",
            "label":"Dewey-Dezimalklassifikation"
         }
      }
   ]
}
```
Also dewey.info seems to be down. I only get an error 500 (internal server error).

ChristophEwertowski:
TT050427581: Resource with MAB 705

Resolves #400


src/test/resources/jsonld/TT050427581
```diff
@@ -0,0 +1,236 @@
+{
```
dr0i: Is it really necessary to add this test resource?

src/test/resources/output/json/05042/TT050427581.json
```diff
@@ -0,0 +1,256 @@
+{
```
dr0i: This resource, is it really needed?

ChristophEwertowski:
Removed the superflous resource TT050427581 because I overlooked HT015455455 which also has a MAB 705.

src/main/resources/morph-hbz01-to-lobid.xml
```diff
@@ -1654,10 +1672,9 @@
 		<!-- #######subjects -->
 		<!-- ####################### -->
 		<!-- ddc links -->
-		<data source="70[05][-b][-12345].a" name="@deweyId">
-			<regexp match="([0-9]{3})" format="http://dewey.info/class/${1}/"/>
-			<sanitizeUrl></sanitizeUrl>
-		</data>
+		<call-macro name="addDeweyNotationAndLabel" field="700[-b][-12345].a"/>
+		<call-macro name="addDeweyNotationAndLabel" field="705?[-12345].a"/>
+		<call-macro name="addDeweyNotationAndLabel" field="705?[-12345].c"/>
```
dr0i: Why not making one rule instead of two? I think this should work fine: `field="705?[-12345].[ac]"/>`

ChristophEwertowski: I tried this first but then in http://lobid.org/hbz01/HT015455455 subfield a is taken for the notation of subfield c (surprisingly we have no label for `338.76629220904`).

ChristophEwertowski:
Multiple packages aren't included at the moment. Removing the review request.

src/main/resources/morph-hbz01-to-lobid.xml
dr0i: Oh. Ok then +1.

dr0i:
Ok, but you could have just assigned me, no? I have now positively reviewed your PR, assigned myself and going to deploy.

ChristophEwertowski:
I had left a comment about also adding the "Grundnotation" in addition to "Vollständige Notation" and waited for a sign that acka47 had read it. That change itself was foolproof so you could assign it to yourself (otherwise I would have written that approval would be necessary).

dr0i:
Deployed to production,closing.

PR #582, opened 2017-10-02
Remove mapping of MAB 551 to "PublishedScore"

acka47:
Here's the next PR for this as we had some problems in the first round (see https://github.com/hbz/lobid-resources/pull/571). This also doesn't have an issue associated with it but I think this is ok for this minor change.

As MAB field 551 holds the "VERLAGS-, PRODUKTIONS- UND BESTELLNUMMER VON MUSIKALIEN UND TONTRAEGERN" (see [MAB documentation](http://www.dnb.de/SharedDocs/Downloads/DE/DNB/standardisierung/mabTitelBibliographischeDaten2001.txt?__blob=publicationFile)) resources with content in this field can be either PublishedScores or AV documents and thus you shouldn't automatically add the publication type "PublishedScore". See also https://jira.hbz-nrw.de/browse/FRL-275. Includes an update of the test data.

### DIFF
--- a/src/main/resources/morph-hbz01-to-lobid.xml
+++ b/src/main/resources/morph-hbz01-to-lobid.xml
@@ -737,7 +737,7 @@
 		<data source="334-1.a" name="@rdftype">
 			<regexp match="Musikdruck" format="http://purl.org/ontology/mo/PublishedScore"/>
 		</data>
-		<data source="541[-ab]?.?|551??.?" name="@rdftype">
+		<data source="541[-ab]?.?" name="@rdftype">
 			<regexp match=".*" format="http://purl.org/ontology/mo/PublishedScore"/>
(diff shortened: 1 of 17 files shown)


acka47:
I don't understand why Travis is complaining. My local build is successful.

dr0i:
You did everything fine. Travis had had some trouble with writing logs - restarting travis resulted in succesful checks.

dr0i:
Deploy went bad for the fulldump wasn't ready when transformation started. Getting the fulldump is not very realiable predictable - between 4 h and 15 h.
Scheduled the programs so that the dumping can take ~20h to be on the save side.

dr0i:
Deployed to production, see e.g. https://lobid.org/resources/HT018939763.json.
Closing.

PR #587, opened 2017-10-12
Fix regex

acka47:
Resolves https://github.com/hbz/nwbib/issues/400

As discussed offline, I refrained from adding a test resource for this case as we otherwise would have to add a test resource for each notation...


dr0i:
Deployed to production,closing.

PR #604, opened 2017-10-24
Make geo data elasticsearch compatible

### LINKED ISSUE #601: spatial.geo should be type: geo_point
To support geo searches from NWBib, the `spatial.geo` field needs to be `type: geo_point`.

See https://www.elastic.co/guide/en/elasticsearch/reference/2.4/geo-point.html

The current data however is not compatible due to its additional `type` field:

```
spatial": [{
        "geo": {
            "lon": 6.36889,
            "type": "GeoCoordinates",
            "lat": 50.9839
        },
        "id": "http://www.wikidata.org/entity/Q1959200",
        "label": "Müntz (Titz)",
        "type": "Location"
}]
```

With `type: geo_point` I get:

`[error] application - Bulk item response failed for index 'resources', ID 'BT000041593', message: MapperParsingException[failed to parse]; nested: ElasticsearchParseException[field must be either [lat], [lon] or [geohash]];`

I see three options.

In API 1.x, we had a simple string field, so this would be most straight-forward to adopt for clients:

```
spatial": [{
        "geo": "50.9839,6.36889",
        "id": "http://www.wikidata.org/entity/Q1959200",
        "label": "Müntz (Titz)",
        "type": "Location"
}]
```

This would be a API change, but the geo stuff has not been on production for long, and has not been officially announced or documented, so I think we could do this change.

Another option is to just remove the type field:

```
spatial": [{
        "geo": {
            "lon": 6.36889,
            "lat": 50.9839
        },
        "id": "http://www.wikidata.org/entity/Q1959200",
        "label": "Müntz (Titz)",
        "type": "Location"
}]
```

And finally, we could add an additional simple field (if we need the `GeoCoordinates` info):

```
spatial": [{
        "geo": {
            "lon": 6.36889,
            "type": "GeoCoordinates",
            "lat": 50.9839
        },
        "point":  "50.9839,6.36889",
        "id": "http://www.wikidata.org/entity/Q1959200",
        "label": "Müntz (Titz)",
        "type": "Location"
}]
```

I like the first option best. @acka47 @dr0i what do you think?

dr0i:
- remove type "GeoCoordinates"
- adapt index-config

Resolves #601.

### DIFF
--- a/src/main/java/org/lobid/resources/run/WikidataGeodata2Es.java
+++ b/src/main/java/org/lobid/resources/run/WikidataGeodata2Es.java
@@ -249,7 +249,6 @@ public class WikidataGeodata2Es {
 				if (!geoNode.isMissingNode()
 						&& !geoNode.findPath("latitude").isMissingNode()) {
 					spatial.set("geo", geo);
-					geo.put("type", "GeoCoordinates");
 					geo.put("lat", geoNode.findPath("latitude").asDouble(0.0));
 					geo.put("lon", geoNode.findPath("longitude").asDouble(0.0));
 				} else {
--- a/src/main/resources/index-config-wd-geodata.json
+++ b/src/main/resources/index-config-wd-geodata.json
@@ -13,18 +13,7 @@
                         "type" : "string"
                      },
                      "geo" : {
-                        "properties" : {
-                           "type" : {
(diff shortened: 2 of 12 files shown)


dr0i:

Issue #585, opened 2017-10-09
Improve wikidata matching


acka47:
Follow-up to #130. Improve the results of the wikidata matching.

Here is an overview over the current status ([source](https://wiki1.hbz-nrw.de/display/SEM/NWBib-Treffen+zur+Neukonzeption+der+Systematik+05.10.2017#NWBib-TreffenzurNeukonzeptionderSystematik05.10.2017-Nachtrag%3AAktuelleMatchingstatistiken)): 

- Wievielen Titeln mit Ortsnennung konnten keine Wikidata-Id zugeordnet werden?
  - 51721
- Wieviele distinkte Orte sind das? 
  - 3128 
  - [Liste](http://stats.lobid.org/fehlendeDistinkteNWBibOrteInWikiData.txt)
- Wieviele Orte habe eine WD-Id mit einem score >4 bekommen? 
  - 259.456
- Wieviele distinkte Orte sind das? 
  - 4975 
  - [Liste](http://stats.lobid.org/successOrteSortUnique.txt) (mit Kontext)
- Bei wievielen gab es eine erfolgreiche Fallback-Anfrage? 
  - 2980
- Wieviele distinkte Orte sind das?
  - 2966 
  - [Liste](http://stats.lobid.org/successFallbackUnique.txt)

acka47:
As discussed offline:

- We won't query the Wikidata API anymore as fallback because all resources we need should be in the set covered by the SPARQL query. If they are not in there, something has to be corrected at Wikidata. (With this, we also won't have distinct Wikidata sets for the lobid data and the visualized classification entry.)
- We will add the label from the superordinated administrative (`P131.rdfs:label""@de` ) to the ES geo index and this will lead to matches for a lot of strings in currently not matched (e.g. "Ahaus Ottenstein" & [Q615951](https://www.wikidata.org/wiki/Q615951).

dr0i:
Not having enough time left yesterday I did the following:
- build a new geo-index with only SPARQL-query entities [1]
- remove fallback
- reduce minimum score threshold to a very low score

Resources are indexing at the moment.

[1] Here is a list of WD-entities lacking geo-coordinates: https://gist.github.com/dr0i/e9bb9e1083a56ebaef602ec60991c458#file-missinggeocoordswikidata

acka47:
[...]

PR #609, opened 2017-10-30
Tweak scoring of field "alias"

dr0i:
See #585.


dr0i:
Deployed to production.
@acka47 Note that this is not a direct cause of this PR but : I checked http://etherpad.lobid.org/p/geo-testset now everything seems to be OK (not the bad catalog entry "Neu-Ehrenfeld" of course).

PR #613, opened 2017-11-06
Remove `<` and `>` from bnode in 700n; allow multiple primaryForm

### LINKED ISSUE #612: Dropped resources: '<''>' in bnodes and multiple primaryForm-entries
Looking at a new fulldump-log around a dozen resources are dropped. The cause is

- &gt; and &lt; in  700n when creating bnodes from this
- `primaryForm` may be as set instead of a simple container

dr0i:
- catch Exception in JsonConverter so that only the field is dropped, not
	the whole resource

Resolves #612.


ChristophEwertowski:
The `>test;<` from src/test/resources/jsonld/HT018924091 or the input file should be removed. Otherwise O.K.

dr0i:
@ChristophEwertowski ah, I deliberately added `>test;<` into the test file to not be forced to add a whole new resource.

dr0i:
Deployed to production, see e.g. http://lobid.org/resources/BT000109478.json appears now.. Closing.

PR #617, opened 2017-11-09
Add geo properties to context

### LINKED ISSUE #616: geo, lat, lon missing in JSON-LD context

acka47:
Fixes #616

### DIFF
--- a/src/main/resources/labels/context-labels.json
+++ b/src/main/resources/labels/context-labels.json
@@ -355,6 +355,27 @@
       "multilangLabel" : {}
    },
    {
+      "name" : "geo",
+      "multilangLabel" : {},
+      "label" : "Geodkoordinaten",
+      "uri" : "http://schema.org/geo",
+      "referenceType" : "@id"
+   },
+   {
+      "name" : "lat",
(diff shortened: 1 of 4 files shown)


acka47:
Strange, tests run with no problems on my local machine.

dr0i:
The problem with defining the geo location as "String" is that in the elasticsearch index this field (https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-point.html) is stored as a a "json Number" (https://www.tutorialspoint.com/json/json_data_types.htm). As we present elasticsearch "_source" field has the json-ld-output to users directly, avoiding retransformation, the most simple way to be coherent of what the "data looks like" versus "what the data is defined to look like" would be to define the geo coordinates as "number" (aka "double") in our context. @acka47 ?

dr0i:
@acka47 forget my last remark: as only the "context.json" is presented to user (not the lables-context.json) it's satisfying to fullfill the boundaries of http://schema.org/latitude . Which we do (

>  Values expected to be one of these types:
>  Number or  Text

dr0i:

Issue #618, opened 2017-11-10
Remove duplicate entries from JSON-LD context


acka47:
As @dr0i noted that there is a double entry in context-labels.json (see https://github.com/hbz/lobid-resources/issues/616#issuecomment-343428363) I thought we should have a look whether there are other double entries.

acka47:
There are two things to check:

1. Whether there is a double JSON key or
2. whether there is a double occurence of a URI

Here is how to create a list of json keys ("name") and URIs ("uri) and their occurence:

- `cat context-labels.json | jq .[].name | sort | uniq -c | sort -n -r`
- `cat context-labels.json | jq .[].uri | sort | uniq -c | sort -n -r`

One can see that there are two double URI entries:
```
$ cat context-labels.json | jq .[].uri | sort | uniq -c | sort -n -r
      2 "http://rdaregistry.info/Elements/u/P60493"
      2 "http://rdaregistry.info/Elements/u/P60345"
      ...
```

Also, there are two entries with name "type":

```
$ cat context-labels.json | jq .[].name | sort | uniq -c | sort -n -r
      2 "type"
      2 "otherTitleInformation"
```

I woll go forward and fix this.

acka47:
Re. "type", I will remove the mapping to "http://www.w3.org/1999/02/22-rdf-syntax-ns#type". It can't be found in the current context anyway.

acka47:
Re. the double "type" key see https://github.com/hbz/lobid-resources/pull/619#issuecomment-343473404

PR #619, opened 2017-11-10
Remove redundant entries in context-labels.json

acka47:
Fixes https://github.com/hbz/lobid-resources/issues/618

### DIFF
--- a/src/main/resources/labels/context-labels.json
+++ b/src/main/resources/labels/context-labels.json
@@ -623,22 +623,6 @@
       "name" : "heldBy"
    },
    {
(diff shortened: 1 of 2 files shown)


acka47:
I didn't remove the "type" entry as it seems to be important for conversion to JSON. When I removed the http://www.w3.org/1999/02/22-rdf-syntax-ns#type part and ran `mvn clean install` all the type information was lost from the test files. (I vaguely remember a discussion on this somewhere but couldn't find it on GitHub.)

dr0i:

Issue #572, opened 2017-09-28
Unthrottled search requests result in errors


fsteeg:
After about 4000 unthrottled search requests the API starts returning errors.

Will set up a scenario for reproduction and add intructions here.

fsteeg:
To reproduce, get NWBib, checkout `392-wikidata`, `run` and visit: 

http://localhost:9000/classification?t=Wikidata

Errors start at some point after request 4000 (see log).

This uses http://stage.lobid.org/resources with changes from `572-errors`.

fsteeg:
Sorry, forgot to mention the crucial step to actually get the errors: disable the client side throttling by commenting out this line: https://github.com/hbz/nwbib/blob/392-wikidata/app/controllers/nwbib/WikidataLocations.java#L115

dr0i:
I got now error, although outcommented the sleep. Using VPN 4500 requests take 12m. Can imagine that that is slow enough in itself. Going to use it directly on our es-server.

dr0i:
Omitting VPN and use a direct external connection indeed quickens the querying by factor 2 ( so now takes 6m ) but is without error, too. So the step to reproduce seems to be to start the web-app as near to the index as possible.

dr0i:
Running on the same server the *Request* is again faster,  now it takes ~1m for 1k. And the `error` now indeed occurs. Web-app logs: 

> Could not query index: None of the configured nodes are available: [{#transport#-1}{193.30.112.84}{aither.hbz-nrw.de/193.30.112.84:9300}, {#transport#-2}{193.30.112.82}{gaia.hbz-nrw.de/193.30.112.82:9300}]

Since the `java` versions where slightly different I synced them. Also a `memlock=unlimited` was not configured on aither. Updated all, made sure it's in sync, restarted. Errors again. My suspicion now: `es` must also be restarted on gaia (which I didn't do because it may be hardwired into webapps). Will try this tomorrow.

dr0i:
Restarted `es@gaia` to no avail. As throttling is a workaround this has to suffice at the moment. Posponing this issue.

dr0i:
We got too much of the `None of the configured nodes are available:` errors to ignore them. I wasn't able to track this down - network seems to be ok. So maybe it is the mix of `Client`  and `TransportClient`. Haven't figured that out yet, and there are some ES-patches regarding this. I propose to switch to:
1. newest ES
2. weywot[34] as new cluster
and see how things fugured out. What do you think @fsteeg ?

[...]

PR #621, opened 2017-11-14
572 update es to 5.6.3

dr0i:
Updates elasticsearch from 2.3 to 5.6.3.

Resolves (hopefully) #572.


dr0i:
@fsteeg indexing is working albeit not via VPN yet - try e.g. `curl -L "http://weywot5.hbz-nrw.de:9200/resources-staging/resource/HT019336970"`.
The web api is not deployed yet on weywot5 because of an `NoNodeAvailableException` log entry ;)

I have updated the web api as far as I could, to the point that some `scala` tests not running ok and having trouble to analyze this. Maybe you want to have a look.

fsteeg:
I have tweaked the Scala view code to skip missing items. The remaining failing test exposes the first actual issue: items are not found (second test in `ResponseStatusTests`: expected:<[200]> but was:<[404]>). The tests use a `NodeClient` with the embedded test index, which basically seems to work (except for the items issue).

The second issue is that the `TransportClient` (not used by the tests, used by the actual UI) does not work at all (as you mentioned: `NoNodeAvailableException`). It was still trying to connect to the ES 2 cluster on gaia and aither. I updated the config to use weywot5. It still does not work, I think because port `9300` is blocked (on weywot5? On the entire 10.x net?) `http://weywot5.hbz-nrw.de:9300/` is not accessible (Unable to connect), while `http://weywot5.hbz-nrw.de:9200/` works fine. On the old cluster, I do get a response for `http://gaia.hbz-nrw.de:9300/`, saying `This is not a HTTP port`. All of this from within the hbz, not via VPN.

Since this is now all back in Java land, can I hand this back to you at this point @dr0i?

dr0i:
Yep - thx!

dr0i:
Deployed to weyowt5, see e.g. `http://weywot5.hbz-nrw.de:7500/resources`.

dr0i:
Interestingly while the lookup-test of `TT003059252:DE-5-58:9%2F041` fails the real-life lookup works: http://weywot5.lobid.org/items/TT003059252:DE-5-58:9%2F041

dr0i:
Re my last comment: I am clueless why its working in production but not in the tests. Assigning @fsteeg.

fsteeg:
I have no idea why, but for some reason the GetRequest does not work on the embedded test index. Using an IdsQuery works fine on both the production and the test index. I guess the GetRequest does some low-level stuff (unlike the IdsRequest, it requires the parent to be set) that does not work with our item IDs on the embedded index. So I replaced the GetRequest with an IdsQuery, see commit above.

dr0i:
@fsteeg thx a lot!

dr0i:
Some notes for the configuration in `elasticsearch.yml`:
- set `discovery.zen.minimum_master_nodes: 2` in . This prevents the possibility of a "split brain" by refusing indexing if not all master nodes (aka the cluster) are up. It doesn't prevent querying if a node is missing, though, which is good. 
-  set `bootstrap.memory_lock: true`  . For this to work it's crucial to set the `ulimit -l unlimited` at  OS-level.

dr0i:
Turns out ES 5.6.3 is really slow with our data. Even when setting `mem` to 28 GB. Space one index uses with ES 2.3: 71 GB. With ES 5.6.3 : 78 GB. So there is slight (10%) more consumption.
I have the hope that the index-mapping is not well configured. That can certainly be adjusted. Also, looking at the mapping, there are a lot of fields indexed which are not meant to be there in the first place. This was also true for ES 2.3, however. Solution would be to somehow ensure well docs beeing indexed. Could imagine a post-processing which would punch the data through the `inde-config.json` so that all unexpected data would be erased.

fsteeg:
Today weywot5 was having a high load (>12) and was responding very slowly. The stage web application, which is configured to use the weywot cluster was not responding when searching. I shut down elasticsearch on weywot5 as a first step, making the cluster (now only weywot4) responsive again.

I then took a look into the configs in `/etc/elasticsearch/elasticsearch.yml` and noticed this section:

```
# Make sure that the heap size is set to about half the memory available
# on the system and that the owner of the process is allowed to use this
# limit.
#
# Elasticsearch performs poorly when the system is swapping the memory.
```

So in `/etc/elasticsearch/jvm.options`, I've set  the heap to 16g on both weywot4 and weywot5, and restarted elasticsearch on both.

But weywot4 did not start up successfully, logging to `/var/log/elasticsearch/weywot.log`:

`memory locking requested for elasticsearch process but memory is not locked`

From reading [1] I concluded the user's memlock settings are the issue. I applied the suggested `limits.conf` settings from [2] on weywot4, restarted the machine, and elasticsearch is running. I think it's worth to try some more of the settings suggested in [2], and to apply them both on weywot4 and weywot5.

[1] https://www.elastic.co/guide/en/elasticsearch/reference/current/_memory_lock_check.html
[2] https://ozzimpact.github.io/development/elasticsearch-configuration-tuning

dr0i:
This issue depends on #643.

dr0i:
Re: limit
> but, other than on weywot5, this must be done every time ES is restarted. Haven't figured that out yet.

I set `MAX_LOCKED_MEMORY=unlimited` in `/etc/init.d/elasticsearch`. This should also work.  To be tested when index is buildt tomorrow.

dr0i:
Re: ulimited memlock
> To be tested when index is buildt tomorrow.

Tested. Works.

Issue #622, opened 2017-11-16
Improve SPARQL query for getting administrative areas in NRW


acka47:
Currently, we use something like this for building the geo index, respectively for creating the classification at https://test.nwbib.de/classification?t=Wikidata: http://tinyurl.com/y8rpexfm

- This queries for things that are sub*classes of wd:Q387917 (Verwaltungseinheit in Deutschland) and are locatedIn* NRW.
- It also queries for things that are geographic regions, e.g.  Münsterland, Südwestfalen etc.

We could improve this:
- The geographic regions bit could be removed imo.
- We might rather use a union query for filtering the types using the following:
   - [Q22721](https://www.wikidata.org/wiki/Q22721): Regierungsbezirk
   - [Q106658](https://www.wikidata.org/wiki/Q106658): Landkreis in Deutschland
   - [Q22865](https://www.wikidata.org/wiki/Q22865): kreisfreie Stadt in Deutschland
   - ~[Q515](https://www.wikidata.org/wiki/Q515): Stadt (Maybe we even don't need this.)~
   - [Q262166](https://www.wikidata.org/wiki/Q262166): Gemeinde in Deutschland
   - [Q253019](https://www.wikidata.org/wiki/Q253019): Ortsteil

acka47:
Here is the query as described above: http://tinyurl.com/ya4kzm8l 

@fsteeg Please try this out for creating the classification. Just comment out lines 20-21 for creating the classification. There should be far less entries under "Sonstiges" with this query.

acka47:
for the geocache, we will have to comment out lines 22-23.

dr0i:
@acka47 the SPARQL-query wasn't allowed to have comments in it. With the new PR it does. Also updated the query.

dr0i:
@acka47 resources with new geo-data in production now. Do you have any ID to test this? Geo-cache seems to be all right, at least ~130 entries less.

acka47:
Looks good in general, I updated the numbers in https://github.com/hbz/nwbib/wiki/Wikidata-Matchingverfahren#abdeckung. As soon as you have updated http://stats.lobid.org/scoreCoverageCsv_alleWerte.csv I can create another overview.

[...]

PR #623, opened 2017-11-17
Update SPARQL-query

dr0i:
- allow comments in SPARQL-query

See #622.


acka47:
You can now merge this, @dr0i, so that the new geocache will be built using this query and the new index on Monday will be built on the new geocache.

PR #651, opened 2017-12-19
Fix logging config for Play

dr0i:
Application log-level was not set to DEBUG but that was in fact used.
Using logback.xml fixes this.
Also, this comes with a FileRoller. Should roll every week.

### DIFF
--- a/web/conf/application.conf
+++ b/web/conf/application.conf
@@ -58,15 +58,5 @@ application.langs="de"
 
 # Logger
 # ~~~~~
-# You can also configure logback (http://logback.qos.ch/),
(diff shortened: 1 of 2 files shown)


dr0i:
@fsteeg if you like it you may also configure the other web apps like nwbib with this logging config. Logfiles are too big at the moment in many our projects.

fsteeg:
You committed as sol, otherwise +1

I will check out the logging situation in NWBib and others and use your config, thanks.

dr0i:
As we don't force push on open PRs I will just merge.

Issue #652, opened 2017-12-20
RDF serialization fails for some resources


fsteeg:
See:

https://lobid.org/resources/HT018472857.rdf
https://lobid.org/resources/BT000166905.rdf
https://lobid.org/resources/BT000044841.rdf

Locally:

`[BadURIException: Only well-formed absolute URIrefs can be included in RDF/XML output: <L> Code: 57/REQUIRED_COMPONENT_MISSING in SCHEME: A component that is required by the scheme is missing.]`

`In /home/fsteeg/git/lobid-resources/web/app/controllers/resources/RdfConverter.java:69`

TTL and NT work fine:

https://lobid.org/resources/HT018472857.ttl
https://lobid.org/resources/HT018472857.nt

For others, `.rdf` works:

https://lobid.org/resources/HT007448117.rdf
https://lobid.org/resources/HT007053130.rdf
https://lobid.org/resources/HT007053132.rdf

PR #653, opened 2017-12-20
Fix issue in context, improve RDFConverterTests

fsteeg:
Will resolve https://github.com/hbz/lobid-resources/issues/652

Data uses uppercase `Location` for `type` of `spatial`, see:
https://lobid.org/resources/HT018472857.json

Context declares a lowercase `location` key, see:
http://lobid.org/resources/context.jsonld

Besides the context change, this pull request expands the RDFConverterTests to test multiple resources.

### DIFF
--- a/src/main/resources/labels/context-labels.json
+++ b/src/main/resources/labels/context-labels.json
@@ -354,6 +354,11 @@
       "multilangLabel" : {}
    },
    {
+      "name" : "Location",
+      "uri" : "http://purl.org/dc/terms/Location",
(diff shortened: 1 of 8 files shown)


fsteeg:
Updated this pull request: The actual context is generated from context-labels.json, so this is where this has to be fixed. Also, the correct fix is not to make the existing `location` key uppercase (which is used elsewhere as a property), but to add a new `Location` key to be used as `type`.

dr0i:
+1

PR #667, opened 2018-01-09
Add labels of complex subject

### LINKED ISSUE #639: Add labels to ComplexSubjects
For implementing efficient topic search in NWBib based on API 2.0, we need full subject chains in a single field, which can then be used in an aggregation. @dr0i suggested to add `label` fields to subjects with type `ComplexSubject`. So e.g. for https://lobid.org/resources/HT019305260.json we'd get these labels:

```json
{
    "componentList" : [ ... ],
    "type" : [ "ComplexSubject" ],
    "label" : "Rheinland | Geologie"
},{
    "componentList" : [ ... ],
    "type" : [ "ComplexSubject" ],
    "label" : "Köln-Klettenberg | Gebäude | Garten | Stein"
}
```

Additionally, we need a `raw` `keyword` subfield for the `subject.label` field in `index-config.json`:

```json
"label": {
    "type": "text",
    "fields": {
        "raw": {
            "type": "keyword",
            "index": "true"
        }
    }
}
```

The above is exactly what we already have for `subject.componentList.label` in the `572-updateEsTo5.6.3` branch. This could be copied for `subject.label`.

ChristophEwertowski:
BT000040377 is still incorrect but will be dealt with in a different ticket because it's a fringe case.

Resolves #639


ChristophEwertowski:
* Test case with multiple MAB subject fields: [Diff of BT000003404](https://github.com/hbz/lobid-resources/pull/667/files#diff-72ba3ec0d6b7d25c5d5e9b127002874d).
* Test case with the same subfield multiple times: [Diff of BT000103077](https://github.com/hbz/lobid-resources/pull/667/files#diff-0078ebd8dac0436420596f03e40ae0fd).

The second example is an old NWBib subject which originally was "Schulgeschichte / Herne" but is splitted in Aleph in two subfield a. To cite Mrs. Blankenhagen: 
> Da Unterfeld a nicht wiederholbar ist, wurde für den WorldCat-Export eine eigene Vorgabe geschrieben:
Bei Vorhandensein von mehreren $a soll der Inhalt des zweiten und weiteren $a jeweils nach Komma Blank (,_) an den Inhalt des ersten $a angefügt werden (betr. MARC 65X und 689).

Because "/" could be misinterpretated as a separator I stick to the WorldCat convention.

fsteeg:
Test system is now running again (the proxy was still configured to use quaoar1, not weywot2):

http://test.lobid.org/resources/BT000003404
http://test.lobid.org/resources/BT000103077
http://test.lobid.org/resources/search

I'll look into the naming issue mentioned in https://github.com/hbz/lobid-resources/issues/639#issuecomment-348441148 and will then review this.

fsteeg:
I think there should be a way to make the issue mentioned in https://github.com/hbz/lobid-resources/issues/639#issuecomment-348441148 work. (Re-)using `subject.label` (until now only used by for non-componentList subjects like the NWBib classification) will make the aggregation more involved (we only want the values where the type is `ComplexSubject`, we'll probably have to combine a nested aggregation and a bucket selector aggregation), but it's nice that we'll get a single `subject.label` field which can also be used for queries against all subjects (both with and without `componentList`).

So +1 from me on this. If it turns out this does not work, we can still rename the field to something specific, since we don't take any field away (`subject.label` already exists). Reassigning @dr0i for additional review and subsequent deployment.

dr0i:
Deployed to prodcution, closed.

Issue #670, opened 2018-01-11
HT012990825 isn't indexed


ChristophEwertowski:
[HT012990825 (MAB)](http://lobid.org/hbz01/HT012990825) can't be found at http://lobid.org/resources/HT012990825.

acka47:
Same for HT000052280, see [MAB source](http://lobid.org/hbz01/HT000052280) and http://lobid.org/resources/HT000052280. Reported by @hagbeck

acka47:
@dr0i you are already on this, aren't you? @hagbeck asked about it...

dr0i:
HT012990825 has a URI in in 088 subfield (where a litral is expected to create `bibframe:note`).
HT000052280 has no `dct:created` (MAB 002 is missing).
Both are fixed with the above mentioned PRs. Will be deployed with next fulldump index at monday morning.

PR #676, opened 2018-01-18
Add prefix of "Siehe:" if a URI is detected in bib:note

dr0i:
This enforces bib:note to be a literal rather than an URI (=> an object).

See #670.

- add HT012990825


dr0i:
http://lobid.org/resources/HT012990825 deployed to production. The fix for all the concerned resources will be done when the next full index is build.

ChristophEwertowski:
+1

PR #680, opened 2018-01-19
675 fix mapper parsing exception

### LINKED ISSUE #675: MapperParsingException object mapping
After resolving #674 these MapperParsingException are left:
1 x `alternativeTitle`, 1 x `abstract`, 5 x `otherTitleInformation`.
As they break bulk-requests around 7 x 1.5k = 10.5k resources are still missing. ratio of resource/item is ~ 1/4 => 10.5k /5 = 2.4k MAB-resources are probably lost.

The error-message ("Can't get text on a START_OBJECT" ) means Elasticsearch was expecting an entry of type "string" but  is given an object as input. If so, the solution would be to ensure a literal even when hitting on a URI in the catalog.

dr0i:
Resolves #675.


ChristophEwertowski:
Please also add regular expressions for https, so that we don't have to open this ticket again. Otherwise, +1.

dr0i:
Yes probably a good idea. The "https" will now also be catched.

PR #693, opened 2018-01-26
Set replica to 0 only when creating new index

### LINKED ISSUE #692: Don't reduce replicas when updating an index
In https://github.com/elastic/elasticsearch/issues/302 it's asked:
> In case of replica reduction, do indices get physically removed from disk?

And the answer was:

> Yes

This seems to be selfevident, but this is our routine at the moment:
To improve index performance the replicas are set to `0` when indexing. After the completion of the indexing of all 20M documents the replicas is set to `1`. This is fine when creating a _whole new_ index but as these settings are reused also for _updating_  the reduction of the replica means to first completely _erase_ the replicas (~35 GB). And this to index only a handful of MB. And after the update the replica is rebuild _completely anew_ (35 GB!).

This gives not only bad performance but can very well be the source of our `FailedNodeExceptions` which happen to occur after the updating of the productive index and at the following updating of the staging index. It also explains why it takes 30m for the cluster to regain GREEN status after indexing only a handful of documents.

So, when updating: don't reduce replicas.

Complements #648.

dr0i:
Resolves #692.

### DIFF
--- a/src/main/java/org/lobid/resources/ElasticsearchIndexer.java
+++ b/src/main/java/org/lobid/resources/ElasticsearchIndexer.java
@@ -84,6 +84,7 @@ public class ElasticsearchIndexer
 	private static ObjectMapper mapper = new ObjectMapper();
 	// TODDO setter?
 	public static double MINIMUM_SCORE = 5.0;
+	private HashMap<String, Object> settings = new HashMap<>();
 
 	/**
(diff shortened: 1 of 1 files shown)


fsteeg:
+1

dr0i:
Deployed to production. Closing.

Issue #598, opened 2017-10-23
Word wrapping for subjects


ChristophEwertowski:
There is a special dewey label which is a bit too long for the box: Geografie Mitteleuropas und Reisen in Mitteleuropa    Geografie Deutschlands und Reisen in Deutschland. The problem is that it's present at 6.8 million resources ([query](http://lobid.org/resources/search?subject=Geografie+Mitteleuropas+und+Reisen+in+Mitteleuropa%C2%A0%C2%A0%C2%A0%C2%A0Geografie+Deutschlands+und+Reisen+in+Deutschland)) and part of it is sometimes hidden by the map, see http://lobid.org/resources/HT007007830.
[edit dr0i: ]  you should phrase the search:  http://lobid.org/resources/search?subject=%22Geografie+Mitteleuropas+und+Reisen+in+Mitteleuropa%C2%A0%C2%A0%C2%A0%C2%A0Geografie+Deutschlands+und+Reisen+in+Deutschland%22 . This results in ~70k hits.

acka47:
We decided to correct the label in the lookup file we use for adding DDC labels in the transformation process.

ChristophEwertowski:
Already adjusted on quaoar1 in folder "lobid-resources".

ChristophEwertowski:
Forgot to open a pull request and therefore didn't assign it for review. You can see the change now at [HT017451384 (test, json)](http://test.lobid.org/resources/HT017451384.json).

acka47:
+1

dr0i:
Deployed to production, see e.g. http://lobid.org/resources/HT007007830.
Closed.

PR #711, opened 2018-02-05
598 correct ddc label

ChristophEwertowski:
Resolves https://github.com/hbz/lobid-resources/issues/598


dr0i:
Ah, I see the productive `hduser@weywot1:~/git/lobid-resources/src/main/resources/deweyLabels.tsv` already adapted, but not the one in `~/git/lobid-resources-stage` which is of great importance for it's the source when building the new full index.
Updated that now.

Issue #536, opened 2017-07-27
Move content from 710 to subject array


acka47:
Currently these free subjects (resprctively subject heading from other authority files) are put into `subjectAltLabel`, see e.g. http://lobid.org/resources/TT002234459. In https://github.com/hbz/lobid-resources/issues/8#issuecomment-214744018 I wrote that it wouldn't make sense to put it into the subject array but that was before we had RSWK subjects seperated in that array with `componentList`. As we have this now we can just add the free text subjects from 710 like this:

```json
{
   "id":"http://lobid.org/resources/TT002234459#!",
   "subject":[
      {
         "id":"http://dewey.info/class/914/",
         "label":"Geografie Europas und Reisen in Europa",
         "source":{
            "id":"http://d-nb.info/gnd/4149423-4",
            "label":"Dewey-Dezimalklassifikation"
         }
      },
      {
         "label":"Wandern, Diez, Lahnstein, Rhein-Lahn-Info, Rhein, Lahn, Wirtschaftsförderung"
      },
      {
         "componentList":[
            {
               "id":"http://d-nb.info/gnd/4049787-2",
               "label":"Rhein-Lahn-Kreis",
               "source":{
                  "id":"http://d-nb.info/gnd/7749153-1",
                  "label":"Gemeinsame Normdatei (GND)"
               },
               "type":[
                  "PlaceOrGeographicName"
               ]
            },
            {
               "id":"http://d-nb.info/gnd/4114067-9",
               "label":"Jakobsweg",
               "source":{
                  "id":"http://d-nb.info/gnd/7749153-1",
                  "label":"Gemeinsame Normdatei (GND)"
               },
               "type":[
                  "PlaceOrGeographicName"
               ]
            },
            {
               "label":"Karte",
               "type":[
                  "SubjectHeading"
               ]
            }
         ],
         "type":[
            "ComplexSubject"
         ]
      }
   ]
}
```

@jschnasse, this would be opk for you like this, right?

Relevant part of the current morph is [lines 1796-1800](https://github.com/hbz/lobid-resources/blob/70de741d310fa2e61ae558227259fef7fa400782/src/main/resources/morph-hbz01-to-lobid.xml#L1796-L1800).

ChristophEwertowski:
[...]

PR #713, opened 2018-02-05
Improve rdf-parsing by dropping invalid triples

dr0i:
Generating triples via the morph sometimes results in invalid triples, e.g.
when URIs are invalid, e.g. consist of `<`. To date the parsing execption of
such resource resulted in the dropping of the complete document with all their
triples, even when te other triples were fine. Now, only the invalid triples are
dropped. Log-messages are written so that at a later stage the moprh can be
adjusted to produce valid triples.

See e.g. #536.


dr0i:
This will resolve http://lobid.org/resources/HT019582722.

Issue #718, opened 2018-02-06
HT019582722 has empty objects


dr0i:
Resolving #713 results in HT019582722 being indexed . But the web app has problems to render it: http://lobid.org/resources/HT019582722.

fsteeg:
Root cause of this are empty objects in `isPartOf` and `publication`, see http://lobid.org/resources/HT019582722.json. I've seen the same in `subject` the other day. Fixed the null pointer in https://github.com/hbz/lobid-resources/pull/721 to avoid the error, but details do not display correctly due to the empty object. Should additionally be fixed on the data level.

fsteeg:
Fix to avoid error on page load fixed and deployed, underlying data issue persists and causes wrong details, see http://lobid.org/resources/HT019582722

fsteeg:
Original issue fixed, remaining problem already mentioned in https://github.com/hbz/lobid-resources/issues/569, closing.

PR #721, opened 2018-02-06
Avoid null pointer on empty objects

fsteeg:
Partially resolves https://github.com/hbz/lobid-resources/issues/718


dr0i:
+1
I know these empty objects shouldn't be build in the first place. However. Your commit makes the web app more resilient :)

PR #741, opened 2018-02-22
Changes related to NWBib switch to API 2

fsteeg:
@acka47: assigning PR for functional review since we have no dedicated issue.

Group subjects in details view by their source: http://stage.lobid.org/resources/HT019570509


acka47:
+1 Though, I think we should shorten the labels a bit ("DDC", "GND" & "NWBib-Sachsystematik"), we can leave it like this for now.

Issue #745, opened 2018-02-23
Change order of volume statement


ChristophEwertowski:
A publisso and NWBib customer wants "Bandangabe in Vorlageform" (MAB 089) and not "Bandangabe in Sortierform" (MAB 090) as a volume statement. We couldn't find a ticket which explains why we have done it this way so probably it was just non-librarian thinking ;) 

Issue: https://jira.hbz-nrw.de/browse/FRL-118.

ChristophEwertowski:
Deployed to test. See [HT004944075 (production, json)](http://lobid.org/resources/HT004944075.json) / [HT004944075 (test, json)](http://test.lobid.org/resources/HT004944075.json). If there are multiple subfields `a` they are combined by commata, see [HT017529028 (test, json)](http://test.lobid.org/resources/HT017529028.json). There could be subfields `n` (numbering) and `p` but I couldn't find examples for it. I will ask the costumer for examples. If I get none before the next merge I propose to add it nonetheless.

ChristophEwertowski:
RDA (subfield `n`) also deployed to test. See `numbering` and `title` at [HT019357035 (production, json)](http://lobid.org/resources/HT019357035.json) / [HT019357035 (test, json)](http://test.lobid.org/resources/HT019357035.json).

acka47:
+1

PR #746, opened 2018-02-23
745 change order of volume statement

ChristophEwertowski:
Resolves https://github.com/hbz/lobid-resources/issues/745


dr0i:
Principally +1.
As there are tags introduced in e.g. HT004381366 `"numbering" : "<Tezemu>",` we should check if this doesn't make problems when viewing in our html frontend.
@ChristophEwertowski can you index a small test to check this?

ChristophEwertowski:
It is already put on test.lobid.org, see http://test.lobid.org/resources/HT004381366 or http://test.lobid.org/resources/HT004381366.json.

dr0i:
Ah, ok. But the newly introduced tags seems indeed to make troubles - see http://lobid.org/resources/HT004381366  "Band Tezemu" , whilst there is only "Tezemu" in http://test.lobid.org/resources/HT004381366. So there are at least two ways to fix this:
1. remove the tags from the data
2. fix the html-rendering in the frontend

I would propose to go with 1. because that help ALL possible HTML-frontends. Also, I think the `<>` don't serve any normal-human-understandable purpose, am I right?

ChristophEwertowski:
Well, I only see the difference between "Tezemu" in production and "<Tezemu>" in test but no "Band Tezemu". Like it's in the [MAB-XML](http://lobid.org/hbz01/HT004381366). 
Supposedly "<Tezemu>" is how it's stated on or in the volume. On the frontend the sharp brackets are removed anyway, see the title field in [HT004381366 (test, frontend)](http://test.lobid.org/resources/HT004381366). So if it doesn't lead to indexing errors we should let it in the data with the brackets and in the frontend without it.

dr0i:
Look at http://lobid.org/resources/HT004381366 , see column "Erschienen als". There is the "Tezemu".

acka47:
Although I already gave my "+1", I am all for removing `<>` during the transformation (as I already said in the standup).

ChristophEwertowski:
Brackets removed from RAK (subfield a) and RDA resources (subfield n). `curl 'http://weywot5.hbz-nrw.de:9200/resources-smalltest/resource/HT004381366' |json_pp ` gives the right results in `numbering` and `title` but the frontend hasn't catched up yet.

dr0i:
+1

dr0i:
> frontend hasn't catched up yet.
That's probably the cache. Could restart but that's not necessary since in your updated tests the problematic resource is gone for good.

dr0i:
Deployed to production, se e.g. https://lobid.org/resources/HT003654516.json. Closed.

PR #766, opened 2018-03-08
367 add ISSN of parallel ressource

### LINKED ISSUE #367: Add link to related zdb resources
A customer from the ZBMed wants to have the ISSN from the fields 542 and 545 in the PUBLISSO repository. 542 is the standard ISSN field which is already implemented. 545 is used by the ZDB for paralel edition (e.g. an online version of a printed resource or the other way around). Since this relation is not coved elsewhere we should implement it.  

Jira issue: https://jira.hbz-nrw.de/browse/FRL-156

ChristophEwertowski:
Resolves #367


ChristophEwertowski:
As we will link on other ZDB- and non-ZDB-resources instead of using the ISSNs of these resources, I will close this pull request.

Issue #768, opened 2018-03-09
Presentation problem in sub heading when two otherTitleInformation strings


acka47:
See http://lobid.org/resources/HT018230040

![bug](https://user-images.githubusercontent.com/160292/37200670-be0b62fa-2385-11e8-8569-0620f7a31f25.png)

I propose to only use the first entry in the arrary for the sub heading.

fsteeg:
Deployed to stage: http://stage.lobid.org/resources/HT018230040

acka47:
....and the award for quickest fix of the month goes to @fsteeg: :trophy: 

+1

PR #769, opened 2018-03-09
Use first element of `otherTitleInformation` as subtitle

fsteeg:
Will resolve https://github.com/hbz/lobid-resources/issues/768


dr0i:
+1

PR #771, opened 2018-03-14
367 add links to related resources

ChristophEwertowski:
Resolves https://github.com/hbz/lobid-resources/issues/367


src/test/resources/jsonld/HT002192569
```diff
@@ -0,0 +1,172 @@
+{
```
dr0i: why adding this new test resource - it seems that the already existing resources do already well testing the "related" data.

src/test/resources/output/json/00219/HT002192569.json
```diff
@@ -0,0 +1,172 @@
+{
```
dr0i: why adding this new test resource - it seems that the already existing resources do already well testing the "related" data.

dr0i:
@acka47 and @ChristophEwertowski:
Everytime a new field is introduced we must also update the `index-config` _if_ the field should be searchable. If not, all is done .
If it _should_ be searchable: 
Should  `related.label` be lower in scoring when evaluate the ranking of the documents? https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-boost.html suggests to not put this in the index mapping but  to put the boost parameter into the query.

ChristophEwertowski:
The resource was needed for an earlier commit but not anymore. Thanks for the tip!
The label doesn't have to be searchable, it's only a nicer hint for humans about the content of the related resource. Important is the hbz-URL/`related.id` where the information also can be looked up.

dr0i:
Ok. Just to be cleared: it's not necessary to query `related.id` ? One would be able to query e.g. "give all resources that are related to http://lobid.org/resources/HT017233564#!" , which would be nice IMO because these information is not reciprocally existing in HT017233564#! , i.e. you can't just have a look at HT017233564#! to list all the related resources.

ChristophEwertowski:
Argh! I asked some collegues about linked resources. Links to parallel resources (print/online, different languages) aren't automatically created in the linked resource. Therefore we should have `related.id` indexed. This would lead to a more ZDB-like search, e.g. you search for an ISSN and also get the parallel edition (but instead with `hbzId`). What do you think @acka47 ?

acka47:
+1 for indexing the `related.id`

dr0i:
+1

dr0i:
Deployed to production, see e.g. http://lobid.org/resources/HT010726584.json. Closed.

Issue #775, opened 2018-03-21
Add missing hbz-URLs to predecessor and successor


ChristophEwertowski:
The [FAZ (HT002192569, MAB-XML)](https://lobid.org/hbz01/HT002192569) has a predecessor but the hbzId from MAB 780-1.9 isn't used. If I look at the morph I see the same error for successors (MAB 785).

ChristophEwertowski:
Deployed to test, see ¸predecessor` and `successor` at [HT016138627 (production, json)](http://lobid.org/resources/HT016138627.json) and
[HT016138627 (test, json)](http://test.lobid.org/resources/HT016138627.json).

acka47:
The hasbang is missing at the end of the URI, it shoudl read `http://lobid.org/resources/HT015869763#!`. Otherwise +1.

PR #777, opened 2018-03-21
Add missing hbz-URLs to predecessor and successor

ChristophEwertowski:
Resolves https://github.com/hbz/lobid-resources/issues/775

### DIFF
--- a/src/main/resources/morph-hbz01-to-lobid.xml
+++ b/src/main/resources/morph-hbz01-to-lobid.xml
@@ -1975,19 +1975,21 @@
 		<!-- ####################### -->
 		<!-- ########## predecessor and successor -->
 		<!-- ####################### -->
-	    <data source="531[-xyz]1.a|780-1.t" name="@predecessorLabel">
-	        <replace pattern=".*---&gt; :? ?" with=""/>
-	        <replace pattern="&gt;|&lt;" with=""/>
-	    </data>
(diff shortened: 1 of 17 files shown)


dr0i:
+1
A question is arising when comparing `succesor/predecessor.label` to the one mentioned in #367: at the moment these labels are indexed meaning they are not only shown but also queriable (and queried when doing an `_all` query, weighted the same as e.g. the `title` field _if_ not otherwise declared within in the query which could be answered by @fsteeg ). Question is: is this ok so?

ChristophEwertowski:
If we don't show a property in the frontend we shouldn't have it indexed. But since the `successor/predecessor.label` is an important information it should be indexed, shown and have links which redirect to the resource (if we have a lobid-URL of course). I don't think the `successor/predecessor.id` has to be indexed. What do you think, @acka47 ?

dr0i:
> If we don't show a property in the frontend we shouldn't have it indexed. 

Not true in general , note to clarify possible misunderstandings: 
You _can retrieve_ data _without having it indexed_. By _indexed_ it is meant that you can also _search_ for it.

ChristophEwertowski:
We will let them indexed. For looking at the index see https://github.com/hbz/lobid-resources/issues/782.

dr0i:
Deployed to production, see e.g. http://lobid.org/resources/HT000009600.json. Closed.

Issue #258, opened 2017-02-20
Badly cataloged MAB data


ChristophEwertowski:
URL | MAB field | Error in jsonld 
---- | --- | ---
~~http://lobid.org/resources/BT000103077~~ | ~~599~~ | ~~http://lobid.org/resources/von is created as an Id for containedIn. Should be http://lobid.org/resources/BT000102973#! -> BT000102973~~
~~http://lobid.org/resources/search?q=publication.startDate%3A%5B2029+TO+*%5D~~ | ~~089, 090, mostly 425, 595~~ | ~~28 hits. Mostly impossible years as publication.startDate. Latest [resource](http://lobid.org/resources/HT015186671) has a publication date of 2029. Multiple of these resources have volume data in field 425-.~~ Corrected where plausible data was available.
~~http://lobid.org/resources/search?issued=0001-1400&size=100&sort=oldest~~ | '' | ~~388 hits. Mostly wrong year statements.~~
http://lobid.org/resources/HT019407711, HT019401543, HT019469434 | " | Year 2013, 1925, 1937. Can't edit.
~~http://lobid.org/resources/search?issued=0000~~ | '' | ~~19363 hits.~~ No results anymore.
~~http://lobid.org/resources/search?issued=9999~~ | '' | ~~25 hits.~~ No results anymore.
http://lobid.org/resources/search?q=title%3Ajournal | - | Lots of titles with only the title "Journal".
http://lobid.org/hbz01/HT016247254 | 425c | 2912 -> according to ZDB year 2012. Can't edit.
http://lobid.org/hbz01/HT009063035, http://lobid.org/hbz01/HT017524266, http://lobid.org/hbz01/HT018535049 | 101c | Should be earlier name but contains fields of activity (isn't referred by lobid-resources). Found multiple more. Search for more resources like this in Aleph.
~~http://lobid.org/hbz01/HT016638916~~ | ~~100a~~ | ~~Role mentioned after name. Should probably be in subfield 4.~~
~~http://lobid.org/hbz01/TT001099235~~ | ~~101p~~ | ~~Role mentioned after name. Should probably be in subfield 4. Aleph ungleich HBZ01~~
~~http://lobid.org/hbz01/HT018242370, http://lobid.org/hbz01/HT017794760~~ | ~~50.a (Druckschrift)~~ | ~~AudioVisualMedia which is also a book. Should be done after https://github.com/hbz/lobid-resources/issues/349~~
[...]

PR #789, opened 2018-03-29
Create new index "deleted"

dr0i:
This provides a new index to lookup deleted hbz01-resources. It works like this:
So called "Loeschsaetze" in aleph Mab xml are stored in an own index called
"deleted". As each Loeschsatz only contains an internal aleph system number
a lookup to associate this internal id with a lobid-resources id is needed.
Since lobid/lodmill#258 the API at http://lobid.org/hbz01 provides a lookup
service for aleph internal sysnumbers. This API is used to enrich the minimal
data residing in the Loeschsatz. As even no date is provided to specify when
the resource was deleted this date is set in the morph. The date used is just
the date at when the transformation is done, so this is not very accurate but
the best available. Note: it may sometimes be necessary to reindex a Loeschsatz
at a later point, say some days. The date would then become inaccurate, but
again - it's all we have.

- parametrize the classes to control which and how an index should be build
- improve logging by catching and printing more exceptions
- enrich loeschsatz with data gained from hbz01-API
- provide index-config and a deletion-labels directory for new jsonld-context
- provide junit test for a Loeschsatz

See #258.


fsteeg:
+1

PR #798, opened 2018-04-12
Use a proxy for tiles

dr0i:
Using a lobid proxy anonymizes outgoing traffics of our users.
The wikimedia map server is used via the reverse proxy.

- remove some comments

See hbz/lobid#363.


fsteeg:
+1

PR #804, opened 2018-04-17
Build new index entityfacts

dr0i:
Splits the line separated json of the entityfacts dump, gets the GND-ID as the
internal ID of ES and index the records as bulk into ES on our cluster.

Resolves hbz/lobid-gnd#44.

### DIFF
--- a/src/main/java/org/lobid/resources/ElasticsearchIndexer.java
+++ b/src/main/java/org/lobid/resources/ElasticsearchIndexer.java
@@ -9,6 +9,8 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


src/main/java/org/lobid/resources/ElasticsearchIndexer.java
```diff
@@ -91,6 +93,14 @@
...
 	/** Defines if a wikidata lookup should be done */
 	public boolean lookupWikidata;
+	/**
+	 * The date now. Handy to append to index-name to build multiple index' in
+	 * parallel. Switch then by setting the alias.
+	 */
+	public static final String DATE =
+			LocalDateTime.now().toLocalDate().format(DateTimeFormatter.BASIC_ISO_DATE)
+					+ "-" + LocalDateTime.now().getHour()
+					+ LocalDateTime.now().getMinute() + LocalDateTime.now().getSecond();
```
fsteeg: Perhaps it would be safer to extract `LocalDateTime.now()` to a separate constant, otherwise we might get weird errors if the minute switches before we get the second or something like that.

Otherwise, or if that's not an issue, +1.

fsteeg:
@dr0i I can't reassign you, pinging like this.

dr0i:
@fsteeg right, this rare case might occur. Fixed that by using just one point in time.

ChristophEwertowski:
@dr0i : Is this the reason for the wrapInMapperParsingException from 8 a.m. today?

fsteeg:
Will test new index and access via internal ID instead of query from lobid-gnd.

fsteeg:
Using the new index via GET by ID instead of query: https://github.com/hbz/lobid-gnd/pull/82

I have deleted the old index, and set up alias `entityfacts` for the new index.

+1

dr0i:
> @dr0i : Is this the reason for the wrapInMapperParsingException from 8 a.m. today?

@ChristophEwertowski yes. It was caused by some test-indices like [entityfacts-20180417-113650] till I got it right.

Issue #809, opened 2018-04-20
Add provenance information


acka47:
Similar to https://github.com/hbz/lobid-gnd/issues/89. An easy way would be to add add a ~`derivedBy`~ `isBasedOn` triple to the `describedBy` object:

```json
{
  "id":"http://lobid.org/resources/HT019025795#!",
  "describedBy":{
    "dateCreated":"20160706",
    "dateModified":"20170920",
    "id":"http://lobid.org/resources/HT019025795",
    "label":"lobid Ressource",
    "isBasedOn":"http://lobid.org/hbz01/HT019025795"
  }
}
```

A more complex way would also cover information on the software used etc. something like this:

```json
{
  "@context":{
    "id":"@id",
    "describedBy":{
      "@type":"@id",
      "@id":"http://www.w3.org/2007/05/powder-s#describedby"
    },
    "dateCreated":"http://schema.org/dateCreated",
    "dateModified":"http://schema.org/dateModified",
    "wasGeneratedBy":{
      "@id":"http://www.w3.org/ns/prov#wasGeneratedBy",
      "@type":"@id"
    },
    "Activity":"http://www.w3.org/ns/prov#Activity",
    "startedAtTime":{
      "@id":"http://www.w3.org/ns/prov#startedAtTime",
      "@container":"xsd:dateTime"
    },
    "endedAtTime":{
      "@id":"http://www.w3.org/ns/prov#startedAtTime",
      "@container":"xsd:dateTime"
    },
    "used":{
      "@id":"http://www.w3.org/ns/prov#used",
      "@type":"@id"
    }
  },
  "id":"http://lobid.org/resources/HT019025795#!",
  "describedBy":{
    "dateCreated":"20160706",
    "dateModified":"20170920",
    "id":"http://lobid.org/resources/HT019025795",
    "label":"lobid Ressource",
    "wasGeneratedBy":{
      "type":"Activity",
      "startedAtTime":"2018-04-01T01:30:00Z",
      "endedAtTime":"2016-04-01T05:00:00Z",
      "used":[
        {
          "id":"http://lobid.org/hbz01/HT019025795"
        },
        {
          "id":"https://github.com/hbz/lobid-resources"
        }
      ]
    }
  }
}
```

(If we did it all correctly, we'd have to move `dateCreated` and `dateModified` to `<http://lobid.org/hbz01/HT019025795>` resource but that would mean an API break.)

dr0i:
[...]

PR #813, opened 2018-04-26
809 add provenance information

ChristophEwertowski:
Resolves https://github.com/hbz/lobid-resources/issues/809


dr0i:
Checks failed and conflicts. Reassigning @ChristophEwertowski.

ChristophEwertowski:
@dr0i I resolved the conflicts. The checks probably fail because the `endTime` changes every second and the output file can't "keep up". Is this assumption right and can the tests be adjusted?

dr0i:
I have to adjust this in the programs. Will do this later.

dr0i:
You may want to have a look at the java files @fsteeg.

ChristophEwertowski:
@dr0i : I tried it out. The `endTime` of the `createAction` doesn't change, not even after a `mvn clean install` (I looked in the json file). Of course it should change, otherwise the `endTime` statement after the transformation doesn't make sense.

ChristophEwertowski:
+1

src/main/java/de/hbz/lobid/helper/EtikettMaker.java
```diff
@@ -43,7 +43,7 @@
 
 	private static final String TYPE = "type";
 	private static final String ID = "id";
-	private String contextLocation = "web/conf/context.jsonld";
+	private static String contextLocation = "web/conf/context.jsonld";
```
fsteeg: I think this is dangerous, because we now have a static getter and a non-static instance setter for this static field. So we could call the setter on an instance, which we would expect to change only the instance, but would actually change the context for all instances, since it's actually static. Maybe it would be fine to just hard-code the context location for the tests? Or to set up a separate constant, that is used by the instance variable?

src/test/java/org/lobid/resources/Hbz01MabXml2ElasticsearchLobidTest.java
```diff
@@ -272,16 +273,23 @@ static private SearchResponse getElasticsearchDocuments() {
 		static String getAsNtriples() {
 			return Arrays.asList(getElasticsearchDocuments().getHits().getHits())
 					.parallelStream()
-					.flatMap(hit -> Stream.of(toRdf(hit.getSourceAsString())))
+					.flatMap(
+							hit -> Stream.of(toRdf(cleanseEndtime(hit.getSourceAsString()))))
```
fsteeg: Oh, `Stream.of` is cool, didn't know that. But in this case here, it seems we could use `map` and avoid the stream altogether, like this: `.map(hit -> toRdf(cleanseEndtime(hit.getSourceAsString())))`.

src/test/java/org/lobid/resources/Hbz01MabXml2ElasticsearchLobidTest.java
```diff
@@ -272,16 +273,23 @@ static private SearchResponse getElasticsearchDocuments() {
...
 		private static void getAsJson() {
 			for (SearchHit hit : getElasticsearchDocuments().getHits().getHits()) {
-				stripContextAndSaveAsFile(hit.getSourceAsString());
+				stripContextAndSaveAsFile(cleanseEndtime(hit.getSourceAsString()));
 			}
 		}
 
+		private static String cleanseEndtime(String jsonld) {
+			return jsonld.replaceFirst(
+					"\"endTime\":\"\\d\\d\\d\\d-\\d\\d-\\d\\dT\\d\\d:\\d\\d:\\d\\d\"",
```
fsteeg: We could shorten this a bit with quantifiers: `"\"endTime\":\"\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\"",`

src/test/java/org/lobid/resources/Hbz01MabXmlEtlNtriples2Filesystem.java
```diff
@@ -86,8 +85,11 @@ public void etlOutputAsNtriples2Filesystem() {
 				.setReceiver(new AlephMabXmlHandler())
 				.setReceiver(
 						new Metamorph("src/main/resources/morph-hbz01-to-lobid.xml"))
-				.setReceiver(new PipeEncodeTriples()).setReceiver(triple2model)
-				.setReceiver(rdfModelFileWriter);
+				.setReceiver(new PipeEncodeTriples())
+				.setReceiver(new SimpleStringSubstituter(
+						"endTime> \"\\d\\d\\d\\d-\\d\\d-\\d\\dT\\d\\d:\\d\\d:\\d\\d\"",
```
fsteeg: Same here maybe: `"endTime> \"\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\""`

src/main/java/de/hbz/lobid/helper/EtikettMaker.java
dr0i: Agreed. Will (re)use instances instead of introduce statics.

src/test/java/org/lobid/resources/Hbz01MabXml2ElasticsearchLobidTest.java
dr0i: Admittedly not aware of the difference. Using the simpler one as you propose.

src/test/java/org/lobid/resources/Hbz01MabXmlEtlNtriples2Filesystem.java
dr0i: Ok.

dr0i:
Deployed to production, see e.g. http://lobid.org/resources/HT018833649.json. Closed.

PR #834, opened 2018-06-04
814 dewey label fallback

### LINKED ISSUE #814: Dewey-label fallback
If the specific notation is missing in our dewey-concordance (like ` 491.4682421` in our test resource `HT019578856`) one could use a fallback to choose the `491` as label - in this case `Ostindoeuropäische und keltische Sprachen`.

ChristophEwertowski:
Resolves #814 

It seems that MAB 700 contains only main classes ("Hauptgruppen"). Therefore I didn't integrate the Dewey fallback solution for this field. Also now all Dewey notations from 705 are covered.


ChristophEwertowski:
@dr0i : If I create a test index the labels for the Dewey notations don't get transferred. For example HT015455455 ([pull request link](https://github.com/hbz/lobid-resources/pull/834/files#diff-83dcb0c1fda036845681373f3db6573a)) which leads to wrong results if I curl it. What am I missing here? The JSON file, morph and deweyLabels.tsv look O.K.

dr0i:
If we are talking of `sol@weywot2:[~/git/lobid-resources-test] master`:
The `labels.json` is broken. Check with e.g `cat labels.json|json_pp`.
If you do a `mvn clean install -DgenerateTests=true` this fails.

dr0i:
Deployed to production, see e.g. lobid.org/resources/HT015455455.json. Closed.

PR #839, opened 2018-06-07
Use MAB 003 for dateCreated if no dateCreated exists

### LINKED ISSUE #817: Missing dateCreated
We have 752 resources which have a `dateModified` but no `dateCreated` ([query](http://lobid.org/resources/search?q=_exists_%3AdescribedBy.dateModified+NOT+_exists_%3AdescribedBy.dateCreated)).

Case 1: MAB field 002 isn't filled, see http://lobid.org/hbz01/HT000052280. Old resource of the previous system which maybe didn't have this field or the transformation was forgotten. In this case we should use `dateModified` for `dateCreated`.
Case 2: MAB field 002 is filled but not with the right content, see http://lobid.org/hbz01/HT000925978 (filled with just "2002").

Argh!!! `dateCreated` (MAB 002) can be edited in any format (see `11.11.2009` in [HT000210058](http://lobid.org/hbz01/HT000210058) or [HT017320303](http://lobid.org/hbz01/HT017320303) where it doesn't even have a subfield) and saved. There is no restriction or even statement in the *Feldhilfe* how to enter data in this fields. I will talk to the collegues of the Verbundgruppe about it. The id can be edited but not saved.

ChristophEwertowski:
Resolves #817

### DIFF
--- a/src/main/resources/morph-hbz01-to-lobid.xml
+++ b/src/main/resources/morph-hbz01-to-lobid.xml
@@ -2474,6 +2474,26 @@
 				<regexp match="([0-9]{8})" format="${1}"/>
 			</data>
 		</combine>
+		<combine name="~rdf:subject" value="${a}">
+			<if>
+				<none>
+					<data source="002[ab]1.a"/>
+				</none>
+			</if>
+			<data source="@id" name="a">
+				<regexp match="(.*)" format="$[ns-lobid-resource]${1}"/>
+			</data>
+		</combine>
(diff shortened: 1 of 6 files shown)


dr0i:
I tried different simpler approches in the morph ("choose" etc.) but nothing works as expected.
So +1 for your solution!

dr0i:
Deployed to production, see e.g. http://lobid.org/resources/HT000052280.json. Closed.

PR #842, opened 2018-06-11
841 library notation not dewey

### LINKED ISSUE #841: Notation wrongly taken as Dewey notation
From a [comment of #814](https://github.com/hbz/lobid-resources/issues/814#issuecomment-396138200). 
In the resource [HT008815963](http://lobid.org/resources/HT008815963) MAB 700 is filled like this:
```
<datafield tag="700" ind1="-" ind2="1">
   <subfield code="a">406120</subfield>
</datafield>
```
`ind1="-"` is defined as: "Systematik der katalogisierenden Institution". As the ZDB uses this indicator for DDC notations we use it for Dewey notations. Because the ZDB uses a subfield for defining it's usage it can be used to exclude non-Dewey notations. Of course the right notation should also be shown.

ChristophEwertowski:
Resolves #841


dr0i:
+1

dr0i:
Deployed to production, see e.g. http://lobid.org/resources/HT015455455. Closed.

Issue #627, opened 2017-11-21
Fix matchings based on notation 


acka47:
From https://github.com/hbz/lobid-resources/issues/622#issuecomment-345963369:

>[W]e have a systematic mismatch problem with Kreis/Landkreis. Generally, the town is taken as match that has the same name as the Kreis (e.g. Euskrichen for "Kreis Euskirchen"). A good example is http://lobid.org/resources/HT013863573.json. This makes sense as both "label" and "locatedIn" match the query string (for example "Kreis Euskirchen"):

```json
{
  "spatial":{
    "type":"Location",
    "id":"http://www.wikidata.org/entity/Q4074",
    "label":"Euskirchen",
    "geo":{
      "lat":50.659722222222,
      "lon":6.7916666666667
    }
  },
  "locatedIn":{
    "language":"de",
    "value":"Kreis Euskirchen"
  }
}
```

We have to find a way to deal with this.

acka47:
I guess this also holds for Regierungsbezirke, as soon as we will have them back in the geo cache with #631.

dr0i:
#631 is closed, so this issue can be worked on again.

dr0i:
With #643 some of the `Kreis`-issues are resolved, see e.g. [Kreis Euskirchen](http://lobid.org/resources/HT013863573.json). Some remain, e.g. [Kreis Höxter](http://lobid.org/resources/BT000117093.json).

acka47:
I think we could improve the matching results significantly if we matched the strings from each notation separately. Here are the different notations with strings:

- 99: Städte, Kreisfreie Städte, Gemeinden, Ortsteile plus teilweise Stadtbezirke und (leider auch andere wie z.B. Klöster etc.)
- 97: (Land)Kreise
- 96: Regierungsbezirke
- 74: Kleinere weltliche Territorien in Westfalen
- 72: Kleinere geistliche Territorien in Westfalen
- 54: Kleinere weltliche Territorien im Rheinland
- 52: Kleinere geistliche Territorien im Rheinland
- 37: Katholische Kirche. Dekanate
- 36: Katholische Kirche. Diözesen
- 35: Evangelische Kirche. Kirchenkreise
- 28: Eifel
- 24: Niederrhein-Gebiet
- 14: Sauerland
- 12: Weserbergland
- 10: Westfälische Bucht

As this issue makes clear, we already have problems to separate districts and localities in them so it would make sense to spearate their matchings. For this we need two things:

[...]

PR #854, opened 2018-06-25
Also pull Prussian districts from Wikidata for matching

acka47:
See https://github.com/hbz/lobid-resources/issues/627#issuecomment-395698971

@dr0i I also adjusted the code so that entries with notations 97 are also matched with the prussian districts. I hope I did it correctly.


dr0i:
Looks good. 
Going to deploy to stage and let's see if it works.

dr0i:
New index is built. It has around 40 more entries than the old index.

src/main/java/org/lobid/resources/run/WikidataGeodata2Es.java
```diff
@@ -70,7 +70,7 @@
 
 				{
 					put("99", "Q22865 Q262166 Q253019 Q1852178 Q15632166");
-					put("97", "Q106658");
+					put("97", "Q106658 Q5283531");
 					put("96", "Q829277");
 					put("36", "Q3146899 Q2072238");
 					put("97", "Q106658");
```
dr0i: Oh, there are _two_ entries of `97`, which is not good. As long as the entries are same this doesn't have an effect. Now, as they are different, it does behave differently - it overwrites the first entry with the second entry, ignoring "Q5283531".

dr0i:
I was wondering e.g. https://www.wikidata.org/wiki/Q1787322 wasn't enriched. There is a bug in defining `97` two times, effectively ignoring the Prussian districts (`Q5283531`). Will add a fix.

dr0i:
Deplyoed to production, see e.g. http://lobid.org/resources/search?q=spatial.id%3A"http%3A%2F%2Fwww.wikidata.org%2Fentity%2FQ1787322". Closed.

Issue #869, opened 2018-07-06
Configure fields of index "resources" for kibana 


dr0i:
Result of hbz/lobid-gnd#66.
To create nice stats the more index fields be typed as `number` (or even more accurate: `date`) the better.
See [ES-docs for possible types](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/mapping-types.html). There is no need to change the format of the values e.g. of the field `describedBy.dateCreated` -  it seems to be sufficient to just configure the index to:

> "type": "date", 
> "format": "yyyyMMdd"

(https://www.elastic.co/guide/en/elasticsearch/reference/5.6/date.html)

I would say that these fields are reasonable `date` candidates:
- [x] `describedBy.dateCreated`
- [x] `describedBy.dateModified`
- [x] `publication.startDate`
- [x] `publication.endDate`

dr0i:
All done. Kept #873 open for review. Closing this one.

PR #871, opened 2018-07-09
Switch type to "date" for describedBy.date*

dr0i:
See #869.


dr0i:
This led to some `MapperParsingException`, probably a result of resources lacking both dateCreated and dateModified (which is defined as a fallback). So a fallback for the fallback is needed, I would set it to 00010101 which make it implicitly clear that this is cannot be true while still syntactically valid.

Nonetheless one can make some charts. @acka47 I set up some `Resource` charts. Go to `Visualize` and select the names starting with "Resource".

I will work on this to make the MPE go away. Also, there are other interesting fields one would like to make charts out of, e.g. something like https://wiki1.hbz-nrw.de/display/SEM/2012/10/10/ .

dr0i:
This is the cause:

> Caused by: org.joda.time.IllegalFieldValueException: Cannot parse "20062808": Value 28 for monthOfYear must be in the range [1,12]

Yeah, so `date` is sometimes catalogued as yyyyddMM and mostly as yyyyMMdd.
Sad again to see no helping regex in cataloging software.

ChristophEwertowski:
Where's the difference between your first and second dummy resource? Also please align the number of tabs in the code.

dr0i:
Re: difference in test resources:
Have a look in the commit comments:
of 3c46806e4550ce44d2b543ef9ada10f6256d1ecd : 
> - add test resource with bad pattern `yyyyddMM`"

of 0858d3e6a8e5fba4200b8ad424749983c9bbad23: 
> - add two new test resources, ohne without "002" and one without "002" and "003"

Re: formatting:
You are right, have to look in the eclipse settings again. Updated the morph formatting.

dr0i:
Assigning @fsteeg for code review.

dr0i:
@ChristophEwertowski please have a look if the MapperParsingExceptions of the "dateCreated" field are gone, then deploy to production as discussed at friday.

ChristophEwertowski:
Sadly there are a lot of MapperParsingExceptions concerning `describedBy.dateCreated`. There are > 50 exceptions but only 29 results are shown if I search for [resources without `describedBy.dateCreated`](http://lobid.org/resources/search?q=NOT+_exists_%3AdescribedBy.dateCreated&from=15). Of these 29 four are corrected in Aleph which I overlooked last time.

dr0i:
@ChristophEwertowski the MPE's results from e.g. http://lobid.org/hbz01/TT000000489 where a "20010221095428" is notated as creation date. Sickly the SimpleDateFormat converts this into "597580528" , which is malformed , thus MPE. I say "sickly" because it was given an input/output pattern, but the default of the java class is "lenient" meaning "make the best out of it" , but woe. In effect it's unpredictable, so I set "lenient" to `false`. See https://stackoverflow.com/questions/16014488/simpledateformat-parse-ignores-the-number-of-characters-in-pattern/19503019#19503019 for background.

dr0i:
Forgot to say: setting "lenient" to `false` means much more fallback ("00010101") as also the good repairing mechanism is thereby deactivated.

dr0i:
Ah - no more MPEs. 1195 resources don't have a valid entry in 002a (aka "date of creation"). @ChristophEwertowski do you think our Verbundgruppe would like to have a list of these resources?

dr0i:
Deployed to production.
Get all resources with the "fallback" value: https://lobid.org/resources/search?q=describedBy.dateCreated%3A00010101.
Closed.

ChristophEwertowski:
@dr0i : I will send them a list of `hbzId`s after I corrected the HT numbers where possible. The rest will consist of resources which require further research or are TT resources (resources from imports of mostly external people).

ChristophEwertowski:
Friday I corrected most of the HT numbers. Today I sent a mail to the Verbundgruppe with the wish for correction and an explanation that the `dateCreated` is especially important for users who get resources over the API.

PR #875, opened 2018-07-12
Add missing agent to resources with MAB 342 etc.

### LINKED ISSUE #874: Add agent to new contribution object
I forgot to index `contribution.agent.altLabel` in https://github.com/hbz/lobid-resources/issues/859.

ChristophEwertowski:
Resolves #874


dr0i:
+1

dr0i:
This is not reflected in http://lobid.org/resources/HT007015768.json because the MAB XML source of HT007015768 has no more a second agent.
Nonetheless, deployed to production assuming it is good as it is.
But, it would be nice if @ChristophEwertowski could approve it's working with another resource.

PR #884, opened 2018-07-27
Switch property for Braille text

### LINKED ISSUE #827: BrailleBook URL doesn't resolve
`BrailleBook` is defined by http://purl.org/library/BrailleBook which doesn't resolve. I searched for "Braille" in Schema.org, bibframe, RDA and BibO which didn't lead to a result.

ChristophEwertowski:
Resolves #827

### DIFF
--- a/src/main/resources/labels/labels.json
+++ b/src/main/resources/labels/labels.json
@@ -14560,7 +14560,7 @@
       "label" : "Audio-Visuell"
    },
    {
-      "uri" : "http://d-nb.info/gnd/4145976-3",
+      "uri" : "http://purl.org/lobid/lv#Braille",
       "label" : "Blindenschrift"
    },
    {
--- a/src/main/resources/labels/labels.json
+++ b/src/main/resources/labels/labels.json
@@ -14402,9 +14402,10 @@
       "uri" : "http://purl.org/lobid/lv#Festschrift"
    },
    {
+      "label" : "Spiel",
+      "multilangLabel" : {},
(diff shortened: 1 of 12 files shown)


ChristophEwertowski:
@fsteeg : The icon for "Blindendruck" doesn't appear in test.lobid.org/resources/search. The resources.conf is adjusted and the test resource [TT001210514](http://test.lobid.org/resources/TT001210514) has the right icon and label. What did I overlook?

fsteeg:
I've opened https://github.com/hbz/lobid-resources/issues/885 for the underlying label lookup issue.

dr0i:
Assigning @acka47 , please have a look at the Files changed. As there are more added definitions than just the Braille one , and if you agree, we could merge this already to have the data in our index. Icons and other frontend issues can be solved later.

TobiasNx:
@acka47 This PR is quite old and I think we should close it, since it only changes the old morph transformation.

The issue still seems to be valid, the URI for BrailleBook ( http://purl.org/library/BrailleBook) seems to be not resolvable.
@ChristophEwertowski changed the property first to a GND property and then to a lobid vocabs concept "http://purl.org/lobid/lv#Braille". I dont know if we still want to change this but then we should do this in a new PR.


Additionally the changes in https://github.com/hbz/lobid-resources/pull/884/commits/cfe256e9f11e3ae2fee7af0ad2a81b49be7ee494 seem to be valid and could be cherry picked, since they add missing labels to the context labels file.

PR #886, opened 2018-07-27
Ensure killing of web app

dr0i:
This may prevent some monit restart problems.
See mail "Lobid-resources web app unerreichbar für 5 Stunden"
from Fri, 27 Jul 2018 11:46:07 +0200 to lobid-admin.

Message ID: <0ef83206-bffb-418d-a847-aa728f9a1c21@hbz-nrw.de>

### DIFF
--- a/web/monit_restart.sh
+++ b/web/monit_restart.sh
@@ -31,10 +31,14 @@ cd $HOME/git/$REPO
 case $ACTION in
 	start)
 		kill $(cat target/universal/stage/RUNNING_PID)
+		sleep 14
+		kill -9 $(cat target/universal/stage/RUNNING_PID)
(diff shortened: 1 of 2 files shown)


fsteeg:
This will always run both `kill` and `kill -9`, right? Couldn't we then just run `kill -9` in the first place?

dr0i:
Right, we could, but `kill` is a _graceful_ kill meaning that the web-app may do some things when going down (e.g., logging that it's going down), whilest `kill -9` is a brutal killing without giving the to be killed process time to pray.
So we try first to take down the process gracefully and wait a reasonable time before killing the process brutally if still standing.

fsteeg:
Right, makes sense. +1

dr0i:
Deployed. Closing.

PR #893, opened 2018-08-01
Support RSS as response format for search requests

### LINKED ISSUE #887: Support RSS as response format for search requests
Inspired by https://www.hughrundle.net/2018/03/25/watching-the-feeds/

> RSS has been around for so long, there's a chance your library catalogue includes it as a feature. This may be more useful than you imagine, because RSS in library catalogues commonly hooks in to searches: effectively it allows you to subscribe to a particular set of search parameters.

fsteeg:
Will resolve #887


dr0i:
+1

PR #896, opened 2018-08-02
Only cover dates up to days, add example

### LINKED ISSUE #894: Set dates with in day time statements to yyyyMMdd
Yesterday I sent the Verbundgruppe a list of `dateCreated` which were transformed to "00010101" because we don't cover the contents of the field. Today Mrs. Bl. told me that we should also cover the time information up to seconds because they are valid.

ChristophEwertowski:
TT000000493: Example with hours, minutes and seconds.

Resolves #894


dr0i:
Deployed to production, see e.g. http://lobid.org/resources/TT000000493.json.

Issue #897, opened 2018-08-02
Labels aren't transferred to web/conf/context.jsonld


ChristophEwertowski:
See title. Pascal already tested it by removing a class which didn't have any effect on the web/conf/context.jsonld file.

dr0i:
See #899.

ChristophEwertowski:
Will be dealt with in https://github.com/hbz/lobid-resources/pull/884.

dr0i:
From https://github.com/hbz/lobid-resources/commit/2062ff66c549e0993c3dbfe6b8a4e9c9bde2ebbe : 
> Using "name" instead of "label" makes the labels exported to context.json.

Closing.

PR #899, opened 2018-08-02
Add "Braille"

dr0i:
Using "name" instead of "label" makes the labels exported to context.json.

This seems strange because:

1. we have many definitions of Uris using "label" instead of "name"
2. I think somewhere in the code is a fallback for "name" when "name" is missing so "label" is used.

However I now I grasped once the idea of the difference between "name" and "label", but it's lost to me again. Maybe this "label-name" thing should be made simpler by dropping one definition, it seems superflous @jschnasse ?

See #897.

### DIFF
--- a/src/main/resources/labels/labels.json
+++ b/src/main/resources/labels/labels.json
@@ -14666,7 +14666,7 @@
    },
    {
       "uri" : "http://purl.org/library/BrailleBook",
-      "label" : "Blindenschrift"
+      "name" : "Blindenschrift"
    },
    {
       "uri":"http://rdaregistry.info/termList/RDAMediaType/1003",
--- a/web/conf/context.jsonld
+++ b/web/conf/context.jsonld
@@ -363,6 +363,9 @@
     "Thesis" : {
       "@id" : "http://purl.org/ontology/bibo/Thesis"
     },
+    "Blindenschrift" : {
+      "@id" : "http://purl.org/library/BrailleBook"
+    },
     "ArchivedWebPage" : {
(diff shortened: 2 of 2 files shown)


jschnasse:
Hi there,


|    uri*    |    label    |    name*    | icon |    reference_type (default is String)    |    container    |    comment    |    weight    |
|-----------|-----------|-----------|-----------|------------|-----------|-----------|-----------|

'*' are mandatory, all others are optional

**uri***: The rdf predicate 
**label**: A label you might want to use for display 
**name***: the name of the json element 
**icon**: use this to add an icon font 
**reference_type**: the json-ld type (default is String)
**container**: the json-ld container 
**comment**: a comment you might want to add 
**weight:** a weight you can use to display elements in the correct order 

Example:

```
{
  "uri" : "http://purl.org/dc/terms/alternative",
  "comment" : "",
  "label" : "Titelzusatz",
  "icon" : "",
  "name" : "alternative",
  "referenceType" : "String",
  "container" : "@set",
  "weight" : "3",
  "type" : "CONTEXT",
  "multilangLabel" : { }
}
```
converts to context entry ...
```
alternative": {
    "@id": "http://purl.org/dc/terms/alternative",
    "@container": "@set"
}
```
or (annotated context)
```
alternative": {
    "icon": "",
    "weight": "3",
    "comment": "",
    "@id": "http://purl.org/dc/terms/alternative",
    "label": "Titelzusatz",
    "@container": "@set"
}
```

ChristophEwertowski:
Will be dealt with in https://github.com/hbz/lobid-resources/pull/884.

PR #900, opened 2018-08-03
688 info about superordinate resource from note field

### LINKED ISSUE #688: Use note field for information about superordinates
In MAB 501 there are whole lot of resources which contain information about titles they are contained in. The resources I looked at were from 1999. 

[HT010107570  (MAB)](http://lobid.org/hbz01/HT010107570)
```
<datafield tag="501" ind1="-" ind2="1">
   <subfield code="a">Aus: Stoma. Zs. f. Zahn-, Mund- u. Kieferheilkunde. 19. 1966</subfield>
</datafield>
```
Even if we have no id we should include this information, shouldn't we?

ChristophEwertowski:
Resolves #688


dr0i:
Deployed to production, see e.g. http://lobid.org/resources/TT001451033.json. Closed.

PR #915, opened 2018-09-10
Update dataset description

### LINKED ISSUE #913: Update dataset desrciption for consumption by google
See https://www.blog.google/products/search/making-it-easier-discover-datasets/

The test tool finds some errors, see https://search.google.com/structured-data/testing-tool#url=http%3A%2F%2Flobid.org%2Fresources%2Fdataset. Mainly:

- [x] use `schema:name` instead of `rdfs:label`. (I guess we can do this on dataset level although we use rdfs:label as generic label property in the actual data.)
- [x] use `schema:distribution` instead of `dcat:distribution`

The test tool also doesn't like `dcat:Distribution`. schema.org only has [`DataDownload`](https://schema.org/DataDownload) which does not really fit for the API which may be used to get a complete dump but rather offers access to the dataset via subsets.

acka47:
Resolves #913. We will probably have to fix some things for the HTML view before deploying, thus assigning @fsteeg .

### DIFF
--- a/web/app/views/dataset.scala.html
+++ b/web/app/views/dataset.scala.html
@@ -12,7 +12,7 @@
 		<tr>
 			<td>Titel</td>
 			<td><a href='@json.get("id").asText()'>
-				@json.get("name").asText() &mdash; @json.get("alternateName").asText()
+				@json.get("name").get("de").asText() &mdash; @json.get("alternateName").get("de").asText()
 			</a></td>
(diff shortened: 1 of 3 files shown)


acka47:
I noticed that `name` was not consistently used with language tags and adjusted this. Also added `type` information to each node. Please redeploy on staging, @fsteeg .

PR #919, opened 2018-09-18
851 etl loc bibframe works

### LINKED ISSUE #851: Set up experimental index with LoC Bibframe data?
From an [email to the Bibframe listserv](https://listserv.loc.gov/cgi-bin/wa?A2=BIBFRAME;3141fdaf.1806):

> The Library of Congress has had a second BIBFARME Pilot underway for a year, with 60 catalogers keying descriptions directly into a BIBFRAME editor as the initial step in cataloging a resource.   To support that pilot we converted the entire Library catalog to a BIBFRAME RDF structured file so that the Library’s Pilot catalogers would have the “usual” catalog against which to make cataloging decisions.  There have been numerous requests for access to that file from outside LC and we are working on that.  But in the meantime we are making available a bulk download of that file for systems that can load it and want to experiment.  The converted file is still continually being improved in the Pilot so this is just a snapshot as of May 24 and May 30. 

> The file can be accessed from the download pages of id.lov.gov:  go to http://www.loc.gov/bibframe/implementation/ and follow the link [BIBFRAME Works and Instances dataset](http://www.loc.gov/bibframe/implementation/pdf/bibframe-bulk-download.pdf) which will give you a “ReadMe” for these files, and the link to the ID bulk download page.  

I think we might be able to set up with modest resources an experimental index à la lobid-resources with this data for the community to play around with. What do you think?

dr0i:
Resolves #851.


fsteeg:
Since this is more of an initial experiment to get familiar with the data, and not implemented in the way we plan to do for the workshop, I don't think we should merge this into lobid-resources master, or even do a code review at this point.

dr0i:
After offline discussion: deploying this as it also provides some improvements for lobid-resources as a side effect.

dr0i:

PR #929, opened 2018-10-08
Switch to "insert" instead of "upsert"

### LINKED ISSUE #928: Update seems to be a merge rather then an overwrite
Reported by @jschnasse : 
http://stage.lobid.org/resources/HT019826954 differs from http://lobid.org/resources/HT019826954 in that the first has (wrongly) a doi. Both resources were modified at `"dateModified": "20181005"` and since nothing in the transfomatio process was changed both should therefore be equal. The only difference is that the first was _updated_ while the last was completely _reindexed_.
Question is:
Does the update process do a `merge` rather than a preferrable `overwrite`?

dr0i:
This ensures an overwrite of updated documents rather then a merge.

Resolves #928.


fsteeg:
+1

dr0i:

PR #959, opened 2019-01-29
Increase aggregation size for subject ID

fsteeg:
For hit counts in Sachsystematik

See https://github.com/hbz/nwbib/issues/453


dr0i:
+1

PR #969, opened 2019-02-12
Add division joy for multithreading

### LINKED ISSUE #967: Use more cpu cores when using metafacture
Atm the ETL uses only one cpu core. As we need more than 25h to ETL the whole hbz catalog it would be nice to quicken this by using more cores.

There are two ideas:
1. maybe the `ObjectTee` can be used for demuxing
2. one could provide a plugin demuxing objects

dr0i:
This adds the ObjectDivider which (in conjunction with an ObjectPipeDecoupler
and added to a Tee) creates a multithreaded object pipeline.

- add 7 threads for transformation and indexing catalog data

Resolves #967.


dr0i:
Reassigning myself since there are some minor things improvable, like declarations of variables and their naming, e.g. uppercasing. Readability of code can be improved but inserting line breaks.

dr0i:
Added a simplified way to make threads. Depends on hbz/metafacture-core#10.

fsteeg:
+1

Interesting stuff. Wondering if and where `ObjectPipeDecoupler` is/was used before.

dr0i:

dr0i:
Ack. I couldn't find anything. Also, no tests! It's probably used _somewhere_, but as that code is not online it's like that useful class doesn't exist at all.

Issue #968, opened 2019-02-05
Provide script for switching staging alias' to productive alias'


dr0i:
The switching of the staging indices of `resources` and `geo_nwbib` to production involves 4 alias' renaming. It's dumb and error prone to do this by hand one or two times a week, so a script (which could be started by hand after reviewing the staging indices) doing this would be nice.

dr0i:
As discussed offline: integrate sanity check e.g. if there are enough spatial entries (which would have helped to discover #970 before deploying).
Sanity checks should be configurable using a simple txt doc.

dr0i:
Run like this:

` hduser@weywot1:lobid-resources-stage$ mvn clean install ; mvn exec:java -Dexec.mainClass="org.lobid.resources.run.UpdateAliases"`

dr0i:
Added this to crontab:
every Monday at 11:00 the index will be switched automatically.
I think we are on the safe side to do this, even when the sanity checker (#975) is not implemented yet, because:
1. it very rarely occurred in the last half year that we not shouldn't manually switch the index at Monday
2. we have enough time to get a short glimpse on the sanity of the index manually till 11:00
3. it [is announced](https://wiki1.hbz-nrw.de/display/SEM/Datenaktualisierungsrhythmen) that the index should be updated at least at Monday midday, so now this cannot be forgotten, even if the main index-switcher (me ;)) is on vacation

PR #974, opened 2019-02-22
Add class for switching index aliases

dr0i:
This adds the alias "resources" to the newest index of "lobid-resources" and
the alias "resources-staging" to the second newest index of "lobid-resources".
The same goes for "geo_nwbib" and "geo_nwbib-staging".

I.e. the aliases found after a fulldump update are switched to go productive.

- improve some logic
- rename some variables to be more pecise in what they stand for

See #968.

### DIFF
--- a/src/main/java/org/lobid/resources/ElasticsearchIndexer.java
+++ b/src/main/java/org/lobid/resources/ElasticsearchIndexer.java
@@ -469,14 +469,37 @@ public class ElasticsearchIndexer
 				groupByIndexCollection(indexName);
 		for (String prefix : indices.keySet()) {
 			final SortedSet<String> indicesForPrefix = indices.get(prefix);
-			final String newIndex = indicesForPrefix.last();
+			final String newestIndex = indicesForPrefix.last();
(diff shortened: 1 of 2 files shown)


src/main/java/org/lobid/resources/ElasticsearchIndexer.java
```diff
@@ -469,16 +469,39 @@ public void updateAliases() {
...
+	/**
+	 * This adds the productive_prefix as alias to the newest index (which names
+	 * begins with productive_prefix) and adds the staging alias to the second
+	 * newest index name.
+	 * 
+	 * @param productive_prefix the productive alias (and also the prefix of the
+	 *          names of the indices)
+	 * @param staging the "-staging" alias
+	 */
+	public void swapProductionAndStagingAliases(final String productive_prefix,
```
fsteeg: To follow the naming conventions of the other variables, this should probably be `productivePrefix` (or, since it's actually the full productive alias, maybe just `productive`).

Otherwise, +1

dr0i:
Renamed the variable.

PR #982, opened 2019-03-19
Enforce literal for field 521-1.[at]

### LINKED ISSUE #979: MapperParsingException: containsExampleOfWork.label
> MapperParsingException at 2019-03-09:
weywot3:
[2019-03-09T04:17:46,237][DEBUG][o.e.a.b.TransportShardBulkAction] [weywot3] [resources-20190306-0100][1] failed to execute bulk item (index) BulkShardRequest [[resources-20190306-0100][1]] containing [1423] requests
org.elasticsearch.index.mapper.MapperParsingException: failed to parse [containsExampleOfWork.label]

#957 would help to identify the problematic resource, however we have no plans to switch update  to elasticsearch 6.x.
As a workaround I will build the fork for ES 5.6.3 as mentioned in #957, but only as a local instance, and index /data/DE-605/mabxml/updates//DE-605-aleph-update-marcxchange-20190308-20190309.tar.gz to identify the resource.

dr0i:
- add test resource

Resolves #979.


dr0i:
@acka47 there was a http-URI in the `label` field, making it an Object. So I forced the label to be a string by prefixing  "Siehe: "  as already done for other, similiar cases. Have a look if this may be a catalogue error for `521-1.[at]` resp. if it's ok to enforce the literal (it's at least _better_ as else the resource would be dropped as a whole): 

>   "containsExampleOfWork" : [ {
    "type" : [ "Work" ],
    "label" : "Siehe: http://www.loc.gov/catdir/description/prin031/2002066257.html"
  } ],

Though this doesn't seem to be correct the question is if it's worth the amount of work to fix this *one* field for *one* resource.

acka47:
+1

PR #984, opened 2019-04-01
Add sanity check before updating the index aliases

### LINKED ISSUE #975: Add sanity check in script when switching indices
From https://github.com/hbz/lobid-resources/issues/968#issuecomment-464694191
> As discussed offline: integrate sanity check e.g. if there are enough spatial entries (which would have helped to discover #970 before deploying).
Sanity checks should be configurable using a simple txt doc.

dr0i:
The "deletedCountTest" is facultative because it depends on the synchrony
of the creation of the fulldump and the creation of the index (must be the same
day). So if e.g. in the middle of the week a new staging index is build with an
4 day old fulldump the test may probably fail. So use "deletedCountTest" only
when doing the main etl procedure at the weekend.

Resolves #975.

### DIFF
--- a/src/main/java/org/lobid/resources/run/UpdateAliases.java
+++ b/src/main/java/org/lobid/resources/run/UpdateAliases.java
@@ -3,24 +3,33 @@
 package org.lobid.resources.run;
 
 import java.net.InetAddress;
-import java.net.UnknownHostException;
+import java.net.URL;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.elasticsearch.client.Client;
 import org.elasticsearch.client.transport.TransportClient;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.common.transport.InetSocketTransportAddress;
 import org.elasticsearch.transport.client.PreBuiltTransportClient;
+import org.junit.Assert;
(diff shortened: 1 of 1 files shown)


fsteeg:
+1

dr0i:
The comparison between deleted resources and the difference between the stage and the productive index was not succesful, so the sanity check failed:

> INFO - Difference between production (20874272) and staging:  (20871501) :2771
INFO - Amount of deletions between 20190403 and 20190407: 2297

This is because of https://github.com/hbz/lobid-resources/pull/984#issue-266267993:

>  it depends on the synchrony of the creation of the fulldump and the creation of the index

Right, so it's wrong to pull out the timestamp off the last productive index, because that could very well be the one created in the middle of the week. As there is no possibility to determine the creation of the fulldump when only taking provenance or index information into account, a simple, static way is to say "it's 7 days ago", as we know that to be true. Will provide that fix.

Note of despair:
http://lobid.org/resources/deletions/search?q=describedBy.deleted%3A[20190331+TO+20190406] reveals only `2770` instead of `2771`.

dr0i:
@fsteeg please review the changes in 74e76db96d04493ee7515034d4ab22c863af9259.

dr0i:
@fsteeg renamed the variable.

Issue #987, opened 2019-04-09
Report program status with java sending a mail


dr0i:
Following #975 , it would be nice to send directly program statuses out of java programs instead of using shell-scripts grepping through the log files.

Idea is to have one simple helper class which can be used by all programs, sending mail reports via localhost smtp. The latter is also the precondition which is satisfied on most (all?) our servers and otherwise easily installable.

PR #988, opened 2019-04-09
Add java mailing helper class

dr0i:
See #987.

### DIFF
--- a/src/main/java/de/hbz/lobid/helper/Email.java
+++ b/src/main/java/de/hbz/lobid/helper/Email.java
@@ -1,0 +1,68 @@
+/* Copyright 2019 hbz, Pascal Christoph. Licensed under the EPL 2.0*/
+package de.hbz.lobid.helper;
+
+import java.net.InetAddress;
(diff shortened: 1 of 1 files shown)


src/main/java/de/hbz/lobid/helper/Email.java
```diff
@@ -0,0 +1,68 @@
...
+	 * 
+	 * @author Pascal Christoph (dr0i)
+	 * 
+	 * @param FROM_NAME the name before the "@" of an email
+	 * @param TO_EMAIL the complete email address of the receiver
+	 * @param SUBJECT the subject of the mail
+	 * @param MESSAGE the plain text message of the mail
+	 */
+	static public void sendEmail(final String FROM_NAME, final String TO_EMAIL,
+			final String SUBJECT, final String MESSAGE) {
```
fsteeg: All caps is typically used for constants, not parameters. I'd suggest `fromName`, `toEmail`, `subject`, and `message`.

PR #995, opened 2019-05-13
Fix comparing of deletions

### LINKED ISSUE #994: Switching index: comparing amount of deletions broken
"x <= x + y" returns false.

dr0i:
- rename variables
- mark important key words in log messages

Resolves #994.

### DIFF
--- a/src/main/java/org/lobid/resources/run/UpdateAliases.java
+++ b/src/main/java/org/lobid/resources/run/UpdateAliases.java
@@ -119,20 +119,20 @@ public class UpdateAliases {
 		final String creationDateResourcesStagingMinusOneWeek = LocalDate
 				.parse(creationDateResourcesStaging, DateTimeFormatter.BASIC_ISO_DATE)
 				.minusDays(7).format(DateTimeFormatter.BASIC_ISO_DATE);
-		final int resourcesCount =
+		final int totalHitsOldIndex =
 				queryEsAndGetJNode("resources/resource/_search?q=*").at("/hits/total")
 						.asInt();
-		final int resourcesStagingCount =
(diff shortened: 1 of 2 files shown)


src/main/java/org/lobid/resources/run/UpdateAliases.java
```diff
@@ -145,13 +145,12 @@ private static void deletedCountTest() throws IOException {
...
-				("Going to compare if " + differenceDocsCountStagingVsProduction
-						+ " is less or equals " + deletionsCount
-						+ " allowing a 0.3% tolerance (i.e. " + tolerance + ")");
+		logMessage = ("Going to compare if " + differenceOfTotalHitsBetweenOldAndNew
+				+ " is less or equals " + deletionsCount
+				+ " allowing a 0.3% tolerance (i.e. " + tolerance + ")");
 		log(logMessage);
 		Assert.assertTrue(
-				deletionsCount - differenceDocsCountStagingVsProduction < tolerance);
+				differenceOfTotalHitsBetweenOldAndNew - deletionsCount < tolerance);
```
fsteeg: I'm not sure I understand, is the the fix to swap the operands? Maybe I missed something, but should we not compare the *absolute* value to cover both a slightly larger and a slightly smaller new index? Something like `Math.abs(differenceOfTotalHitsBetweenOldAndNew - deletionsCount) < tolerance`?

dr0i: Hm , I think it's okay to have (many) *more* docs than expected. It's a different situation if you have *less*.

fsteeg: Right, of course.

PR #997, opened 2019-05-17
Add SPARQL query to get info on hand-selected items

acka47:
See https://github.com/hbz/nwbib/issues/468#issuecomment-493189871. @dr0i, you will set the headers somewhere in the code to accept text/csv, I guess?

See #998.


dr0i:
Thx!

dr0i:
@acka47 hm, thinking about it: should this tsv substitute src/main/resources/getNwbibSubjectLocationsAsWikidataEntities.sparql ? Then, the shelling commit should be ignored. Also, it seems that this tsv should be somehow used. I guess this shall supplement the geo_nwbib index? How, so? Because formerly the QIDs got by  getNwbibSubjectLocationsAsWikidataEntities.sparql where indexed into geo_nwbib and the whole retrieving logic is build on that. Or was the idea now to get rid of the geo-index and build a tsv with 8200 entries (the ones from the geo-index and the ones from the tsv) and use this as a map via the morph (which would be more consistent but would also mean to rework a lot)? I didn't recall what we agreed upon.

acka47:
This sparql query is to regularly create an updated tsv as lookup table for metafacture with the entries from /src/main/resources/string2wikidata.tsv (which are not cover by src/main/resources/getNwbibSubjectLocationsAsWikidataEntities.sparql) so that we can create a `focus` object for those resources.

dr0i:
Ok, let's talk about that to get things straight, also re the MbXmlClob you send me.

PR #1003, opened 2019-06-11
Use char-filter to remove hyphens from isbn/issn

### LINKED ISSUE #201: Set up custom analyzer for ISBN and ISSN
Originates from https://github.com/hbz/lobid-resources-web/issues/33

We should set up a custom analyzer [1] with a char filter (probably a pattern replace char filter, [2]) that removes dashes and blanks for the `isbn` and `issn` fields. Since we will keep the preprocessing in Metafacture, it should suffice to switch only the `search_analyzer` to the custom analyzer.

[1] https://www.elastic.co/guide/en/elasticsearch/reference/2.4/analysis-custom-analyzer.html
[2] https://www.elastic.co/guide/en/elasticsearch/reference/2.4/analysis-pattern-replace-charfilter.html

dr0i:
Resolves #201.

### DIFF
--- a/src/main/resources/index-config.json
+++ b/src/main/resources/index-config.json
@@ -36,6 +36,10 @@
                   ],
                   "tokenizer" : "letter",
                   "type" : "custom"
+               },
+               "hyphen_analyzer": {
+                  "tokenizer": "standard",
+                  "char_filter": "remove-hyphens"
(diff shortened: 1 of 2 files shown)


dr0i:
Deployed to production, closing.

PR #1006, opened 2019-07-04
Fix string2 WDmapping for "Lippischer Wald"

acka47:


dr0i:
+1

dr0i:
Going to deploy, check next Monday.

acka47:
It is now correct in lobid-resources but NWBib has to be adjusted as well, see https://nwbib.de/BT000059116 and https://nwbib.de/spatial#Q31312912. Will add a reminder to an NWBib issue.

PR #1024, opened 2019-09-17
1011 remove superordinated map

dr0i:
See #1011.


dr0i:
Deployed small test, see http://test.lobid.org/resources/search?q=map.

Issue #1029, opened 2019-10-01
Add missing foaf:focus statements to data


acka47:
Similar to #1017. I stumbled over this while writing a post about NWbib data for Coding da Vinci and wanted to show that most entries have links to places in Wikidata where you can get additional information.

Task: We need to pull all the `foaf:focus` statements out of the Turtle vocab and add it to the lobid-resources data. http://lobid.org/resources/HT020170194.json is an example where the statement is missing:

```json
{
    "spatial":[
        {
            "id":"https://nwbib.de/spatial#N62",
            "type":[
                "Concept"
            ],
            "source":{
                "id":"https://nwbib.de/spatial",
                "label":"Raumsystematik der Nordrhein-Westfälischen Bibliographie"
            },
            "notation":"62",
            "label":"Herzogtum Westfalen"
        }
    ]
}
```

In the turtle file, the `foaf:focus` statements exists:

```ttl
nwbib-spatial:N62  a    skos:Concept ;
        skos:broader    nwbib-spatial:N4-7 ;
        skos:inScheme   <https://nwbib.de/spatial> ;
        skos:notation   "62" ;
        skos:prefLabel  "Herzogtum Westfalen"@de ;
        foaf:focus      wd:Q657241 .
```

acka47:
Looks good in the test files. However, looking at this, it made me think whether adding a geo point for the larger regions (NRW, Rheinland, Westfalen, probably others) makes actually sense. (We already discussed this several times in the past. E.g. for the map on the start page (https://nwbib.de), all entries with NRW, 01 as spatial subject would be locate din Kreis Unna...  Maybe discuss a possible solution at the standup?

dr0i:
Deployed to staging, see e.g. http://stage.lobid.org/resources/search?q=spatial.focus.id%3A%22http%3A%2F%2Fwww.wikidata.org%2Fentity%2FQ152243%22 .

dr0i:
@acka47 as discussed yesterday at scrum, @fsteeg will filter out `focus` entries based on its type, before we can merge and deploy.

acka47:
[...]

Issue #1030, opened 2019-10-04
NullPointer Exception using more Threads than documents


dr0i:
Experienced this when shrinking the hbz01XmlClobs.tar.bz2 test set to one, while there are two threads to compute the data. When calling `closeStream()`, a NullPointer Exception appears. This stops the ETL process finishing correctly, and thus no resource (aka document) is serialized to the disk.
A workaround is to catch the Exception.
However, as this is a problem can occur everywhere, it would be nicer to somehow control this at the modules in metafacture-core, probably at the [ObjectPipeDecoupler](https://github.com/metafacture/metafacture-core/blob/cb43933ec8eb01a4ddce4019c14b2415cc441918/metafacture-flowcontrol/src/main/java/org/metafacture/flowcontrol/ObjectPipeDecoupler.java). Also, to say the least, it's rather unexpected to use more threads than documents, it _may_ happen.

acka47:
@dr0i Does this problem still exist? Shouldn't we open issue over at metafacture-core then and close this one?

dr0i:
Problem still exists. Opened https://github.com/metafacture/metafacture-core/issues/466. Closing this one.

PR #1031, opened 2019-10-04
1029 add focus nwbib spatial

dr0i:
Adds a focus object if the nwbib spatial object resides in nwbib-spatial.tsv.

See #1029.

### DIFF
--- a/src/test/java/org/lobid/resources/Hbz01MabXml2ElasticsearchLobidTest.java
+++ b/src/test/java/org/lobid/resources/Hbz01MabXml2ElasticsearchLobidTest.java
@@ -79,7 +79,7 @@ public final class Hbz01MabXml2ElasticsearchLobidTest {
 	@BeforeClass
 	public static void setup() {
 		try {
-			if (System.getProperty("generateTestData", "false").equals("true")) {
+			if (System.getProperty("generateTestData", "true").equals("true")) {
 				Files.walk(Paths.get(DIRECTORY_TO_TEST_JSON_FILES))
 						.filter(Files::isRegularFile)
 						.forEach(fname -> testFiles.add(fname.getFileName().toString()));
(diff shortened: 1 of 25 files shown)


src/test/java/org/lobid/resources/Hbz01MabXml2ElasticsearchLobidTest.java
```diff
@@ -128,7 +128,11 @@ static void etl() {
 		opener.process(
 				new File(Hbz01MabXmlEtlNtriples2Filesystem.TEST_FILENAME_ALEPHXMLCLOBS)
 						.getAbsolutePath());
-		opener.closeStream();
+		try {
+			opener.closeStream();
+		} catch (Exception e) {
+			// ignore, see #1030
```
fsteeg: This will catch and ignore anything that could go wrong here, right?

We could catch ` NullPointerException` only to narrow it down a bit.

Is there way to test here if we actually have more threads than documents?

And I'd suggest we use the full issue URL (`https://github.com/hbz/lobid-resources/issues/1030`) for #1030 to make it unambiguous and easier to access from the code.

dr0i: Added a commit to catch NPE and use URI as reference in the comment.
> Is there way to test here if we actually have more threads than documents?

I wouldn't know of that, because the number of processed documents will only be known when the etl is finished, while the threads are already set. One could think of a preprocessing pipeline, but what would be the point?  I think the real fix would be to somehow make the metafacture library more robust like mentioned in #1030.

dr0i:
note: after deploying to staging, assign @fsteeg for he has to filter out some geo data based on the type of the `focus` object.

dr0i:
Deployed to staging. Assigning @fsteeg to adapt web UI.

dr0i:

Issue #1032, opened 2019-10-07
Missing "date" in index-update script; improve waiting when rebuilding geo index


dr0i:
Automatic index switching doesn't work.

dr0i:
Seems to work. Closing.

PR #1033, opened 2019-10-07
Set date variable and improve waiting rebuilding geo index

dr0i:
Date variable was not set.
Improved waiting of rebuilding the geo index, if it failed to be ok:
starting with 15 minutes, then 2h, then 6h45m, 16h16.

See #1032.


dr0i:
Deployed, closed.

PR #1045, opened 2019-12-10
Nwbib 507 add700.1a 03

dr0i:
See hbz/nwbib#507.


dr0i:
Only the `coverage` field is added, not the `spatial` because the wikidata-entity would have to be part of the local geo-index. But when used with our geo-nwbib-index it seems to be ok:
 ```curl weywot5.hbz-nrw.de:9200/resources-smalltest-20191210-133448/_search?q=coverage:Siebengebirge | jq .```

dr0i:
Deployed to staging, see http://stage.lobid.org/resources/search?q=coverage%3ASiebengebirge.
This seems wrong: "id": "https://nwbib.de/spatial#Q4236" yes @acka47 ?

acka47:
> This seems wrong: "id": "https://nwbib.de/spatial#Q4236" 

No, exactly right: https://www.wikidata.org/wiki/Q4236

dr0i:
Deployed, closing.
@acka47 has to update the turtle.

PR #1047, opened 2020-01-08
Improve tests, remove `stop` filter for default analyzer

fsteeg:
Will resolve https://github.com/hbz/nwbib/issues/506

@dr0i: This only affects the index config (except tests). Could you deploy this to (small) test and stage?

### DIFF
--- a/src/main/resources/index-config.json
+++ b/src/main/resources/index-config.json
@@ -7,7 +7,6 @@
                "german_analyzer" : {
                   "filter" : [
                      "lowercase",
-                     "stop",
                      "snowball"
                   ],
                   "tokenizer" : "standard",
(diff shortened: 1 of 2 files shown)


dr0i:
Done, see https://test.lobid.org/resources/search?q=*.

dr0i:
Do you really need also the "stage" for it? That would be the whole catalog, and as we have the end of the week, if the result is no good, we loose the normally fine rebuild at the beginning of the next week. What I want to say:
Please test the smalltest *before* stage-deployement.

fsteeg:
Looks good:

http://test.lobid.org/resources/search?word=m%C3%BClheim+AND+an+AND+der+AND+ruhr
http://stage.lobid.org/resources/search?word=m%C3%BClheim+AND+an+AND+der+AND+ruhr

We get a hit on test. Could you deploy to stage?

dr0i:
Already switched automatically to production. Seems good, see http://lobid.org/resources/search?word=m%C3%BClheim+AND+an+AND+der+AND+ruhr vs. the old one http://stage.lobid.org/resources/search?word=m%C3%BClheim+AND+an+AND+der+AND+ruhr.

dr0i:
just a note:
yes , it was switched, but I didn't merged it into master, i.e. daily updates on production never used these commits.
Merged just now.

Issue #1052, opened 2020-01-30
Collection from Union Catalogue missed in lobid


hagbeck:
In the union catalogue exists collection i.e. for e-books which are not starting with "ZDB". These are missed in lobid.

In http://lobid.org/resources/HT020022241.json there is the collection name "cuvillier". Other collection names can be found in https://service-wiki.hbz-nrw.de/display/VDBE/Produktsigel+und+interne+Selektionskennzeichen

acka47:
Thanks for the request. Collection information is in `078` in the source data, from the example http://lobid.org/hbz01/HT020022241:

```xml
<datafield tag="078" ind1="e" ind2="1">
  <subfield code="a">cuvillier</subfield>
</datafield>
```

Currently, we are only transforming some dedicated collection information (NWBib, Edoweb, FRL, ZDB) to RDF. We will have to think about a way for adding generic collection information. Maybe with a bnode, so that the resulting JSON could look like this:

```json
{
    "inCollection": [
        {
            "label": "cuvillier"
        }
    ]
}
```

However, before approaching this it would be nice to know how many different collections are named in `078` and what their tags are.

dr0i:
> However, before approaching this it would be nice to know how many different collections are named in 078 and what their tags are.

We could do this with ES `aggs` but this would mean to explicitly allow aggregations over labels (meaning: new index config, and performance impact).  Doing this on the 36k smallTest reveals just:

> {
          "key" : "Zeitschriftendatenbank (ZDB)",
          "doc_count" : 163
        },
        {
          "key" : "eResource package",
          "doc_count" : 129
        },
        {
          "key" : "Nordrhein-Westfälische Bibliographie (NWBib)",
          "doc_count" : 123
        },
        {
          "key" : "Edoweb Rheinland-Pfalz",
          "doc_count" : 16
        },
        {
          "key" : "Elektronische Zeitschriftenbibliothek (EZB)",
          "doc_count" : 10
        },
        {
          "key" : "Fachrepositorium Lebenswissenschaften",
          "doc_count" : 10
        },
        {
          "key" : "Rheinland-Pfälzische Bibliographie",
          "doc_count" : 8
        }
so you may still want to have a complete list on the whole index?

acka47:
> However, before approaching this it would be nice to know how many different collections are named in 078 and what their tags are.

The list with all the collections is linked in the original issue comment:
[...]

PR #1060, opened 2020-03-13
Add collections also when not an Isil

dr0i:
See #1052.


acka47:
> @acka47 here you see how to define a label for a URI. Note that many of these definitions are missing - and you may want to provide them :)

Done

dr0i:
Oh, this was fast! I am going tu update the tests and prepare fullindexing on staging. We can then extract the collections we don't have covered in `labels` yet.

dr0i:
This is on production. Closing.

Issue #1089, opened 2020-11-27
Transform Alma MARC21 XML to lobid JSON-LD


dr0i:
At some point in the future there will be no more Aleph and thus no more MAB XML as our primary data input. It will be substituted by Alma's MARC21 XML. The transition is a hybrid approach where both kinds of data are indexed into the same index.

Based on  https://github.com/hbz/limetrans/compare/alma provided by @blackwinter .

dr0i:
Don't know if a PR should be opened. @acka47 see what's produced atm and if that would be reasonable enough to start a PR : https://github.com/hbz/lobid-resources/tree/almaMarc21XmlToLobidJson/src/test/resources/alma.

acka47:
I think this already looks very good! A PR would be good I think. I think we should ASAP get this into a status where we can include this in lobid-resources so that @hagbeck can start testing and we have the possibility to have a few more iterations with Hans-Georg in the loop.

Some comments on the output data:
- As far as I can see without going deeper into it, only the metametadata needs some more work (`sourceOrganisation` etc.).
- Looking at https://github.com/hbz/lobid-resources/blob/b446f785ebc98b37ee41c300fc78e6b9462ed7fe/src/test/resources/alma/HT012734833.json#L5-L33 and https://github.com/hbz/lobid-resources/blob/b446f785ebc98b37ee41c300fc78e6b9462ed7fe/src/test/resources/alma/KUR00770801.json#L6-L13 I see it is indicated with `heldBy` that an organization has obtained a license to access an online resource. This absolutely makes sense as we don't have to adjust our data model for licensed online resources and lobid users will presumably be able to cover all the use cases with this approach.
- We should also check out the Bibframe RDF that Alma provides for the resources. It might give us some ideas on what is missing or what might be handled in another way.

dr0i:
I would say we go to close this issue.
Opened #1095. Go on opening new issues  when identified.
Ok @acka47 ?

acka47:
+1

Issue #1021, opened 2019-09-13
Ensure the same automatical xml formatting for at least eclipse


dr0i:
The howto of formatting of the xml is still a think to do.
As a starter, shipping an xml-formatter for only one IDE, eclipse, should be doable.

fsteeg:
I think the automation was the problem here, right? You can always manually set up the same options in the XML > XML Files > Editor preferences.

dr0i:
At least shipping this preferences to ensure all are using the same formatting configs if manually formatting xml. Then, eclipse can be configured to "automatically format when saving".
**Identatation size** should be **1**.
There are other setting like **line width: 121** atm and "insert whitespace before closing tags" etc.
In metafacture-core there is also a problem with formatting, also in the code, so this would be a starter. Just opened https://github.com/metafacture/metafacture-core/issues/307.

dr0i:
We should go like metafacture, using .editorconfig. See https://github.com/metafacture/metafacture-core/pull/388.

TobiasNx:
@dr0i still relevant?

PR #1094, opened 2020-12-14
Alma marc21 xml to lobid json

dr0i:
Initial transformation of Alma Marc21 XML to lobid JSON-LD.

See https://github.com/hbz/lobid-resources/issues/1089.


fsteeg (approved):
Looks good on the code level as far as I can tell. Just three things I noticed:

- Formatting in the Morph files is often inconsistent, perhaps some automatic formatting and/or checks might be useful here?
- I'm surprised to see metamorph.xsd as part of the repo and the diff, but I think you talked about that in a recent standup?
- If I understand correctly, the JSON validation via script is not part of the build? But we'll get that anyway with https://github.com/metafacture/metafacture-core/issues/347

dr0i:
Re formatting: this pops https://github.com/hbz/lobid-resources/issues/1021 up
Re metamorph.xsd : opened the issue https://github.com/hbz/lobid-resources/issues/1097
Re JSON validation: right

dr0i:
Closed.

PR #1096, opened 2020-12-14
Extend JSON Schema to cover a lot more fields

acka47:
Resolves #1079

Issue #1095, opened 2020-12-14
Add sourceOrganization


dr0i:
Taken from https://github.com/hbz/lobid-resources/issues/1089#issuecomment-741625434 :
`sourceOrganisation` is missing yet.

Question is:
Where to get the information of the institution who modified or created the resource ? And how map this to ISIL ?
 In principal, this should be stored in `MNG` ("Management Information")
> $$a = "created by" ; $$b = "Create date";  $$c = "updated by";  $$d = "Update date", $$g = "Originating system ID"

but these look atm not promising (see e.g. https://github.com/hbz/lobid-resources/pull/1094/commits/c57ef745c9de6853eae54c503557e5d7e3ef7b55#diff-93a0ac06390d531a7e79c7a83d77cde3b0016bf81a2376e32fe8c76c7ac3780eR142 where `MNG a="import"` and `MNG c` is missing for all the data)

As the data is still in beta this information is likely to be there at the end of the beta stadium, but then it would be nice to know which kind of identifier will be used and how to map it to get the ISIL.

acka47:
I'll add some background and reference to MAB-MARC mapping and to MARC21 docs to have a better informed discussion.

## Background (MAB transformation)

We added `sourceOrganization`, `provider` & `modifiedBy` in #455. The mapped MAB2 fields are documented as follows:

```
070       IDENTIFIZIERUNGSMERKMALE DER BEARBEITENDEN INSTITUTION

          Indikator:
          blank = Kennzeichen der katalogisierenden Institution
          a     = Kennzeichen der liefernden Institution
          b     = Kennzeichen der korrigierenden Institution
``` 

The mapping is:

- `070`-  -> `sourceOrganisation`
- `070a` -> `provider`
- `070b` ->  `modifiedBy`

## Transfer to MARC21 

The corresponding fields in MARC21 are (according to [Konkordanz MAB2 – MARC 21 from DNB](https://d-nb.info/988382318/34), see also [MARC21 docs](https://www.loc.gov/marc/bibliographic/bd040.html)):

- `040$a`: Original cataloging agency (corresponds with `070-`/`sourceOrganisation` "Kennzeichen der katalogisierenden Institution")
- `040$c`: Transcribing agency (corresponds with `070a`/`provider` "Kennzeichen der liefernden Institution")
- `040$d`: Modifying agency (corresponds with  `070b`/`modifiedBy` "Kennzeichen der korrigierenden Institution")

[...]

PR #1100, opened 2020-12-15
1095 source organization

dr0i:
This should be merged after #1096 to have the json validate script correctly testing. This PR validates at least the `sourceOrganization` ok.

This PR introduces a resource that is originating from Aleph AND from Alma, which gives us quite a good template of comparing.

See #1095.


dr0i:
Ah, and I still don't think we will solve the issue as data is still in beta.
Maybe we should go on merging and let #1095 open?

src/main/resources/alma/common/describedBy.xml
```diff
@@ -82,9 +82,14 @@
 			</entity>
 		</entity>
 		<entity name="sourceOrganization">
-			<data name="id" source="MNG  .g">
-				<regexp match="(.*)" format="http://lobid.org/organisations/${1}_todo"/>
-			</data>
+			<choose>
+				<data name="id" source="040  .a">
+					<regexp match="(.*)" format="http://lobid.org/organisations/${1}"/>
```
acka47, line 87:
```suggestion
					<regexp match="(.*)" format="http://lobid.org/organisations/DE-${1}"/>
```
...to build correct lobid-organisations URIs.

src/main/resources/alma/common/describedBy.xml
```diff
@@ -82,9 +82,14 @@
...
+				</data>
+				<data name="id" source="MNG  .g">
+					<regexp match="(.*)" format="http://lobid.org/organisations/${1}"/>
```
acka47, line 90:
```suggestion
					<regexp match="(.*)" format="http://lobid.org/organisations/DE-${1}"/>
```

src/main/resources/alma/common/describedBy.xml
```diff
@@ -82,9 +82,14 @@
 			</entity>
 		</entity>
 		<entity name="sourceOrganization">
-			<data name="id" source="MNG  .g">
```
acka47, line 85: Are you sure that those `MNG` fields contain the information we are looking for here?

src/test/resources/alma/KUR00770801.json
```diff
@@ -93,6 +89,10 @@
     "license" : [ {
       "id" : "http://creativecommons.org/publicdomain/zero/1.0",
       "label" : "Creative Commons-Lizenz CC0 1.0 Universal"
-    } ]
+    } ],
+    "sourceOrganization" : {
+      "label" : "lobid Organisation",
+      "id" : "http://lobid.org/organisations/004250681-DUE01"
```
acka47, line 95: These (here: `004250681-DUE01`) look like ALma-intenral identifiers. Is there a mapping to Sigel/ISILs?

acka47 (changes requested):
Some things have to be changed and some answers to be found.

src/main/resources/alma/common/describedBy.xml
dr0i: +1 and suffixing #!

src/main/resources/alma/common/describedBy.xml
dr0i: changed to "MNG  .a"

src/test/resources/alma/KUR00770801.json
dr0i: Did "grep 681-DUE01 " on a 500MB alma Marc export and it just revealed this one entry, so I assume it's not mapable. There is indeed [alma-institution-code-to-isil.xml](https://github.com/hbz/lobid-resources/blob/master/src/main/resources/alma/maps/alma-institution-code-to-isil.xml) but it has an entry like  `"a 6443" value="DE-61"` so this doesn't help here.

src/main/resources/alma/common/describedBy.xml
dr0i: Resolved in https://github.com/hbz/lobid-resources/pull/1100/commits/3c3109845839d9c49051c7f48bac0202cf0bd5ad#diff-946dcf62c20237a36122ec3ee9ef389c4c4b06f9e8c061e53e87543d42640236R89.

src/test/resources/alma/KUR00770801.json
dr0i: Hm, I found that `004250681` does not only exist in `MNG  .g="004250681-DUE01"` but is also part of the identifier `MBD  .i="990042506810206443"` which ends in `6443` which maps to `DE-61`. This is strange  because `MBD` stands for `MemberDetails`. I would at least have expected the  `6443` part of `MNG  .g`. Will ask our Alma experts.

dr0i: Also note that `MBD`  will probably be no more mandatory, we can say this after testload2. If it's not mandatory we cannot rely on it, obviously.

dr0i: @blackwinter do you know more about how to get the ISIL from `MNG .g` resp. why the `6443` is missing there?

blackwinter: Don't know about MNG, but as you already alluded to, we're [mapping](https://github.com/hbz/limetrans/blob/b31230f/src/main/resources/transformation/macros.xml#L83) the ISIL from the MMS ID's institution code.

blackwinter: Also, `MNG$$g` is the "Originating system ID". It's the record's ID *in* the originating system, not the ID *of* the system.

dr0i: Hm, good pointer, if not entirely correct:  `002136771-HBZ01` is a mishmash of `aleph-id` and an identifier _of_ the originating system - at least of the name of the database (`HBZ01`), not necessarily the name/isil of the institution.
However, it's true that `MNG$$g` shouldn't be used to determine the originating system.

dr0i:
As sometimes fields `040  .[acd]` are missing it's good to have a fallback.
But can we utilize `MNG` at all?
- `MNG  .a ` "created by" doesn't denote an organisation as the value is always "import")
- `MNG  .c` "updatedBy" doesn't denote an organization as the value is often "NON_SFX_CREATOR"
- `MNG  .g` is mishmash of database name and database internal id

So, MNG doesn't help here at all.

src/test/resources/alma/KUR00770801.json
blackwinter: Ah, right, totally forgot to go into that. The same ID is included in `035$$a` so I assume that it is in fact an ID of some system and that that system might be the SISIS importer (or something like that). It's true that it isn't - in its entirety - the ID from the actual SISIS system.

acka47 (approved):
+1 Looks good now.

Re. fallback for missing `040 .[acd]`: I don't see a problem with some records missing the information. Currently, the fields are also sometimes missing in lobid-resources:
- [2% are missing `describedBy.sourceOrganization`](http://lobid.org/resources/search?q=NOT+_exists_%3AdescribedBy.sourceOrganization)
- [63,4% are missing `describedBy.modifiedBy`](http://lobid.org/resources/search?q=NOT+_exists_%3AdescribedBy.modifiedBy)
- [68,4% are missing `describedBy.provider`](http://lobid.org/resources/search?q=NOT+_exists_%3AdescribedBy.provider)

PR #1101, opened 2021-01-05
Add extract xml test records

dr0i:
This gets single records out of a big xml file. It is used to easily
update the test marc xml sources.

- add install metafacture-core master SNAPSHOT locally
- use 'null' namespace in MarcXmlHandler
- update pom.xml
- add single xml tar file consisting of the test xml records

Note that Umlaute are XML escaped, but this is ok and will get us Umlaute back again when serializing as json.
Note that of convenience two `xmlElementSplitter` are used, didn't get an other pipeline working, it's working so.


fsteeg (approved):

dr0i:
Deployed, closed.

Issue #1110, opened 2021-02-01
Provide proper test set: e-ressources from Alma


dr0i:
Prioritizing e-ressources over everything else, we need a proper test set.

Originates from https://github.com/hbz/lobid-resources/issues/1102.

dr0i:
From 614e4163221b39f40ef8ee9e7d7562b4a444fac9:

>1.  copy your tar.bz2 (actually, it can also be a tar.gz as provided by alma-publishing. The name does not matter (so tar.gz can be renamed to tar.bz2 - it will nonetheless work)) test file to src/test/resources/alma/HT012734833_etAl.xml.tar.bz2
>2.   run only the alma test: "mvn failsafe:integration-test -Dit.test=AlmaMarc21XmlToLobidJsonTest"

As long as we have not direct access to single Alma Marc Xml records we have to go through (in the worst case: all) NZ-published dumps to get the updated source files which are part of [our test set](https://github.com/hbz/lobid-resources/tree/master/src/test/resources/alma). It is therefore, and not only therefore, good use to add  a test file only if it actually covers something new.

Closing.

dr0i:
With #1102 we got direct  access to Alma Marc Xml via an API.
I  am going to update the test data using this via API calls (https://indexes.devel.digibib.net/export/$MMSID).

PR #1111, opened 2021-02-05
Extracts portfolio (aka E-resources) from test data

dr0i:
1. copy your tar.bz2 (actually, it can also be a tar.gz as provided by alma-publishing.  The name does not matter (so tar.gz can be renamed to tar.bz2 - it will nonetheless work)) test file to src/test/resources/alma/HT012734833_etAl.xml.tar.bz2
2. run only the alma test: "mvn failsafe:integration-test -Dit.test=AlmaMarc21XmlToLobidJsonTest"

See #1110.


dr0i:
@TobiasNx as discussed, this is the branch which deals solely with #1110, i.e. providing  e-ressource's.
The commits from the branch "1110-provideEressourcesTestSet" should be cherry picked into a new branch.
Going to merge this branch and going to update #1110 to include more background information of the problems with getting test resources for the purpose of reference.

dr0i:
Added one portfolio E-resource.
As discussed, we should increase our test set only if it actually covers something new.

PR #1117, opened 2021-02-10
Delete note-Element

TobiasNx:
The transformation to "note" is creating problems in the mapping.

Since the information in "note" is not coherent we decided 
off-board that at the moment there is no need for "note".

### DIFF
--- a/src/main/resources/alma/common/fields.xml
+++ b/src/main/resources/alma/common/fields.xml
@@ -558,10 +558,7 @@
 
 
   <!-- 5xx -->
-  <data name="note" source="500  .a" />
-  <entity name="abstract[]" flushWith="record">
-    <data name="" source="520[ 13] .[ab]" />
-  </entity>
+
 
   <!-- 6xx -->
 	<!--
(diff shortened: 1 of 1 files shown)


src/main/resources/alma/common/fields.xml
```diff
@@ -558,10 +558,7 @@
 
 
   <!-- 5xx -->
-  <data name="note" source="500  .a" />
-  <entity name="abstract[]" flushWith="record">
```
dr0i, line 562: Why remove "abstract"?

src/main/resources/alma/common/fields.xml
```diff
@@ -579,7 +576,8 @@
       <data name="prefix" source="77[356]??.i">
         <replace pattern=":?$" with="" />
       </data>
-      <entity name="note[]" flushWith="77[356]??" sameEntity="true">
```
dr0i, line 582: Don't understand this change (also, it is commented).

dr0i (changes requested):

Issue #1084, opened 2020-10-20
"note" sometimes with array, mostly not


acka47:
An example with array is HT018857620:
https://github.com/hbz/lobid-resources/blob/a41f1711b34ece8c5f447501425bff098747fc92/src/test/resources/jsonld/HT018857620#L110

We should make it an array everywhere I guess. (But I will make it string in the JSON schema for now.)

acka47:
I think this could easily be solved by adjusting the context (same as #1098 #1083). Will make a PR.

Issue #1009, opened 2019-07-31
Missing data in note field


hagbeck:
In the record HT020008571 the value "Seite 76-80" in the note field is not complete. 
The value in the original footnote field starts with "Literaturverzeichnis: ". 
So I expect "Literaturverzeichnis: Seite 76-80".

acka47:
Thanks for the report. We will deal with it as soon as vacation time is over. From the source:

```xml
<datafield tag="501" ind1="-" ind2="1">
  <subfield code="p">Literaturverzeichnis</subfield>
  <subfield code="a">Seite 76-80</subfield>
</datafield>
```

Currently, we only use subfield `a` for the `note` field. See the morph:

https://github.com/hbz/lobid-resources/blob/397c648375ea444fdef1885f7d6e061e3865f1b5/src/main/resources/morph-hbz01-to-lobid.xml#L7555-L7559

dr0i:
@hagbeck please have a look, data should be there now.

fsteeg:
Closing, see: http://lobid.org/resources/HT020008571.json

Issue #755, opened 2018-03-05
Missing note field


ChristophEwertowski:
A FRL customer asks for a missing note field in [HT019488785](http://lobid.org/resources/HT019488785). At the moment we use just the first result of a `choose` element but since multiple entries of MAB 501 are allowed we should use all contents of this MAB field.

Related jira issue: https://jira.hbz-nrw.de/browse/FRL-87

ChristophEwertowski:
Deployed to test. See [HT018857620 (production, json)](http://lobid.org/resources/HT018857620.json) / [HT018857620 (test, json)](http://test.lobid.org/resources/HT018857620.json).

acka47:
Is it possible (with little effort) to retain the original order in the array?

ChristophEwertowski:
Not with little effort. I remember that at some place I concatted contents from MAB fields with the same number and then splitted the string but that's too much work for the `note` field.

acka47:
Ok then, +1.

PR #1118, opened 2021-02-11
Add new testfile and run maven

TobiasNx:
Saw that note was requested #1009,  #755 and it should be an array #1084.
So i have withdrawn my other pr (#1117) and changed the approach.

note is now an array and filters out duplicate values in the string-array.


src/test/resources/alma/HT005207972.json
```diff
@@ -107,7 +107,7 @@
     },
     "resultOf" : {
       "type" : [ "CreateAction" ],
-      "endTime" : "2021-01-05T14:18:03",
```
dr0i, line 110: While these data indeed changes, I recommend to not add it: the java unit tests ignores the "endTime", so test are successful - and if we don't have these resources in our commits the diff is smaller. I know this may be confusing, but the tests uses the productive morphs where this field is created.
Maybe it would be better to filter this field out before writing these test files so that they wouldn't appear here. Feel free to open an issue if this would be made things much easier.

dr0i: We have two options:
1. remove these modifications in the commit and force push them
2. leave it as it is and just remember to not add them in the future

web/conf/context.jsonld
```diff
@@ -363,8 +361,7 @@
       "@id" : "https://d-nb.info/standards/elementset/gnd#SubjectHeading"
     },
     "note" : {
-      "@id" : "http://id.loc.gov/ontologies/bibframe/note",
-      "@container": "@set"
```
dr0i, line 367: Declaring @conatiner as @set was correct, it shouldn't be removed!
I am admittedly confused to see the tests build succesfully nonetheless :(
Shoudl be further investigated in a new issue.

dr0i: Ah, I see: web/conf/context.jsonld is build automatically using the context-labels.json. And the latter wasn't updated since end of 2019.  But web/conf/context.jsonld was updated multiple times then. This is not good because changing web/conf/context.jsonld will be overriden by context-labels.json. Phew. Will make an issue.

dr0i (commented):
Comments inline.
Commited updates of "note" into labels and context and schema.
Also going to open a new issue concerning jsonld-context.

src/test/resources/alma/HT005207972.json
TobiasNx: How could I prevent this in the future since the endTime is always changing when testing?

web/conf/context.jsonld
acka47: Thanks for the reminder. I and obviously you too, totally forgot how it works.

src/test/resources/alma/HT005207972.json
dr0i: You can't. Just don't add them if you see that only the endTime was changed.

TobiasNx: ah, okay, I see what you mean. I should only add those testfiles that change concerning the issue that I am working at. Then lets keep it like this, for this time. Next time I will only `git add` the relevant files.

dr0i: +1

Issue #1116, opened 2021-02-10
Map key POR-tfolio-information to "heldBy"


TobiasNx:
The ALMA-XML is carrying holding information in field "POR". UB Dortmund needs this information.
We could add this information to ` "hasItem"` - objects that already exist in lobid resources.

e.g. HT012734833

```JSON
  "hasItem" : [ {
    "type" : [ "Item" ],
    "heldBy" : {
      "id" : "http://lobid.org/organisations/DE-605#!",
      "label" : "lobid Organisation"
    },
    "id" : "https://lobid.org/item/990108873860206441"
  }, {
    "type" : [ "Item" ],
    "heldBy" : {
      "id" : "http://lobid.org/organisations/DE-A96#!",
      "label" : "lobid Organisation"
    },
    "id" : "https://lobid.org/item/991002245659706444"
  }, {
    "type" : [ "Item" ],
    "heldBy" : {
      "id" : "http://lobid.org/organisations/DE-290#!",
      "label" : "lobid Organisation"
    },
    "id" : "https://lobid.org/item/991000300959706445"
  } ]
```

```XML
    <datafield tag="POR" ind1=" " ind2=" ">
      <subfield code="x">System</subfield>
      <subfield code="a">5333897520006444</subfield>
      <subfield code="M">49HBZ_FHA</subfield>
      <subfield code="c">static</subfield>
      <subfield code="l">Deutschlandweit zug&#228;nglich</subfield>
      <subfield code="y">2020-07-23 14:31:47 Europe/Berlin</subfield>
      <subfield code="w">2020-07-23 14:31:42 Europe/Berlin</subfield>
      <subfield code="D">https://eu04.alma.exlibrisgroup.com/view/uresolver/49HBZ_NETWORK/openurl?u.ignore_date_coverage=true&amp;portfolio_pid=5333897520006444&amp;Force_direct=true</subfield>
      <subfield code="v">P2E_JOB</subfield>
      <subfield code="z">2020-07-23 12:31:42</subfield>
      <subfield code="i">false</subfield>
      <subfield code="S">5233897530006444</subfield>
      <subfield code="d">e?u.ignore_date_coverage=true&amp;rft.mms_id=991002245659706444</subfield>
      <subfield code="b">Available</subfield>
      <subfield code="f">BOOK</subfield>
      <subfield code="8">5333897520006444</subfield>
    </datafield>
    <datafield tag="POR" ind1=" " ind2=" ">
[...]

Issue #1119, opened 2021-02-11
Manually updating "jsonld-context" is overriden by "labels"


dr0i:
Originates from #1118:

`web/conf/context.jsonld` is build automatically when executing mvn. It's based on the `context-labels.json`. And the latter wasn't updated since end of 2019.  While `web/conf/context.jsonld` was updated multiple times then. This is not good because changing `web/conf/context.jsonld` will be overriden by `context-labels.json`. Phew.

I think my decision to do this was the dependency of "labels" and "context" and the desire to have only one place to change the definitions. On the other hand I wanted the "context" in it's place because it referenced in every resource  (it's http://lobid.org/resources/context.jsonld). Also, this would show if there would be an issue with the context , even if `context-labels.json` would look ok.
So.
The easiest would be to make `context.jsonld` only readable by the user resp. to add it to gitignore. But then it would be necessary to generate the `context` when building the "play" app.

dr0i:
In principle, yes, this is still a (potential) problem. It affects all ETLs which make use of `context.jsonld`.

dr0i:
I've created a wiki page https://github.com/hbz/lobid-resources/wiki/How-context.jsonld-is-generated.
There will be no deployment problem since we have `&& git restore web/conf/context.jsonld ...` in the cronatb. Closing.

PR #1121, opened 2021-02-15
Add POR-data to heldBy (#1116)

TobiasNx:
This worked out faster than I thought.

### DIFF
--- a/src/main/resources/alma/common/item.xml
+++ b/src/main/resources/alma/common/item.xml
@@ -38,7 +38,7 @@
       <data name="sublocation" source="POR  .d" />
       <entity name="heldBy" sameEntity="true">
         <data name="id" source="POR  .a">
-          <regexp match=".*(\d{4})$" format="{1}"/>
+          <regexp match=".*(\d{4})$" format="${1}"/>
           <lookup in="alma-institution-code-to-isil" />
           <compose prefix="http://lobid.org/organisations/" postfix="#!"/>
         </data>
--- a/src/main/resources/alma/common/item.xml
+++ b/src/main/resources/alma/common/item.xml
@@ -2,9 +2,9 @@
 <rules xmlns="http://www.culturegraph.org/metamorph">
   <entity name="hasItem[]" flushWith="record">
     <entity name="" reset="true" sameEntity="true">
-			<combine name="id" value="https://lobid.org/item/${i}">
-				<data name="i" source="MBD  .i"/>
-			</combine>
+      <combine name="id" value="https://lobid.org/item/${i}">
+        <data name="i" source="MBD  .i"/>
+      </combine>
       <entity name="type[]" sameEntity="true">
         <data name="" source="@isiliz">
           <constant value="Item"/>
@@ -17,6 +17,28 @@
         <data name="label" source="@isiliz">
           <constant value="lobid Organisation"/>
         </data>
+      </entity>
+    </entity>
+    <entity name="" reset="true" sameEntity="true">
+      <combine name="id" value="https://lobid.org/item/${i}">
+        <data name="i" source="POR  .a"/>
+      </combine>
+      <entity name="type[]" sameEntity="true">
+        <data name="" source="POR  .a">
+          <constant value="Item"/>
+        </data>
+      </entity>
+      <entity name="heldBy" sameEntity="true">
+        <data name="id" source="POR  .a">
+          <regexp match=".*(\d{4})$" format="{1}"/>
+          <lookup in="alma-institution-code-to-isil" />
+          <compose prefix="http://lobid.org/organisations/" postfix="#!"/>
+        </data>
+        <data name="label" source="POR  .a">
+          <constant value="lobid Organisation"/>
+        </data>
+        <data name="electronicLocator" source="POR  .D" />
+        <data name="sublocation" source="POR  .d" />
       </entity>
     </entity>
   </entity>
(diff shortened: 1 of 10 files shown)


dr0i:
Please resolve conflicts first.

src/main/resources/alma/common/item.xml
```diff
@@ -19,5 +19,27 @@
...
+      </combine>
+      <entity name="type[]" sameEntity="true">
+        <data name="" source="POR  .a">
+          <constant value="Item"/>
+        </data>
+      </entity>
+      <entity name="heldBy" sameEntity="true">
+        <data name="id" source="POR  .a">
+				  <regexp match=".*(\d{4})$" format="{1}"/>
+          <lookup in="alma-institution-code-to-isil" />
```
dr0i, line 34: formatting seems broken

dr0i, line 34: formatting seems broken

dr0i (changes requested):
The inline comment.

src/main/resources/alma/common/item.xml
TobiasNx: recommited it, should be fixed now

dr0i: +1

src/test/resources/alma/HT012734833.json
```diff
@@ -25,6 +25,78 @@
       "label" : "lobid Organisation"
     },
     "id" : "https://lobid.org/item/991000300959706445"
+  }, {
+    "id" : "https://lobid.org/item/5333897520006444",
+    "type" : [ "Item" ],
+    "heldBy" : {
+      "id" : "http://lobid.org/organisations/DE-632#!",
+      "label" : "lobid Organisation",
+      "electronicLocator" : "https://eu04.alma.exlibrisgroup.com/view/uresolver/49HBZ_NETWORK/openurl?u.ignore_date_coverage=true&portfolio_pid=5333897520006444&Force_direct=true",
```
dr0i, line 34: I think this is the wrong level for this key. It should cling directly to Item, see https://id.loc.gov/ontologies/bibframe.html#c_Item.
Same for "sublocation".

dr0i (changes requested):
See inline comment.
Also, the new introduced properties (aka json-keys) should become part of the jsonld-context. To do the latter see e.g. b119202b09b2105713937e790d37c3a283cb88ed as a template.

src/test/resources/alma/HT012734833.json
```diff
@@ -6,25 +6,108 @@
...
       "id" : "http://lobid.org/organisations/DE-290#!",
       "label" : "lobid Organisation"
     },
     "id" : "https://lobid.org/item/991000300959706445"
+  }, {
+    "id" : "https://lobid.org/item/5333897520006444",
+    "type" : [ "Item" ],
+    "label" : "Portfolio",
+    "heldBy" : {
+      "id" : "http://lobid.org/organisations/DE-632#!",
```
dr0i, line 36: Hm this denotes ExLibris. I think we should denote the library who "holds" it. Otherwise, how could we query e.g. "give me all Portfolios of DE-290" ?

dr0i (changes requested):
See inline.

src/test/resources/alma/HT012734833.json
dr0i: resolved with https://github.com/hbz/lobid-resources/pull/1121/commits/f6e40fa5ff26c5fc14bb4d31a39c41ff79803399.

dr0i:
Read carefully:
>  should become part of the jsonld-context. To do the latter see e.g. b119202 as a template.

where it is noted:
> The jsonld-context is build automatically using the context-labels.json.

So, changing the  context.jsonld will not have any effect - it will be overriden.
See also https://github.com/hbz/lobid-resources/issues/1119.

src/test/resources/alma/HT012734833.json
TobiasNx: To be honest, i do not know what is wrong with the lookup here. There is no map to DE-632 in src/main/resources/alma/maps/alma-institution-code-to-isil.xml . And only a reference of it in line 65 :    ` morphVariables.put("isil", "DE-632"); ` of src/test/java/org/lobid/resources/AlmaMarc21XmlToLobidJsonTest.java.

dr0i: Right! It's the latter making the trouble. It is a fallback, hitting if an isil could not be found in the isil-mapping-xml.
There was an error in the morph: a regex group was incorrectly referenced with `{1}`. Correct is `${1}`. Fixed that with https://github.com/hbz/lobid-resources/pull/1121/commits/a2cbab9675590974cfbf2182ea6971f0a6001af3.

TobiasNx:
You are right. Context should be updated now and context.jsonld as testfile is included in the commit. .

dr0i (approved):

dr0i:
Merged, closed.

web/conf/context.jsonld
```diff
@@ -95,6 +95,9 @@
       "@id" : "http://id.loc.gov/ontologies/bibframe/supplement",
       "@container" : "@set"
     },
+    "sublocation" : {
+      "@id" : "http://id.loc.gov/ontologies/bibframe/electronicLocator"
+    },
```
dr0i, line 100: just saw that this is the wrong URL. Will push myself a fix.

dr0i: Fixed with https://github.com/hbz/lobid-resources/commit/96031178a0fd07947ce838404a9d37de8c1b2aa0.

Issue #1125, opened 2021-02-18
Error in "inCollection"


hagbeck:
We have a problem with records of the collection "ZDB-23-DGG-eba". In lobid-resources there only is

```
'inCollection' | 'id' | "http://lobid.org/organisations/ZDB-23-DGG#!"
```

In the corresponding MAB-XML record it is correct:

```
<datafield tag="078" ind1="e" ind2="1">
<subfield code="a">ZDB-23-DGG-eba</subfield>
</datafield>
```

Example: http://lobid.org/resources/HT020719239.json

dr0i:
Fix in master. Data should be fixed after full reindexing at Sunday. To be checked then.

TobiasNx:
The new regex also needs to be integrated in the AMLA-Morph, since inCollection there was rebuild from the ALEPH-Workflow. Should I do that in this branch? Or open up a new one?

dr0i:
Reuse this issue, as it fits to both databases.
Also I propose to not tag this issue with `almaMigration` but leave it as it is, for the issue is not alma specific.

TobiasNx:
I have locally updated the ALMA-Moprh but I need a new ALMA testfile. Where do i get the ALMA-MARC of http://lobid.org/resources/HT020719239.json

dr0i:
Look at https://github.com/hbz/lobid-resources/issues/1110. It's possible to get the source but you have to know the `almaMmmsId`. Don't know if this is sufficient for you, we would have to open an issue of itself if not.

dr0i:
>  It's possible to get the source but you have to know the almaMmmsId.

Uh, no , we are good: 

https://indexes.devel.digibib.net/export/HT017664407

will give you an link to get the resource by `almaMmsId`. \o/

dr0i:
@hagbeck we fixed it for lobid-resources (aleph),  look e.g. at http://lobid.org/resources/HT020719239.json . Now have to fix alma data accordingly.
Thx for the report !

dr0i:
Fixed for both ETLs: Alma and Aleph.
Closing.

PR #1127, opened 2021-02-19
1125 fix zdb id parsing

TobiasNx:
See #1125

### DIFF
--- a/src/main/resources/morph-hbz01-to-lobid.xml
+++ b/src/main/resources/morph-hbz01-to-lobid.xml
@@ -7512,7 +7512,7 @@
 		<!-- ####################### -->
 		<combine name="@epackageDescriptionLink" value="${a}" reset="true">
 			<data source="078e1.a" name="a">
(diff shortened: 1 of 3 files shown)


src/test/resources/jsonld/HT020719239.json
```diff
@@ -0,0 +1,280 @@
+{
```
dr0i, line 1: actually, we don't need this test file, because HT015894164 already tests the morph.
I propose to get rid of my commit .

dr0i (changes requested):

src/test/resources/jsonld/HT020719239.json
TobiasNx: But if HT015894164 is a needed test file later, it will show this change then additional to the "needed" change is should show. Therefore I would disagree.

TobiasNx: I agree, will do.

dr0i (approved):

src/test/resources/jsonld/HT020719239.json
dr0i: +1

Issue #1126, opened 2021-02-19
Add portfolio fields to index profile


dr0i:
Since https://github.com/hbz/lobid-resources/issues/1116 we have `portfolio` in our index:
[Get all electronic resources from DE-290]( http://alma.lobid.org/resources/search?q=hasItem.heldBy.id%3A%22http%3A%2F%2Flobid.org%2Forganisations%2FDE-290%23!%22&medium=http%3A%2F%2Frdaregistry.info%2FtermList%2FRDAMediaType%2F1003&format=json)

But we cannot query `hasItem.label` nor `hasItem.sublocation` nor `hasItem.electronicLocator` for these fields are not configured in https://github.com/hbz/lobid-resources/blob/master/src/main/resources/index-config.json - and the index profile is set to not dynamically add fields (`"dynamic": false`).
We have to configure these fields explicitly to be queryable.

dr0i:
e72dc51 allows to query like this to get all resources which has portfolio: http://alma.lobid.org/resources/search?q=_exists_%3AhasItem.sublocation&format=json.

dr0i:
Ah, no this is not working: `_exists_` only work if (i.a.) this is _not_ the case:
> The field has "index" : false set in the mapping

Wonder why http://alma.lobid.org/resources/search?q=_exists_%3AhasItem.type&format=json _is_ working because the `hasItem.type` is also set to `ìndex: false`.

Going to set `index: true`. But hurts a bit, because it will inflate the index with data we don't really need (indexing the URLs).
This makes it even more valid to implement another way to distinguish the `portfolio` data as asked in https://github.com/hbz/lobid-resources/issues/1116#issuecomment-781959768.

dr0i:
Discussed offline: for now we use `hasItem.label`. This will not persist in production, it's just a workaround, as we have to further discuss the proper way how to do it.

dr0i:
It works so far: get all  62824 Portfolios from Dortmund: http://alma.lobid.org/resources/search?q=hasItem.heldBy.id%3A%22http%3A%2F%2Flobid.org%2Forganisations%2FDE-290%23!%22+AND+hasItem.label:portfolio&format=json CC @hagbeck
Closed.

PR #1129, opened 2021-02-19
Add "electronicLocator" and "sublocation" to index profile

dr0i:
See #1126.

### DIFF
--- a/src/main/resources/index-config.json
+++ b/src/main/resources/index-config.json
@@ -328,6 +328,14 @@
                   "id" : {
                      "type" : "keyword"
                   },
+                  "electronicLocator" : {
+                     "index" : "false",
+                     "type" : "keyword"
+                  },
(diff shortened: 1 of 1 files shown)


dr0i:
Not ready yet - needs testing at play app level.

dr0i:
As discussed offline:
we are not sure how to indicate the data to be of" type" `Portfolio`. We could indeed use an extra `type` entry (but which one?). Atm we use the `label` field to be indexed to be able to query the data like:
http://alma.lobid.org/resources/search?q=hasItem.label%3APortfolio
Ready to be tested with next fullindex tomorrow.

dr0i:
Works. Reveals 1.89M documents with one or more Portfolio, see http://alma.lobid.org/resources/search?q=hasItem.label:portfolio&format=json.
Deployed, closed.

Issue #1131, opened 2021-02-23
Provide script to automatically update test files


dr0i:
As Alma data is  continuously changing it is desirable to have a script which updates the test data.
So we are in sync with our API (alma.lobid.org).
This popped up in different issues, e.g. https://github.com/hbz/lobid-resources/issues/1113.

dr0i:
Deployed, closed.

PR #1132, opened 2021-02-23
1131 update alma test files

dr0i:
See #1131.

This changes a lot of the data, notable removes a lot of `POR` entries (which makes sense, as they are licensed to be used by anyone and thus subsumed under the `DE-605` institution).
I noted a multiplying of `isbn` entries.
Also many gnd-ids are gone :( - but these are missing in the xml sources, so we cannot to a thing about it.

Other things must be investigated.

These lot of changes makes things complicated - we have to carefully adopt the transformation rules. Otherwise we may loose once functioning data transformations to something untested (and unwanted).


dr0i:
Invoke the script like "bash updateAlmaTestFiles.sh".

src/test/resources/alma/updateAlmaTestFiles.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# set head of xml file
+echo '<?xml version="1.0" encoding="UTF-8"?>
+<collection>' > almaMarcXmlTestFiles
+
+# get all JSON test files; filter names from then
+for hbzId in $(ls *.json|cut -d '/' -f2|cut -d '.' -f1 ); do
+	echo $hbzId
+	# lookup and filter alma url
+	almaXmlUrl=$(curl https://indexes.devel.digibib.net/export/$hbzId |jq .|grep -A2 -B2 '"type": "alma"'|grep url|cut -d '"' -f4)
```
blackwinter, line 16: This only works if `$hbzId` is ambiguous; in this case HTTP 300 is returned with a JSON response listing the possible alternatives.

Unique IDs, on the other hand, redirect to the target record; in addition, there's currently no way to restrict the lookup to Alma resources.

dr0i: You are correct, and the script lacks many good things. But atm it's good and working fine - we only are interested in ambiguous data (i.e. data with an Aleph URL _and_ and Alma URL - because we are interested in the Alma Marc sources).
Script will definitely to be improved in the future, as soon as demand shows up.

dr0i: ... and thx for the lookup API - it's doing very well :)

dr0i:
Merged. @TobiasNx  will identify and open issues resulting from this PR.

Issue #1140, opened 2021-02-26
hbzId matching (ALMA)


TobiasNx:
(DE-605)HT020391499 is missing `<controlfield tag="009">` for hbzId, but has 

```
  <datafield tag="035" ind1=" " ind2=" ">
    <subfield code="a">(DE-605)HT020391499</subfield>
  </datafield>
```

We perhaps should take up this ID as `hbzId` and ignore the controlfield

blackwinter:
`009` will be deleted before Go Live, you must use `035$$a` with prefix `(DE-605)`.

dr0i:
@TobiasNx go on  , use `035$$a` with prefix `(DE-605)` instead of `009` for `hbzId`.

PR #1144, opened 2021-03-01
1140 change hbzId

TobiasNx:
See #1140

### DIFF
--- a/src/test/resources/alma/HT017664407.json
+++ b/src/test/resources/alma/HT017664407.json
@@ -1,6 +1,6 @@
 {
-  "hbzId" : "HT017664407",
   "almaIdMMS" : "990197023370206441",
+  "hbzId" : "HT017664407",
   "zdbId" : "2719150-3",
   "title" : "Aegyptische Nachrichten",
   "hasItem" : [ {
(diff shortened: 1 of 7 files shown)


src/test/resources/alma/HT017664407.json
```diff
@@ -1,6 +1,6 @@
 {
-  "hbzId" : "HT017664407",
   "almaIdMMS" : "990197023370206441",
+  "hbzId" : "HT017664407",
```
dr0i, line 3: As discussed -if nothing of importance changes (here only the order changed, which can be ignored) we should not commit the resource.

dr0i (changes requested):

src/test/resources/alma/HT017664407.json
TobiasNx: but this will be highlighted next time when we use this. also this shows that another marc element is used for the hbzId since it is "035" and not "009" which is used for the mapping.

as discussed off-board we keep these

dr0i (approved):

dr0i:
Merged. Closed.

Issue #1145, opened 2021-03-01
Fix POR heldBy


dr0i:
Originating from https://github.com/hbz/lobid-resources/pull/1142#pullrequestreview-599724336.

From:
https://service-wiki.hbz-nrw.de/pages/viewpage.action?pageId=562135157

Beschreibung | Urpsrung | heldBy
-- | -- | --
from NZ but only for some IZ | exists(POR$$A) | POR$$A
from NZ or IZ | !exists(POR$$A) | POR$$a

Mapping for  POR$$A: https://service-wiki.hbz-nrw.de/display/GOAL/A2A+-+Items+-+IZ-Regeln#A2A-Items-IZ-Regeln-ZuordnungOwner-Sigel-IZ

dr0i:
I will create a metafacture map to map POR$$A to an ISIL.

PR #1146, opened 2021-03-01
Add alma-iz-code-to-isil and precedence POR$$A as Item holder

dr0i:
If POR$$A exists it's the holder of this Item.

- add a new map alma-iz-code-to-isil to map between iz code to isil
- add 'POR  .A' to morph and precedence it over 'POR  .a'

See #1145.

### DIFF
--- a/src/main/resources/alma/common/item.xml
+++ b/src/main/resources/alma/common/item.xml
@@ -37,12 +37,18 @@
       <data name="electronicLocator" source="POR  .D" />
       <data name="sublocation" source="POR  .d" />
       <entity name="heldBy" sameEntity="true">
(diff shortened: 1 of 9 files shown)


TobiasNx (changes requested):
Code looks good. Two minor changes:
Commit message uses precedence as verb, which it is not, perhaps you to change it in and ... "and give precedence to POR A ... over POR a"

The connected ticket is missing some context, which explains why POR A has precedence.

dr0i:
- reworded the commit message
- rebased to master
- added link in #1145 to wiki discussion for context

Force-pushed changes.

TobiasNx (approved):

Issue #1156, opened 2021-03-09
Remove "Nichtsortierzeichen" in title


dr0i:
We don't need the "Nichtsortierzeichen" in the data, also these '<<>>' make trouble if not escaped, e.g. in HTML.
See `title` in e.g. http://alma.lobid.org/resources/search?q=HT019841609&format=json.

dr0i:
To be deployed.

dr0i:
Deployed. Closing.

PR #1166, opened 2021-03-23
Delete not sorting signs in title (#1156)

TobiasNx:
See #1156

### DIFF
--- a/src/main/resources/alma/common/fields.xml
+++ b/src/main/resources/alma/common/fields.xml
@@ -680,7 +680,7 @@
           <constant value="Aufgegangen in" />
         </data>
         <data source="78505.t">
-        <any flushWith="700[013] ">
+          <constant value="Teilweise aufgegangen in" />
         </data>
         <data source="78506.t">
           <constant value="Gesplittet in" />
--- a/src/main/resources/alma/common/fields.xml
+++ b/src/main/resources/alma/common/fields.xml
@@ -365,10 +365,15 @@
     </data>
   </entity>
 	-->
-  <call-macro name="regex-del-punctuation-end" dsource="245??.a" dname="title" />
-  <data name="title" source="245??.p">
+  <call-macro name="regex-del-punctuation-end" dsource="245??.a" dname="@title" />
+  <data name="@title" source="245??.p">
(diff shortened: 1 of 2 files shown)


src/main/resources/alma/common/fields.xml
```diff
@@ -675,7 +680,7 @@
           <constant value="Aufgegangen in" />
         </data>
         <data source="78505.t">
-          <constant value="Teilweise aufgegangen in" />
+        <any flushWith="700[013] ">
         </data>
```
dr0i, line 684: Not sure what this field has to do with this issue? Also formatting.

src/main/resources/alma/common/fields.xml
```diff
@@ -365,11 +365,16 @@
     </data>
   </entity>
 	-->
-  <call-macro name="regex-del-punctuation-end" dsource="245??.a" dname="title" />
-  <data name="title" source="245??.p">
+  <call-macro name="regex-del-punctuation-end" dsource="245??.a" dname="@title" />
+  <data name="@title" source="245??.p">
     <occurrence only="1" />
     <replace pattern="\s?\.?\s?$" with="" />
   </data>
```
dr0i, line 372: Not sure why introduce a variable here. You could also simply add the new "replaces" here.

dr0i (changes requested):

src/main/resources/alma/common/fields.xml
TobiasNx: There are two sources of the title. I wanted to insert the use the replacement only once. 245.p is used if 245 .a does not exist.

src/main/resources/alma/common/fields.xml
TobiasNx: you are right, this can be deleted.

src/main/resources/alma/common/fields.xml
dr0i: Ah, yeah I see.So leave that as it is.

TobiasNx:
Fix the inserted error.

dr0i:
Merged. To be deployed.

Issue #1141, opened 2021-02-26
Multiple persons typed as being the same person


dr0i:
See
https://github.com/hbz/lobid-resources/blob/master/src/test/resources/alma/HT005207972.json#L60.

TobiasNx:
It seems that the data is coming from:

https://github.com/hbz/lobid-resources/blob/0fdbc084f7cb24186c885079d0005ac7478a10ed/src/test/resources/alma/HT005207972.xml#L47

There are a lot of different sources of contributer/creator-information:
1)
https://github.com/hbz/lobid-resources/blob/0fdbc084f7cb24186c885079d0005ac7478a10ed/src/main/resources/alma/common/fields.xml#L267

2)
https://github.com/hbz/lobid-resources/blob/0fdbc084f7cb24186c885079d0005ac7478a10ed/src/main/resources/alma/common/fields.xml#L157

3)
https://github.com/hbz/lobid-resources/blob/0fdbc084f7cb24186c885079d0005ac7478a10ed/src/main/resources/alma/common/fields.xml#L190

Also we have a standalone mapping file:

4)
https://github.com/hbz/lobid-resources/blob/master/src/main/resources/alma/common/personCreator.xml


I think 245 should not be the source for our creator information:
https://www.loc.gov/marc/bibliographic/bd245.html

I also think that we should rearrange our creator information-mapping. Is 4) used?

dr0i:
- `245.c` is "Statement of responsibility, etc.", so in principle this seems good. But it seems a bit unstructured and one doesn't know what to expect when using this field (re: mutiple persons). So let us dump that field, especially as it seems to be covered by other , more structured fields.
- `personCreator.xml` seems indeed not to be used, but its content (all?) seems to be copied into `fields.xml. Don't know - we may use  `personCreator.xml` to separate concerns - yes seems to be better.

TobiasNx:
I will move all contribution-metadata to submorph contribution-xml:
100/110 include the most creator informations while 700/710 additional creator and other contributors.
245c would not be needed then.

Also I intend to introduce following changes:

- [ ] Delete 245.c since it is a subfield of MARC-field 245 Title Statement which does not provide additional information as MARC 100/110 or 700/710. 
- [ ] Combine workflows that map 100 and 700 as is already used for CoproreBodies
- [ ] Add label Family for Indicator 3, if at all used? Should then also be allowed in the Schema
- [ ] Allow more roles
[...]

Issue #1155, opened 2021-03-09
GND URI with $1


dr0i:
See e.g. `contribution.agent.id` in 
http://alma.lobid.org/resources/search?q=HT019841609&format=json.

dr0i:
Deployed to production. Looks good. Closing.

PR #1167, opened 2021-03-23
1141 false contributer

TobiasNx:
See #1141. This is upbranched from #1156 so please merge after #1166


src/test/resources/alma/HT016709661.json
```diff
@@ -129,6 +107,18 @@
...
+    "role" : {
+      "id" : "http://id.loc.gov/vocabulary/relators/ctb",
+      "label" : "Mitwirkende"
+    },
+    "type" : [ "Contribution" ],
+    "agent" : {
+      "label" : "Düsseldorf",
+      "type" : [ "CorporateBody" ],
+      "id" : "https://d-nb.info/gnd/$1"
+    }
```
dr0i, line 120: Assuming this is the issue mentioned in https://github.com/hbz/lobid-resources/issues/1155, right?

dr0i (changes requested):
Hm, as you have branched from another branch (not the master) and this other branch may change you have to incorporate that change also here. This may be complicated, and this is what I meant in today's standup with "better branch form the master and let's solve conflicts later".
Besides this: results look good +1 :)

src/test/resources/alma/HT016709661.json
TobiasNx: yes

TobiasNx:
I have set up a new branch and pr. #1170

PR #1170, opened 2021-03-24
1141 false contributor 

TobiasNx:
So reworked these changes and upbranched from master #1141. This should replace #1167.


TobiasNx:
Since i worked at the contribution info I also fixed #1155 here

dr0i:
I am a bit confused because in this branch there are two issues resolved (re last commit).
However we should go on as we don't have much time.
If we see any further problems we will fix them later.
Merged.

Issue #1158, opened 2021-03-11
Generate individual hasItem`s per POR$$A


dr0i:
Reported by @blackwinter in https://github.com/hbz/lobid-resources/issues/1157#issuecomment-796829848:

Generating individual `hasItem`s per `POR$$A` is not the case yet: [99368090810506441](http://alma.lobid.org/resources/search?q=99368090810506441&format=json), as the processing is "flushed with POR$$w" `(<choose flushWith="POR  .w"` )and thus done only once.

TobiasNx:
I think I found the problem see #1164 .

TobiasNx:
It seems that the feature branch was old and behind master. I opened a new feature branch https://github.com/hbz/lobid-resources/tree/1158-takeAllPorAintoAccount_

I think I have a solution for our problem. I used a variation of: https://swissbib.gitlab.io/metamorph-doku/kochbuch/literale-in-entitaeten-wiederverwenden/

TobiasNx:
I did link to the branch but github did not recognize the "_", so check [https://github.com/hbz/lobid-resources/tree/1158-takeAllPorAintoAccount_](https://github.com/hbz/lobid-resources/tree/1158-takeAllPorAintoAccount_) for my proposed solution

dr0i:
Deployed, see e.g. http://alma.lobid.org/resources/search?q=99368090810506441&format=json, closing.

PR #1181, opened 2021-04-01
1158 take all "POR A" into account 

dr0i:
See #1158.


dr0i:
I think this looks very good!
Merged.
To be deployed.

dr0i:
Deployed.

Issue #1171, opened 2021-03-25
Alma index: incorrect data in hasItem/currentLocation


hagbeck:
990217478660206441 (HT019246898):

* current value: "W0001 / Freihand"
* expected: "BAB / Freihand"

990001412590206441 (HT000161712):

* current value: "W0001 / Freihand"
* expected: "EFB / Freihand"

In the second case the expected value for library is in the second hasItem. It seem the values are displaced.

TobiasNx:
@dr0i I need these as testfile, can you help?

dr0i:
@TobiasNx pushed them into https://github.com/hbz/lobid-resources/tree/1171-incorrectCurrectLocation.

TobiasNx:
Thanks, okay it seems that here are different things playing together:

- ITM is only transformed / flused if all mentioned morph information is matched (ITM Subfields a, c, w, x)
- The collectors within the ITM-mapping hasItem-entity were missing reset attributes.

Because of this e.g. HT019246898 which has multiple ITM, but not all have subfields c, w or x.
Then w (value W0001 of another ) was saved and passed to the next x (of the UDE item).

This can also be a reason for the missing number of items, see: #1168 

I added reset and sameEntity to the internal collectors which would correct the mentioned error of @hagbeck but I still need to flush all ITMs not only those that have subfields c, w or x. So when the source entity is closed.

E.g. HT019246898 has ITM's like this, that are missing subfield c and x:
  ```
<datafield tag="ITM" ind1=" " ind2=" ">
    <subfield code="H">2275458510006447</subfield>
    <subfield code="x">91</subfield>
    <subfield code="f">BOOK</subfield>
    <subfield code="v">91</subfield>
    <subfield code="p">42</subfield>
    <subfield code="X">System</subfield>
    <subfield code="U">2018-03-26 10:59:00</subfield>
    <subfield code="Y">2020-12-03 10:15:16</subfield>
    <subfield code="h">false</subfield>
    <subfield code="n">WTA2690+12</subfield>
    <subfield code="M">49HBZ_WUP</subfield>
    <subfield code="s">1</subfield>
    <subfield code="d">#</subfield>
    <subfield code="V">import</subfield>
    <subfield code="b">W00268176</subfield>
    <subfield code="a">2375458450006447</subfield>
    <subfield code="t">8</subfield>
    <subfield code="R">Maintenance count: 001</subfield>
    <subfield code="S">M2018-1738</subfield>
[...]

PR #1183, opened 2021-04-01
1171 incorrect currect location

### LINKED ISSUE #1168: total items in Alma index for DE-290
Currently we're getting 1.024.975 records for DE-290 from the Alma index. [1]
From the Aleph based index we're getting 1.334.514 records [2] 

This gap seems a little bit to big, to be reasoned on not up-to-date sources.

[1] http://alma.lobid.org/resources/search?owner=http%3A%2F%2Flobid.org%2Forganisations%2FDE-290%23!&format=json
[2] http://lobid.org/resources/search?owner=http%3A%2F%2Flobid.org%2Forganisations%2FDE-290%23%21&aggregations=owner

dr0i:
See #1171.

As this brings up some more items it may also helps to fix  #1168.


dr0i:
Going to deploy to see if it help fixing #1168.

src/main/resources/alma/common/item.xml
```diff
@@ -2,7 +2,7 @@
 <rules xmlns="http://www.culturegraph.org/metamorph">
   <entity name="hasItem[]" flushWith="record">
     <!-- ITM -->
-    <entity name="" reset="true" sameEntity="true">
+    <entity name="" reset="true" sameEntity="true" flushWith="ITM  .w">
```
TobiasNx, line 5: If subfield `"ITM  .w"` always exists within an ITM entry AND Also if  `"ITM  .w"` is always behind `"ITM  .x".  then this would be a good solution since not all ITMs have `"ITM  .c"`. Testdata looks like it.

Otherwise we would need another solution. 

I asked for it here: https://service-wiki.hbz-nrw.de/pages/viewpage.action?pageId=562135157

src/main/resources/alma/common/item.xml
```diff
@@ -25,10 +25,10 @@
         </data>
       </entity>
       <data name="callNumber" source="ITM  .c" />
-      <combine name="currentLocation" value="${a} / ${b}">
+      <combine name="currentLocation" value="${a} / ${b}" sameEntity="true">
         <data name="a" source="ITM  .w"/>
         <data name="b" source="ITM  .x"/>
```
TobiasNx, line 30: Testdata seems to show that "ITM .w" is always behind `"ITM .x". +1

TobiasNx:
@blackwinter has pointed out that this solution might be instable due to de dependencies of order and occurrence. Instead of `flushWith="ITM  .w"` we should use `flushWith="ITM  "`. It is important to use the ITM with two spaces at the end for the indicators. This allows to flush at the end of an entity without dependencies the subfields which is great!!!!

src/main/resources/alma/common/item.xml
TobiasNx: I updated the workflow in a new commit.

dr0i:
Good work @TobiasNx .
Going to deploy. Maybe this way better solution will gain us even more resources for #1168.

dr0i:
Deployed. Amount does not change - in fact decreased even a bit, but that may be of some deletions and we should not analyze this deeper.

dr0i:
Merged, closed.

Issue #1174, opened 2021-03-25
Include HOL without ITM as information for hasItem


TobiasNx:
HOL without ITM are used for multivolume works. We need to include them to in our hasItem information:

e.g.:
HT006855611

This should help to reduce the difference discussed in #1168

TobiasNx:
I tried to filter in multiple ways none seems to work:

- if-Conditional in the entity:
```
      <if>
          <none>
            <data source="ITM  .a">
            </data>
          </none>
      </if>
```

works only without flushWith and without sameEntity="true" at the level of entity, but fills all information in one single object and creates duplicate keys

- variable outside the entity for ITM  .A which is used in if-Conditional in the entity
```
      <if>
          <none>
            <data source="@ITMexists">
            </data>
          </none>
      </if>
```

 same result as above


- if-Conditional outside the entity for creating a variable for HOL  .8 to be used in the entity 

only puts out the last version

- choose-Collector outside the entity (first ITM  .1 then HOL  .8) like this https://swissbib.gitlab.io/metamorph-doku/kochbuch/literale-in-entitaeten-wiederverwenden/ to create a variable for HOL  .8 to be used in the entity

only puts out the last version


it is very difficult or impossible (?) to only put out only HOL if ITM does not exist, because:
- when if-conditionals are referring to other entities it is not compatible with sameEntity="true"
- when if-condition is referring to a varibale is is only used once a later iteration seems not possible.
- we need to find a way to conditionally relate different entites of the source data.

dr0i:
Cannot see the problem:
_if_ there is a HOL *and* a corresponding ITM (both having the same id, see e.g. HT005207972) your morph creates only _one_ item (fine!).
If there is only a HOL but no ITM the item is created nonetheless (e.g. HT006855611). Fine!

So you are all set.
Or have I missed something?

TobiasNx:
Unfortunately HOL's and ITM's always have different IDs:

e.g. HT005207972
HOL$$8 -> HOL-ID:  2**2**1040645**7**0006445
ITM$$a -> ITM-ID:    2**3**1040645**6**0006445

The HOL-ID is also stated in ITM as subfield H but that is not used.

Therefore we have for HT005207972 objects in hasItem since there are 3 ITM and 3 HOL entries in this record. If it would be like you say it would have been only 3 entries.

We could separate between HOL and ITM entries if we would give HOL also a different type. But that awould not reduce the number of entries.

TobiasNx:
[...]

PR #1185, opened 2021-04-07
1174 Include HOL without ITM'S

TobiasNx:
#1174 

Start PR to compare branch to master


TobiasNx:
As I stated in the issue #1174.

> Unfortunately HOL's and ITM's always have different IDs:
> 
> e.g. HT005207972
> HOL$$8 -> HOL-ID: 2**2**1040645**7**0006445
> ITM$$a -> ITM-ID: 2**3**1040645**6**0006445
> 
> The HOL-ID is also stated in ITM as subfield H but that is not used.
> 
> Therefore we have for HT005207972 objects in hasItem since there are 3 ITM and 3 HOL entries in this record. If it would be like you say it would have been only 3 entries.
> 
> We could separate between HOL and ITM entries if we would give HOL also a different type. But that awould not reduce the number of entries.

We need to filter out HOL if ITM exists. But we have three difficulties:
1) The order of the record usually has HOL before ITM but the order is not allways like that.
2) hasItem is a very complex entity, with objects generated by different sources: POR, ITM and additionally HOL.
3) if-conditionals seem not to work cross entities, at least not if we want to separate the objects at the end. This seems to be due to the fact, that either the if conditional can only flush with the record or the order (HOL before ITM) prevents a condition (ITM  .a does not exist) to validate.

TobiasNx (commented):
Only (DE-605)HT006855611has HOL without ITM, also it has multiple HOL's.
I assume that the if-conditional is now not working at all. This is the same problem I faced before. No HOL's are process at all instead of being filtered out. This is also due to the fact that ITM's are usually behind HOL.
Or do I miss the point here?

src/main/resources/alma/common/item.xml
```diff
@@ -32,14 +32,14 @@
...
+			<!--TODO: Introduce filter that allows the entity to have separate objects. sameEntity="true" is the roadblock at the moment but we need it. -->
       <if>
-          <none>
+          <none flushWith="record">
             <data source="ITM  .a">
               <buffer flushWith="HOL .8" />
             </data>
           </none>
-      </if> -->
+      </if>
```
TobiasNx, line 42: Only (DE-605)HT006855611has HOL without ITM, also it has multiple HOL's.
I assume that the if-conditional is now not working at all. This is the same problem I faced before. No HOL's are process at all instead of being filtered out. This is also due to the fact that ITM's are usually behind HOL.
Or do I miss the point here?

TobiasNx:
I changed the if conditional and the attributes of the entity-collector a little bit to show you that the if-conditional can work when the sameEntity="true"-Attribute at the object level is deleted. But that the result is not being separated into different objects.

See: 299a08f6754d5d12879aafd27e9466ccf8d8e534  (https://github.com/hbz/lobid-resources/pull/1185/commits/299a08f6754d5d12879aafd27e9466ccf8d8e534)

TobiasNx:
I think i found a solution PR should be closed and be substituted by #1187

TobiasNx:
Substituted by  #1187

PR #1187, opened 2021-04-08
1174 include ho lwithout itm working

TobiasNx:
I think I have a solution for #1174 .
This PR should replace #1185.


dr0i (approved):
+1 great :)

Issue #1191, opened 2021-04-12
Add DDC (ALMA)


TobiasNx:
DDC Metadata is located in Marc field `082`, which is not repeatable.
See: https://www.loc.gov/marc/bibliographic/bd082.html

Also additional DDC Metadata can be found `083`
See: https://www.loc.gov/marc/bibliographic/bd083.html

dr0i:
Deployed, see e.g. http://alma.lobid.org/resources/99368090810506441.json.

PR #1196, opened 2021-04-13
Add DDC (082) as subject

TobiasNx:
(#1191)


src/main/resources/alma/maps.xml
```diff
@@ -4,7 +4,7 @@
   <xi:include href="maps/alma-institution-code-to-isil.xml" />
   <xi:include href="maps/alma-iz-code-to-isil.xml" />
   <xi:include href="maps/rswk-indicator.xml" />
-
+	<filemap name="deweyLabels" files="src/main/resources/alma/maps/deweyLabels.tsv"/>
   <!--
```
dr0i, line 8: indentation

src/main/resources/alma/maps.xml
```diff
@@ -4,7 +4,7 @@
   <xi:include href="maps/alma-institution-code-to-isil.xml" />
   <xi:include href="maps/alma-iz-code-to-isil.xml" />
   <xi:include href="maps/rswk-indicator.xml" />
-
+	<filemap name="deweyLabels" files="src/main/resources/alma/maps/deweyLabels.tsv"/>
```
dr0i, line 7: don't reference it like this. Also, we only need the test file. Analog to https://github.com/hbz/lobid-resources/blob/master/src/main/resources/morph-hbz01-to-lobid.xml#L7685 , should work.

src/main/resources/alma/maps/deweyLabels.tsv
```diff
@@ -0,0 +1,45 @@
+004.62	Kopplungs- und Kommunikationsprotokolle
```
dr0i, line 1: don't put it into the repo, for when copy the "other file" to this location this other file would be overriden everytime we pull.

dr0i (changes requested):
Added some comments inline.
Besides this - great to say hello to ddc :)

src/main/resources/alma/maps.xml
TobiasNx: @dr0i Thisdoes not work, i tried it like this before. The reason is, that in  https://github.com/hbz/lobid-resources/blob/master/src/main/resources/morph-hbz01-to-lobid.xml#L7685 the deweyLabel file is at the same level als the morph. We need to transfer the  map file at least in the maps folder as the other maps..

src/main/resources/alma/maps/deweyLabels.tsv
TobiasNx: As I said at the moment we do not have any deweyLabels.tsv in the maps folder. Either we use this here or you have to create a new file on the server.

dr0i: " we do not have any deweyLabels.tsv in the maps folder"
but we have - in the test-folder. The files there are also available when calling them from "src".
See add11eeaf56eac75cae0c9c230c8b40b2750a556.

Issue #1194, opened 2021-04-13
Add responsibilityStatement (field 245??.c)


dr0i:
Like in http://lobid.org/resources/HT000068434.json we need `responsibilityStatement` in http://alma.lobid.org/resources/990000597810206441.json.

dr0i:
There is already a test resource reflecting this field: `(DE-605)HT020202475.xml`.

TobiasNx:
Subfield .c is prrovided with every 245 field, which appearse once in a record.

TobiasNx:
@acka47: responsibilityStatement is an array. The source of it is a non repeatable field. Should this be changed to a string value, or is this a too strong api break?

acka47:
>  `responsibilityStatement` is an array. The source of it is a non repeatable field `245??c` (MAB `359# $$a`). Should this be changed to a string value, or is this a too strong api break?

This would be an API break, yes. As discussed in the meeting today, we will leave it like this.

dr0i:
Deployed, see e.g. http://alma.lobid.org/resources/990001412590206441.json and http://alma.lobid.org/resources/990000597810206441.json.

PR #1197, opened 2021-04-15
Add responsibility statement

TobiasNx:
see #1194

### DIFF
--- a/src/main/resources/alma/common/contribution.xml
+++ b/src/main/resources/alma/common/contribution.xml
@@ -17,10 +17,10 @@
         <call-macro name="gndIdentifier" field="100" />
         <call-macro name="gndIdentifierId" field="100" />
         <!--
-      <call-macro name="personTitle" field="100" />
-      <call-macro name="personBio" field="100" />
-      <call-macro name="personRole" field="100" />
-      <call-macro name="gndIdentifier" field="100" />
+      <call-macro name="personTitle"      field="100" />
+      <call-macro name="personBio"        field="100" />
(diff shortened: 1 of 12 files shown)


src/main/resources/alma/common/contribution.xml
```diff
@@ -13,10 +17,10 @@
...
         <call-macro name="gndIdentifierId" field="100" />
         <!--
-      <call-macro name="personTitle"      field="100" />
-      <call-macro name="personBio"        field="100" />
-      <call-macro name="personRole"       field="100" />
-      <call-macro name="gndIdentifier"    field="100" />
+      <call-macro name="personTitle" field="100" />
+      <call-macro name="personBio" field="100" />
+      <call-macro name="personRole" field="100" />
+      <call-macro name="gndIdentifier" field="100" />
```
dr0i, line 23: at least this comment about macro calling is redundant (see line 17). I would suggest to remove all this comments here)

dr0i (approved):
Looks good!
You may remove some comments, see inline.

src/main/resources/alma/common/contribution.xml
TobiasNx: Ok this seems to be a formatting highlighting. Perhapse I should tidy up the outcommented macros in a new branch since they are not connected to the responsibilityStatement

dr0i: +1. So you will remove the formatting highlighting here as it also has nothing to do with the responsibilityStatement?

dr0i (changes requested):

Issue #1242, opened 2021-04-30
wrong values in `hasVersion` and `related`


TobiasNx:
e.g. HT000161712

```JSON
  "hasVersion" : [ {
    "id" : "http://digitale-objekte.hbz-nrw.de/storage2/2019/04/06/file_4/8295425.pdf"
  } ],
  "related" : [ {
    "id" : "Inhaltsverzeichnis"
  } ],
```

TobiasNx:
as discussed off-board woth @dr0i : I split this issue in two steps:
1) I will out-comment `hasVersion` and `related`-transformation. So that I can work on #1237 since they encompass tableOfContent information.
2) After solving #1237 I will rearrange `hasVersion` and `related`

dr0i:
Deployed, see e.g. See e.g. https://alma.lobid.org/resources/search?q=HT000161712&format=json. Closed.

TobiasNx:
Reopen for remodelling `"hasVersion"` and `"related"`

TobiasNx:
@acka47 

I am irritated by the definition of "hasVersion" as lobid is using it: https://github.com/hbz/lobid-resources/pull/453/commits/e26a3189d3c516b99399e4d8c4394e7da690b302

> hasVersion contains andything from 655, subfields x or 3 if not a description, fulltext or tableOfContents. Improve definition of fulltextOnline by adding and deleting.

655 is only refering to electronic links in the metadata.

In contrast https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/hasVersion gives a much broader definition:

> 
>Definition | A related resource that is a version, edition, or adaptation of the described resource.
>-- | --
>Comment | Changes in version imply substantive changes in content rather  than differences in format. This property is intended to be used with  non-literal values. This property is an inverse property of Is Version  Of.

I think this does not fit conceptually. Do we need this very narrow link list at all? Or should we even remodel the hasVersion?

acka47:
The issue where the current status regarding `seeAlso`, `hasVersion` and `fulltextOnline` have been discussed is #441. Maybe this helps for your understanding, @TobiasNx ?

TobiasNx:
The values in 856## $3 which used to be 655 $3 and specifiy the content of the link are:

> 655 #  L 3 Abstract
655 #  L 3 Ausführliche Beschreibung
655 #  L 3 Auszug
655 #  L 3 Autorenbiografie
[...]

PR #1243, opened 2021-04-30
Outcomment `hasVersion` and `related` (#1242)

TobiasNx:
Both transformations were not set up correctly.

See #1242


dr0i:
See e.g. https://alma.lobid.org/resources/search?q=HT000161712&format=json.

Issue #1179, opened 2021-03-30
No data for related records (Alma)


hagbeck:
Currently there are no IDs to related records such as multi volume works.

Example: 990225056670206441 (TT003907920) is part of 990050000600206441 (HT006855611). In Alma this relation seems to be documented via field 773

`773	08 |w (DE-605)HT006855611 |q 3`

and only via hbz-ID and not via MMS-ID. But the representation in the publishing results might be different.

Btw.: 990225056670206441 (TT003907920) is an example without DE-290 as owner

blackwinter:
The planned, but not yet enabled, [related records enrichment](https://service-wiki.hbz-nrw.de/pages/viewpage.action?pageId=525926523) might be helpful here. The existing linking entries (based on the hbz-ID) originate from Aleph CC.

hagbeck:
Today I've learned that we don't get MMS-IDs for related records for the time being. So we have to use the hbz-/Aleph-ID in MARC field 773 subfield w. Subfield q seems to be the volume number.

dr0i:
Moving this up to "ready", since we should care about this because at 26th April we have to deal with it. As I understand the field is not yet in the source MARC but we can treat it in the morph as if it were already present.

blackwinter:
Linking fields - based on the hbz-ID - are already present. This doesn't change with the initial publishing after Go Live.

The question is, AFAIUI, if/when MMS IDs become available for linking.

dr0i:
Deployed to production, see e.g. http://alma.lobid.org/resources/990225056670206441.json.

UBmakla:
no data for related records in 990225980070206441

dr0i:
990225980070206441 seems to have it's `hasPart` information in MARC `830$0.w` - at least there is no `773` in the MARC sources (and that's the field that was treated in  https://github.com/hbz/lobid-resources/pull/1217 ):
```
<datafield tag="830" ind1=" " ind2="0">
<subfield code="a">Kunst & Kohle</subfield>
<subfield code="w">(DE-605)HT019688715</subfield>http://lobid.org/resources/HT019688715
```

So it seems we have to take MARC `830$0.w` into account, too?

TobiasNx:
`830` is refering to _Series Added Entry-Uniform Title_


> Series added entry consisting of a series title alone.
> 
[...]

PR #1244, opened 2021-04-30
Add `830` as isPartOf (#1179)

TobiasNx:
I rearranged the isPartOf mapping a little bit too, to prevent mixed objects.


dr0i:
See e.g. https://alma.lobid.org/resources/search?q=HT017015300&format=json.

Issue #1248, opened 2021-05-03
Remodel contributor roles (ALMA)


TobiasNx:
Concerning our ALMA-Transformation:
At the moment organisations as contributors always have the role "Mitwirkende" but "Autor" (author) should be possible too.

https://github.com/hbz/lobid-resources/blob/2d136786157b1a3f4b6837f980a4a5b33d9cc576/src/main/resources/alma/common/contribution.xml#L113-L120

TobiasNx:
To not loose the discussion in the not used PR: #1255

https://github.com/hbz/lobid-resources/pull/1255#issuecomment-833313750:
> I'll try to answer your questions here:
> ### Are the code always [marc relators](https://id.loc.gov/vocabulary/relators.html)?
> 
> I am quite sure that they are.
> ### How do we handle contributors with more than one role? See HT019246898 We need an array perhaps.
> 
> See e.g. [#933 (comment)](https://github.com/hbz/lobid-resources/issues/933#issuecomment-431770264) that links to more sources. I'll repeat what I wrote in [#868 (comment)](https://github.com/hbz/lobid-resources/issues/868#issuecomment-405514382) here:
> 
> > In 2016 I asked on the Bibframe list how to record multiple contributions by the same person. Ray Denenberg of LoC [responded](https://listserv.loc.gov/cgi-bin/wa?A2=ind1612&L=BIBFRAME&D=0&P=13768):
> > > If you want to list two roles, e.g. “illustrator” and “author” for the same person, the rule is, declare two separate Contribution resources.   Two or more role statements within the same Contribution resource should occur only because you want to provide two different representations (e.g. two different URIs) for the same role.
> 
> ### How do we provide labels, is there a mapping list?
> 
> The labels for MARC relators and many other controlled terms are configured in the [labels.json](https://github.com/hbz/lobid-resources/blob/master/src/main/resources/labels/labels.json).



and
https://github.com/hbz/lobid-resources/pull/1255#issuecomment-833557636
> Concerning the multiple roles that seems to be difficult to me to model, since the roles appear in a single field 100/110/700/710 as multiple subfields:
> e.g. HT019246898
> 
> ```
>   <datafield tag="700" ind1="1" ind2=" ">
>     <subfield code="a">Abel, Alexandra</subfield>
>     <subfield code="0">(DE-588)1069204668</subfield>
>     <subfield code="4">edt</subfield>
>     <subfield code="4">aut</subfield>
[...]

Issue #933, opened 2018-10-17
Duplicate corporate body in contribution list 


jschnasse:
Where is the second occurrence of http://d-nb.info/gnd/2116834-9 coming from? As reported by client it does occur only once in MAB source. 

https://lobid.org/resources/HT019838800
http://lobid.org/hbz01/HT019838800

https://lobid.org/resources/HT019837193
http://lobid.org/hbz01/HT019837193

https://lobid.org/resources/HT019826502
http://lobid.org/hbz01/HT019826502

https://lobid.org/resources/HT019826031
http://lobid.org/hbz01/HT019826031

https://lobid.org/resources/HT019825784
http://lobid.org/hbz01/HT019825784

dr0i:
These have in common: a double entry of `200b1.4.`According to our morph this will then fires two times. Fix is to fire when the entity (here: `200`) hits the `?????.9` (the gnd ID) which occurs only once.

dr0i:
Deployed the mentioned resources into staging, see e.g. https://stage.lobid.org/resources/HT019826031.

jschnasse:
+1 this looks much more logical - hm - until we hit a title that defines a contributor type for contributor without GND-ID

dr0i:
Indeed.

dr0i:
Going to deploy this. Needs complete reindexing for all resources => this will be done after complete ETL and after the manual index switch during the course of monday.

dr0i:
As discussed offline with @acka47 the former behaviour was indeed correct: it is a contributor with two roles (also these roles were matched to the default ("Contributor"). In lobid we decided to denote explicitly one contributor with one role. So in this case it would be two times the same contributor, only differing in role.
However, in the frontend we just show under "Mitwirkende" the name of the contributor so that the same name appears two times. If that feels weired, it is up to the rendering to filter to show only one.

acka47:
Generally, it is expected behaviour that we have two `Contribution` objects with the same entity if subfield `4` for a contributor (1--s oder 2--s) is repeated. See #868 for background and especially https://github.com/hbz/lobid-resources/issues/868#issuecomment-405514382 with a link to an email in the Bibframe list where Ray Denenberg says that this is how it should be done (instead of having one object with both roles listed in an array).

[...]

PR #881, opened 2018-07-25
868 multiple roles for agents

ChristophEwertowski:
Resolves #868

Issue #868, opened 2018-07-05
Include multiple roles for agents


ChristophEwertowski:
See MAB 100 at [HT018857620 (MAB-XML)](http://lobid.org/hbz01/HT018857620). In the json only one role is covered: http://lobid.org/resources/HT018857620.json. Because it's the first time that I saw something like this I got the confirmation by a cataloger that it's possible.

ChristophEwertowski:
Since `role` isn't an array these changes will be an API break. So we have to inform all users before deploying this.

acka47:
In 2016 I asked on the Bibframe list how to record multiple contributions by the same person. Ray Denenberg of LoC [responded](https://listserv.loc.gov/cgi-bin/wa?A2=ind1612&L=BIBFRAME&D=0&P=13768):

> If you want to list two roles, e.g. “illustrator” and “author” for the same person, the rule is, declare two separate Contribution resources.   Two or more role statements within the same Contribution resource should occur only because you want to provide two different representations (e.g. two different URIs) for the same role.

ChristophEwertowski:
Note: Apparently you can `<combine>` like the following in the morph. `Field A + field B` / `field A + field C` etc.

ChristophEwertowski:
The solution which with I came up with can cover up to three roles of an MAB number / entity. To test this I added roles to test resources see `Barenboim, Daniel` at [HT018857620 (production, json)](http://lobid.org/resources/HT018857620.json) / [HT018857620 (test, json)](http://test.lobid.org/resources/HT018857620.json) and `Albert-Ludwigs-Universität Freiburg` at [HT018770176 (test, json)](http://test.lobid.org/resources/HT018770176.json).
I will open an issue at the metamorph repository for using loops. With that it would be possible to use all roles if there are four or more and it would reduce the lines of code dramatically.

acka47:
+1

PR #1255, opened 2021-05-05
1248 contributor roles

TobiasNx:
See #1248
Until now we have only cre (label: Autor/-in), oth (Mitwirkende) But there are many more marc relators which seem to be used.

Questions:
- Are the code always [marc relators](https://id.loc.gov/vocabulary/relators.html)?
- How do we handle contributors with more than one role? See HT019246898 We need an array perhaps.
- How do we provide labels, is there a mapping list?

### DIFF
--- a/src/main/resources/alma/common/contribution.xml
+++ b/src/main/resources/alma/common/contribution.xml
@@ -23,12 +23,10 @@
           </data>
         </entity>
       </entity>
-      <entity name="role" reset="true">
-        <data name="id" source="100[01] .[4e]">
-          <regexp match="^([Aa]ut|[Cc]re)" format="http://id.loc.gov/vocabulary/relators/cre"/>
-        </data>
-        <data name="label" source="100[01] .[4e]">
-          <regexp match="^([Aa]ut|[Cc]re)" format="Autor/in"/>
+      <entity name="role" reset="true" sameEntity="true" flushWith="100[01] ">
+        <data name="label" source="100[01] .e" />
+        <data name="id" source="100[01] .4">
+          <compose prefix="http://id.loc.gov/vocabulary/relators/"/>
         </data>
       </entity>
     </entity>
@@ -49,33 +47,14 @@
           </data>
         </entity>
       </entity>
-      <entity name="role" reset="true">
-        <if>
-          <data name="id" source="700[01] .[4e]">
-            <regexp match="^([Aa]ut|[Cc]re)"/>
-          </data>
-        </if>
-        <data name="id" source="700[01] .[4e]">
-          <regexp match="^([Aa]ut|[Cc]re)" format="http://id.loc.gov/vocabulary/relators/cre"/>
-        </data>
-        <data name="label" source="700[01] .[4e]">
-          <regexp match="^([Aa]ut|[Cc]re)" format="Autor/in"/>
-        </data>
-      </entity>
-      <entity name="role" reset="true">
-        <if>
-          <data name="id" source="700[01] .[4e]">
-            <regexp match="^oth"/>
-          </data>
-        </if>
-        <data name="id" source="700[01] .[4e]">
-          <regexp match="^oth" format="http://id.loc.gov/vocabulary/relators/oth"/>
-        </data>
-        <data name="label" source="700[01] .[4e]">
-          <regexp match="^oth" format="Sonstiges"/>
+      <entity name="role" reset="true" sameEntity="true" flushWith="700[01] ">
+        <data name="label" source="700[01] .e" />
+        <data name="id" source="700[01] .4">
+          <compose prefix="http://id.loc.gov/vocabulary/relators/"/>
         </data>
       </entity>
     </entity>
+
     <!-- 110|710[012] -->
     <entity name="" flushWith="[17]10[012] " sameEntity="true">
       <entity name="type[]" reset="true">
@@ -114,6 +93,7 @@
       </entity>
     </entity>
 
+    <!-- 111 -->
     <entity name="Conference" flushWith="record">
       <data name="conferenceName" source="111[012] .a">
         <replace pattern="[.,]$" with="" />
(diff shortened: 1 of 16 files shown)


acka47:
I'll try to answer your questions here:

### Are the code always [marc relators](https://id.loc.gov/vocabulary/relators.html)?

I am quite sure that they are.

### How do we handle contributors with more than one role? See HT019246898 We need an array perhaps.

See e.g. https://github.com/hbz/lobid-resources/issues/933#issuecomment-431770264 that links to more sources. I'll repeat what I wrote in https://github.com/hbz/lobid-resources/issues/868#issuecomment-405514382 here:

> In 2016 I asked on the Bibframe list how to record multiple contributions by the same person. Ray Denenberg of LoC [responded](https://listserv.loc.gov/cgi-bin/wa?A2=ind1612&L=BIBFRAME&D=0&P=13768):
>
>> If you want to list two roles, e.g. “illustrator” and “author” for the same person, the rule is, declare two separate Contribution resources.   Two or more role statements within the same Contribution resource should occur only because you want to provide two different representations (e.g. two different URIs) for the same role. 

### How do we provide labels, is there a mapping list?

The labels for MARC relators and many other controlled terms are configured in the [labels.json](https://github.com/hbz/lobid-resources/blob/master/src/main/resources/labels/labels.json).

dr0i:
As @acka47 notes we need to use the `labels.json`. The file is used by `JsonLdEtikett` which we have to configure as receiver after `JsonEncoder`. The `JsonLdEtikett` will ensure that every object with an `id` has a label based on the definitions in `labels.json`. If no explicit definition is found a fallback is used to generateautomatically  a more-or-less fitting label. Main idea is to _always_ have provide a label.

TobiasNx:
> As @acka47 notes we need to use the `labels.json`. The file is used by `JsonLdEtikett` which we have to configure as receiver after `JsonEncoder`. The `JsonLdEtikett` will ensure that every object with an `id` has a label based on the definitions in `labels.json`. If no explicit definition is found a fallback is used to generate automatically a more-or-less fitting label. Main idea is to _always_ provide a label.

But this is a "global" issue for the ALMA morph right? If we use the labels.json we would need to more than just the creator rol labels, right?

TobiasNx:
Concerning the multiple roles that seems to be difficult to me to model, since the roles appear in a single field 100/110/700/710 as multiple subfields:
e.g. HT019246898
```
  <datafield tag="700" ind1="1" ind2=" ">
    <subfield code="a">Abel, Alexandra</subfield>
    <subfield code="0">(DE-588)1069204668</subfield>
    <subfield code="4">edt</subfield>
    <subfield code="4">aut</subfield>
    <subfield code="0">(uri) https://portal.dnb.de/opac.htm?method=simpleSearch&amp;cqlMode=true&amp;query=idn=1069204668</subfield>
    <subfield code="0">(uri) http://viaf.org/viaf/sourceID/DNB|1069204668</subfield>
    <subfield code="B">GND-1069204668</subfield>
  </datafield>
  <datafield tag="700" ind1="1" ind2=" ">
    <subfield code="a">Rudolf, Bernd</subfield>
    <subfield code="0">(DE-588)1038489679</subfield>
    <subfield code="4">edt</subfield>
    <subfield code="4">aut</subfield>
    <subfield code="0">(uri) https://portal.dnb.de/opac.htm?method=simpleSearch&amp;cqlMode=true&amp;query=idn=1038489679</subfield>
    <subfield code="0">(uri) http://viaf.org/viaf/sourceID/DNB|1038489679</subfield>
    <subfield code="B">GND-1038489679</subfield>
  </datafield>
  ```

**Comment by dr0i** [comment]
Don't knowexactly what you mean with "global". It's just that EtikettMaker would ensure to have a `label` for every entity having an `id``(this was also coined as the "sidecar approach" because every id would have a label). _If_ there is already a label for an id, created in the morph, this morphed label has precedence.
So, we only _gain_ with EtikettMaker.
And yes, if there is an id like a marc relator which is _not_ defined in labels.json and has _not_ been giben a label through the morph there is only a fallbal label created which we may not like very much. We will see this when implemented in the resulting json.

**Comment by acka47** [comment]
> Concerning the multiple roles that seems to be difficult to me to model, since the roles appear in a single field 100/110/700/710 as multiple subfields:

Isn't this analogous to the Aleph XML data that we already process for lobid-resources where also different relator codes are repeated in subfield `4`? See e.g. [HT018857620](https://lobid.org/hbz01/HT018857620) (snippet):

```xml
  <datafield tag="100" ind1="b" ind2="1">
  <subfield code="p">Barenboim, Daniel</subfield>
  <subfield code="d">1942-</subfield>
  <subfield code="4">cnd</subfield>
  <subfield code="3">Dirigent</subfield>
  <subfield code="4">drt</subfield>
  <subfield code="3">Regisseur</subfield>
  <subfield code="9">(DE-588)118506560</subfield>
</datafield>
```

Maybe a look into the hbz01-to-lobid morph makes sense. https://github.com/hbz/lobid-resources/pull/881 is the PR in which the transformation of multiple roles for one agent was implemented. Here is the diff in the morph: https://github.com/hbz/lobid-resources/pull/881/files#diff-c79e89b206237040d700b790040652e320068ef3301084a46eeee16cb715d952

TobiasNx:
> Don't knowexactly what you mean with "global". It's just that EtikettMaker would ensure to have a `label` for every entity having an `id``(this was also coined as the "sidecar approach" because every id would have a label). _If_ there is already a label for an id, created in the morph, this morphed label has precedence.

I wanted to say, that since the EtikettMaker is not used at all at the moment, we provide labels always through the morph. To indtroduce the EtikettMaker as it worked before with Aleph we would check the morph if we need to delete the label provided by the morph. So that would be an issue for itself to introduce the EtikettMaker and update the workflow.

dr0i:
With https://github.com/hbz/lobid-resources/pull/1263 the EtikettMaker is part of the flow.
With it comes the automatic creation of the `web/conf/context.jsonld` (running the test as usually is enough) - which is a nice thing, since it also makes it clear that this `context.jsonld` depends on the `labels`.

TobiasNx:
this PR is replaced by #1266

PR #1256, opened 2021-05-06
Update related-element (#1242)

TobiasNx:
I remodeld the `"related"`- element. Some are missing references to ressource-nr as ISBN or hbz-id e.g. HT015011399.

I also added a note-element to include the description of the resource.

### DIFF
--- a/src/test/resources/alma/(DE-605)HT015011399.json
+++ b/src/test/resources/alma/(DE-605)HT015011399.json
@@ -51,6 +51,10 @@
     "label" : "Culturegraph Ressource"
   } ],
   "thesisInformation" : [ "Dortmund, Univ., Diss., 2007" ],
+  "related" : [ {
+    "label" : "Strukturelle und biochemische Einblicke in Mechanismen des Nukleotidaustauschs von Rab-Proteinen",
+    "note" : "Erscheint auch als Druckausg. u.d.T."
+  } ],
   "contribution" : [ {
     "role" : {
       "id" : "http://id.loc.gov/vocabulary/relators/oth",
@@ -80,7 +84,7 @@
     },
     "resultOf" : {
       "type" : [ "CreateAction" ],
-      "endTime" : "2021-04-30T10:28:49",
+      "endTime" : "2021-05-06T15:26:26",
       "instrument" : {
         "id" : "https://github.com/hbz/lobid-resources",
         "type" : [ "SoftwareApplication" ],
--- a/src/test/resources/alma/(DE-605)HT017015300.json
+++ b/src/test/resources/alma/(DE-605)HT017015300.json
@@ -181,6 +181,10 @@
     "id" : "http://digitale-objekte.hbz-nrw.de/storage/2012/06/16/file_46/4507767.pdf",
     "label" : "Inhaltsverzeichnis"
   } ],
+  "related" : [ {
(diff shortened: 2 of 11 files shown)


src/test/resources/alma/(DE-605)HT017411546.json
```diff
@@ -81,6 +81,11 @@
   } ],
   "extent" : "Online-Ressource",
   "note" : [ "Gesehen am 08.10.12", "Beteil. Körp. anfangs: Historische Kommission für Westfalen in Verbindung mit dem Verein für Geschichte und Altertumskunde Westfalens" ],
+  "related" : [ {
+    "label" : "Westfälische Bibliographie zur Geschichte, Landeskunde und Volkskunde",
+    "id" : "http://lobid.org/resources/(DE-605)HT006951721#!",
+    "note" : "Druckausg."
```
dr0i, line 87: this id doesn't exist. If you want to reference the Aleph-based lobid API, which is ok, remove the (DE-605) first.

src/test/resources/alma/(DE-605)HT017015300.json
```diff
@@ -181,6 +181,10 @@
     "id" : "http://digitale-objekte.hbz-nrw.de/storage/2012/06/16/file_46/4507767.pdf",
     "label" : "Inhaltsverzeichnis"
   } ],
+  "related" : [ {
+    "isbn" : [ "384033490X", "9783840334900" ],
+    "note" : "Erscheint auch als Online-Ausgabe"
+  } ],
```
dr0i, line 187: up to today we have only configured related.id and related.label. (see index-config.json). We might want to use more fields. Note that both fields won't be queryable, which might be ok.

src/test/resources/alma/(DE-605)HT015011399.json
```diff
@@ -51,6 +51,10 @@
     "label" : "Culturegraph Ressource"
   } ],
   "thesisInformation" : [ "Dortmund, Univ., Diss., 2007" ],
+  "related" : [ {
+    "label" : "Strukturelle und biochemische Einblicke in Mechanismen des Nukleotidaustauschs von Rab-Proteinen",
+    "note" : "Erscheint auch als Druckausg. u.d.T."
```
dr0i, line 56: Is the subfield "t" always a reiteration of the main resources title? It does not seem very senseful then and we might better dump that for it's just bloating and obfuscating the metadata IMO.

dr0i (changes requested):

src/test/resources/alma/(DE-605)HT017411546.json
TobiasNx: you are right, it seems, that I did not properly update the data

src/test/resources/alma/(DE-605)HT017015300.json
TobiasNx: You are right I added isbn and note since not all linked ressources have a hbz ID. In this way we can at least state if they have.

src/test/resources/alma/(DE-605)HT015011399.json
TobiasNx: No $t states the title of the refrenced ressource. e.g. if it is a translation it states the name of the original work.

TobiasNx:
This PR is replaced by #1262

PR #1259, opened 2021-05-07
Change label to title, use Lobid-URIs (#1179)

### LINKED ISSUE #1179: No data for related records (Alma)

TobiasNx:
Resolve #1179 

The testdata shows that titles are not always provided, see e.g. TT003907920. isPartOf is build here from `"773"`.

### DIFF
--- a/src/main/resources/alma/common/fields.xml
+++ b/src/main/resources/alma/common/fields.xml
@@ -282,13 +282,11 @@
         <constant value="IsPartOfRelation" />
       </data>
       <entity name="hasSuperordinate[]" sameEntity="true" reset="true">
-        <entity name="" sameEntity="true" reset="true">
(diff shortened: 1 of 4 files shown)


dr0i (approved):
+1 ... and xtra thx for the super fast fix!

dr0i:
Deployed, see e.g. https://alma.lobid.org/resources/search?q=HT017015300&format=json.

Issue #1246, opened 2021-04-30
110 & 710 are missing GND-identifier-fields


TobiasNx:

TobiasNx:
While the GND as`"id"` is created, the field gndIdentifier is not created for 110 an 710. But there seems to be a macro for that we could use.

TobiasNx:
https://github.com/hbz/lobid-resources/blob/2d136786157b1a3f4b6837f980a4a5b33d9cc576/src/main/resources/alma/common/contribution.xml#L17

https://github.com/hbz/lobid-resources/blob/2d136786157b1a3f4b6837f980a4a5b33d9cc576/src/main/resources/alma/common/contribution.xml#L85-L121

dr0i:
Merged but not deployed yet. Will be deployed at Monday.

dr0i:
Deployed, see e.g. http://alma.lobid.org/resources/search?q=HT017411546&format=json .
Note, though, that you cannot query `contribution.agent.gndIdentifier`: http://alma.lobid.org/resources/search?q=contribution.agent.gndIdentifier:*&format=json

PR #1265, opened 2021-05-20
Add gndIdentifier for Organizatio #1246

TobiasNx:
Also changed the workflow so that the gnd makro is used with organizations (110 /710) too


dr0i:
Merged but not deployed yet.

dr0i:
Deployed see e.g. http://alma.lobid.org/resources/search?q=HT016709661&format=json.

PR #1266, opened 2021-05-20
1248 contributor roles new

TobiasNx:
I changed the workflows for the contributor roles and separated the marcRel from the labels. It seems like it works for me. But there needs to be another step to add marcRel Labels that are missing.


dr0i:
Query Testindex: `http://test.alma.lobid.org/resources/search?q=contribution.role.label:"aut" %20AND%20NOT%20type:"Book" &format=json`

src/main/resources/labels/marcRel.json
```diff
@@ -737,9 +777,9 @@
 	},
 	{
 		"uri": "http://id.loc.gov/vocabulary/relators/pat",
-		"label": "Auftraggeber/in",
+		"label": "Im Auftrag von",
 		"multilangLabel": {
-			"de": "Auftraggeber/in",
+			"de": "Im Auftrag von",
 			"en": "commissioning body"
```
dr0i, line 783: You begin German with uppercase but English is lower cased. This should be harmonized.

TobiasNx: i follow the DNB document here: file:///tmp/mozilla_tobias0/AH-017.pdf

Also I think stand alone nones in english are written in lower case. e.g. in dictionaries

TobiasNx: on the other hand: https://www.loc.gov/marc/relators/relacode.html here it is in upper case

dr0i: you got a point, I don't know. You may leave it as it is, then.
Wait:
The labels are supposed to be used as a fallback to name a fields in a catalog. Looking at amazon.com and others it seems to be rather usual to use uppercase to do this.

dr0i:
Conflicts originates from uncommited  aleph-lobid data: the lables of contributor.role.label changed from Autor/in to Schöpfer/in when using marcrel/cre. We are not decided yet if this is some kind of API break. Examing web logs reveals that in May 150 queries were using this field directly.
Going to build an aleph-lobid fullindex using branch WIP_ALEPH_MarcRelAut and an alma-lobid index using the same branch so that we have more numbers to decide at Monday with @acka47 if we change the alpeh-lobid API in that respect  (and re: default marcRel:cre vs marcRel:aut) or not.

dr0i:
I don't know how to go on with this. Maybe @TobiasNx has a clue - assigning.

TobiasNx:
substituted by #1293

PR #1267, opened 2021-05-21
Index related.[isbn|issn]

dr0i:
This makes these fields queryable, like:
http://alma.lobid.org/resources/search?q=related.issn:%220955-8810%22&format=json

See #1242.


dr0i:
Merged, will be deployed after weekend. Nonetheless please have a look at it @TobiasNx to see what is to be done to make fields queryable.

Issue #1268, opened 2021-05-21
Add format subject from 689 to natureOfContent


TobiasNx:
@hagbeck wished to add the format subjects from the Schlagwortfolgen `"689"` those have 

These have type specifications in subfield A and not in subfield D.

e.g.:

```
  <datafield tag="689" ind1="0" ind2="2">
    <subfield code="a">Online-Ressource</subfield>
    <subfield code="A">f</subfield>
  </datafield>
  <datafield tag="689" ind1="0" ind2="3">
    <subfield code="a">Zeitschrift</subfield>
    <subfield code="A">f</subfield>
  </datafield>
```

Subfield "A" refers to values-list:
f - Formschlagwort

dr0i:
Merged, will be deployed after weekend.

dr0i:
Note that the field can occur multiple times but should be unique. Opened issue #1271 for that.

PR #1269, opened 2021-05-21
Add format subject from 689 to natureOfContent (#1268)

TobiasNx:
@hagbeck asked for those specific format subjects unfortunately they do not have gndIdentifiers. Therefore we only have labels.


dr0i:
Merged, will be deployed after weekend.

dr0i:
Deployed, se e.g. http://alma.lobid.org/resources/search?q=HT000312236&format=json.
Note that the field can also be queried: http://alma.lobid.org/resources/search?q=natureOfContent.label:%22Zeitschrift%22&format=json.
Note that the field can occur multiple times but should be unique. Opened issue #1271 for that.

PR #1284, opened 2021-05-31
Update type.json

TobiasNx:
music was missing but was represented in the morph.

### DIFF
--- a/src/test/resources/schemas/type.json
+++ b/src/test/resources/schemas/type.json
@@ -39,7 +39,8 @@
                 "Image",
                 "Map",
                 "Standard",
-                "SubjectHeading"
+                "SubjectHeading",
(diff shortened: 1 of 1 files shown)


acka47:
It isn't used driectly on any lobid resource, is it? See https://lobid.org/resources/search?q=type%3AMusic. Same for `SubjectHeading`, see https://lobid.org/resources/search?q=type%3ASubjectHeading.

TobiasNx:
you are right, music is represented as PublishedScore. Should I delete SubjectHeading?

acka47:
> Should I delete SubjectHeading?

+1

TobiasNx:
Close, deletion of subjectHeading here: #1294

Issue #1271, opened 2021-05-25
natureOfContent: same label multiple times


dr0i:
Originates from #1269, see e.g. http://alma.lobid.org/resources/search?q=99370682219806441&format=json :
```
      "natureOfContent": [
        {
          "id": "http://d-nb.info/gnd/4067488-5",
          "label": "Zeitschrift"
        },
        {
          "label": "Zeitschrift"
        },
        {
          "label": "Zeitschrift"
        },
        {
          "label": "Zeitschrift"
        }
      ],
```

Should be unique.

dr0i:
Deployed, see e.g. https://alma.lobid.org/resources/search?q=99370682219806441&format=json.
Closed.

PR #1285, opened 2021-05-31
1272 gnd as subject source

TobiasNx:
I added GND as possible subject concept source and also uniqued the subject.


src/test/resources/alma/(CKB)5280000000199164.json
```diff
@@ -0,0 +1,204 @@
...
+      "id" : "http://creativecommons.org/publicdomain/zero/1.0",
+      "label" : "Creative Commons-Lizenz CC0 1.0 Universal"
+    } ]
+  },
+  "natureOfContent" : [ {
+    "id" : "http://d-nb.info/gnd/4067488-5",
+    "label" : "Zeitschrift"
+  }, {
+    "label" : "Zeitschrift"
+  }, {
```
dr0i, line 141: Here we have duplications. I thought we have had already fixed that - or is this a new issue?

dr0i (changes requested):

src/test/resources/alma/(CKB)5280000000199164.json
TobiasNx: It is a different issue. This is part of "natureOfContent". But this will be a tricky I think this is related to 
#1268 since the source data comes from 655 and 689 I don't know a easy way to unique the results. But please open another ticket.

dr0i: Ah, you are right, and I opened already https://github.com/hbz/lobid-resources/issues/1271 for it.

dr0i (approved):

PR #1288, opened 2021-06-01
Wip creator roles as mapped in alma

TobiasNx:
This draft PR is connected to #1248:

Since hardcoding will be a little bit difficult, here is a quick solution that would adapt the mapping rule of ALEPH to ALMA of the hbz:

https://service-wiki.hbz-nrw.de/pages/viewpage.action?pageId=483000321

> 
> 1. Geistige Schöpfer (Person)
> 
>     $$4 wenn nicht vorhanden per Default mit "aut" belegt
> 

It then would reproduce the data  and the convention of the Verbundskatalog. It also would not really change the aut labels of all the default entries. This would then not trigger big changes in the GUI of lobid or nwbib since most cre which is was default and was mapped from aut is also labeled Autor/in.

Though it would change the id of the default marcRel to aut. But since the labels and not the id's seem to be queried i think that this would not be such a problem.


TobiasNx:
Additionally good to know:
- this PR takes up all existing creator roles
- maps Author, Verfasser, auth. and marcRel aut to marcRel aut
- changes some minor labels
- sets as explained above the default marcRel of 100|110 to aut analog to the convention of the ALMA-Transformation

src/test/resources/alma/(DE-605)HT000312236.json
```diff
@@ -180,7 +180,7 @@
     },
     "resultOf" : {
       "type" : [ "CreateAction" ],
-      "endTime" : "2021-05-21T11:47:53",
+      "endTime" : "2021-06-01T16:48:11",
```
dr0i, line 183: we agreed to not commit test files which only changes in "endTime"

dr0i (changes requested):
Also try to resolve conflicts.

src/test/resources/alma/(DE-605)HT000312236.json
TobiasNx: I know since this is a Work in Progress and a PR-draft this was not about commit a final result but illustrating the the broader scope of the change.
Adrian and I decided on Wednesday that we can go on from here an adapte the default change from the ALMA transformation.
I will set up a new PR with clean changes.

TobiasNx:
substituted by #1293

PR #1293, opened 2021-06-11
1248 remodel contributor roles

TobiasNx:
- this PR takes up all existing creator roles as given in the source data
- it introduces an separate labels file for marcRel
- maps Author, Verfasser, auth. and marcRel aut to marcRel aut (ALEPH)
- changes some minor labels
- sets as explained below the default marcRel of 100|110 to aut analog to the convention of the ALMA-Transformation (ALEPH)

Concerning the default value aut:

> Here is a quick solution that would adapt the mapping rule of ALEPH to ALMA of the hbz:
> 
> https://service-wiki.hbz-nrw.de/pages/viewpage.action?pageId=483000321
> 
>         Geistige Schöpfer (Person)
> 
>         $$4 wenn nicht vorhanden per Default mit "aut" belegt
> 
> It then would reproduce the data and the convention of the Verbundskatalog. It also would not really change the aut labels of all the default entries. This would then not trigger big changes in the GUI of lobid or nwbib since most cre which is was default and was mapped from aut is also labeled Autor/in.
> 
> Though it would change the id of the default marcRel to aut. But since the labels and not the id's seem to be queried i think that this would not be such a problem.


dr0i:
There are conflicts, please resolve them first.

acka47 (approved):
I pushed some adjustments with b3df387. (Feel free to revert those that are not correct.) Otherwise +1.

TobiasNx:
There should be no conflicts here anymore

dr0i:
There was a failing Aleph API test - updated that.

dr0i:
See e.g. https://alma.lobid.org/resources/search?q=HT000161712&format=json.
Closed.

PR #1296, opened 2021-06-11
Add publication>startDate and endDate (#1279)

### LINKED ISSUE #1279: Read publication dates with regex and not just as publicationHistory
Since there was no coherent data for publication start and end dates we moved it to publicationHistory, but I think we should be able to read the start and end dates with regexp

TobiasNx:
resolves #1279


dr0i:
+1

dr0i:
Deployed, see e.g. http://alma.lobid.org/resources/search?q=99370682219806441&format=json.

Issue #1193, opened 2021-04-13
Correct Alma resources id


dr0i:
Atm Alma resources have their old Aleph lobid-resources id, e.g.:

> id": "http://lobid.org/resources/HT015671602#!",

(see http://alma.lobid.org/resources/990171142550206441.json).

The proper `ìd` would be `https://alma.lobid.org/resources/990171142550206441#!`.

dr0i:
Hm, `curl -Lvvv --insecure  "accept: application/json" 'https://alma.lobid.org/resources/990171142550#!'` does not work, while the one using `http` does. Will check again when SSL certificate is configured (see #1159).

dr0i:
Deployed, see e.g. https://alma.lobid.org/resources/99370682219806441.json.
Closing.

PR #1301, opened 2021-06-17
Build id with almaMmsId (#1193)

TobiasNx:
Change all connected processes and delete id links that are not needed.


dr0i:
Deployed, see e.g. https://alma.lobid.org/resources/99370682219806441.json.

PR #1306, opened 2021-07-26
1305 non official collections

TobiasNx:
I added 962  ## $e if it does not start with ZDB and 960 1#   as collections. Not all inofficial collections have labels. Could you have a look, @hagbeck and @dr0i :

Concerning 962:
- I tried to combine the `regexp` for 962 but that did not work the results seem to be cut short at the beginning. Also i tried to use `not-contains` instead of the first `regexp` but this returned following error: 
`org.metafacture.metamorph.MetamorphException: Error while building the Metamorph transformation pipeline: not-contains not found.`
-  If multiple subfields $e appear all would trigger the process even if one of them has ZBD in it. e.g. 990197293880206441. Here the proposed workflow would ignore "ZDB...." but would pickup the "Springer" value. :
```
 <datafield tag="962" ind1=" " ind2=" ">
    <subfield code="e">ZDB-2-STI</subfield>
    <subfield code="e">Springer</subfield>
  </datafield>
```

Concerning 960:
We should check out if we really need an collection for all subfields. For 960 see: https://service-wiki.hbz-nrw.de/pages/viewpage.action?pageId=483000321


dr0i:
I don't understand what you mean with 

> I tried to combine the regexp for 962 but that did not work the results seem to be cut short at the beginning

Also don't understand the problem with:

> 
    <datafield tag="962" ind1=" " ind2=" ">
       <subfield code="e">ZDB-2-STI</subfield>
       <subfield code="e">Springer</subfield>
     </datafield>

Your morph treats this all right, choosing only `Springer` - the `ZDB-2-STI` appearing in the transformed json originates from `912` labeled as `eResource package` - and this is correct. The same is true for `*HT020391499.json`.
So I would say: +1.
Or did I miss something?

src/main/resources/alma/common/fields.xml
```diff
@@ -442,5 +442,28 @@
         </data>
       </entity>
     </entity>
+    <!-- E-Book and digital copy selections (inofficial) -->
+    <entity name="" sameEntity="true" reset="true">
+      <data source="962  .e" name="id">
+        <regexp match="^[^Z][^D][^B].*" />
+        <regexp match="(.*)" format="https://lobid.org/collections#${1}"/>
+      </data>
```
TobiasNx, line 450: @dr0i I was referring to this. At the moment I have two following regexp which in my opinion should be combined but the results are cut short, therefor I have two regexp. If you can combine them, feel free to change it.

TobiasNx (commented):
> I don't understand what you mean with
> 
> > I tried to combine the regexp for 962 but that did not work the results seem to be cut short at the beginning
> 
> Also don't understand the problem with:
> 
> > 
> 
> ```
> <datafield tag="962" ind1=" " ind2=" ">
>    <subfield code="e">ZDB-2-STI</subfield>
>    <subfield code="e">Springer</subfield>
>  </datafield>
> ```
> 
> Your morph treats this all right, choosing only `Springer` - the `ZDB-2-STI` appearing in the transformed json originates from `912` labeled as `eResource package` - and this is correct. The same is true for `*HT020391499.json`.
> So I would say: +1.
> Or did I miss something?

Since it is part of the same field, I do not know if this is truely an inofficial collection, or if this is just a name for the ZDB package. But being pragmatic I would leave it like this.

TobiasNx:
Also concerning the missing labels, should we aggregate all inofficial collections and create new labels for them?

src/main/resources/alma/common/fields.xml
dr0i: If you combine these like `"^[^Z][^D][^B](.*)"` you would always loose the first three characters.
I think two regexp are fine, they are also easy to understand. So +1.

dr0i:
Pinging @UBmakla for commenting on "labels for unofficial collections".

dr0i (approved):

PR #1315, opened 2021-10-05
Update path in schema test script

acka47:
The script wasn't running anymore so I fixed the path.


dr0i:
How do you execute the script? It's assumed that you are in `lobid-resources/src/test/resources` and execute the script like `bash validateJsonTestFiles.sh`.

acka47:
> How do you execute the script? It's assumed that you are in lobid-resources/src/test/resources and execute the script like bash validateJsonTestFiles.sh.

My bad, I tried to execute it from the root folder. Closing.

PR #1318, opened 2021-12-05
Fix JSON-LD context

acka47:
resolves #1317

Issue #1317, opened 2021-12-04
Samples not working in JSON-LD Playground


gregorbg:
Hello! Upon trying to examine the [default sample](https://lobid.org/resources/HT018472857?format=json) given on your [API documentation page](https://lobid.org/resources/api) by using the [JSON-LD playground](https://json-ld.org/playground/), I am confronted with the following error message:
```
jsonld.SyntaxError: Invalid JSON-LD syntax; a term definition must not contain type
```

I have tried manually switching to either `JSON-LD 1.0` or `JSON-LD 1.1` parsing under `Option`; both fail with the same error. Here's a [permalink for reference](https://tinyurl.com/y5reerjg).

By contrast, both the samples from [lobid-gnd](https://lobid.org/gnd/4074335-4.json) and [lobid-organisations](https://lobid.org/organisations/DE-6?format=json) _can_ be parsed in JSON-LD playground without any problems. I suspect that something is wrong with the `context` schema at http://lobid.org/resources/context.jsonld, because its formatting looks substantially different from the schemas of [lobid-gnd](http://lobid.org/gnd/context.jsonld) and [lobid-organisations](http://lobid.org/organisations/context.jsonld). But this is just a hunch, I am not knowledgeable enough in JSON-LD to exactly put my finger on it.

Thanks for working on this project, I really appreciate that you're making all of the catalogue data openly available! :smiley:

acka47:
Thanks for the detailed bug report. This should be fixed with #1318. The problem is that we used [aliased keywords](https://www.w3.org/TR/json-ld/#aliasing-keywords) not only in the instance data but in the JSON-LD context itself which apparently does not work.

acka47:
@gregorbg I am curious about your use case for lobid. Let us know if you plan to use it in you daily workflows and/or encounter any problems. You can reach us via lobid-admin \[at\] hbz-nrw \[dot\] de or via the chat at https://chat.gwdg.de/channel/lobid.

dr0i:
It's not solved yet, see  #1318.

gregorbg:
Hey, thank you for the quick fixes! As far as I'm concerned it works now; I've sporadically tested a few random HT entries and they were all processed by the JSON-LD Playground without any error. :+1:

I'd love to talk and get in touch some more about using the lobid API, I'll join the chat or write an email when I get the chance later this week.

dr0i:
Thx again @gregorbg for the report. We appreciate that.
[...]

PR #1319, opened 2021-12-06
Fix context.json by removing referenceType (#1317)

dr0i:
Follows up  #1318.

Removing the "referenceType" in context-labels.json fixes
"jsonld.SyntaxError: Invalid JSON-LD syntax; a term definition must not contain type".

The types of the resources can be lookuped by following the uris of the fields
(aka predicates).

### DIFF
--- a/src/main/resources/labels/context-labels.json
+++ b/src/main/resources/labels/context-labels.json
@@ -54,7 +54,6 @@
       "name" : "publishedBy"
    },
    {
-      "referenceType" : "http://www.w3.org/2001/XMLSchema#gYear",
       "label" : "Beginn",
       "uri" : "http://schema.org/startDate",
       "multilangLabel" : {},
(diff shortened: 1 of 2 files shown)


dr0i:
@acka47 I think #1317 can be solved by just removing the "referenceType". Also, the field/predicate of e.g. `startDate` should not be given explicitly but rather be resolved by following its URI (here: "http://schema.org/startDate", which leads to type "Date" which even contradicts a bit the formerly given  "referenceType" (aka: "@type") of "http://www.w3.org/2001/XMLSchema#gYear" ?).

The context is already deployed for convenience testing. I tested the sample given by @gregorbg in #1317 and think this good now.

acka47:
> Also, the field/predicate of e.g. startDate should not be given explicitly but rather be resolved by following its URI (here: "http://schema.org/startDate", which leads to type "Date" which even contradicts a bit the formerly given "referenceType" (aka: "@type") of "http://www.w3.org/2001/XMLSchema#gYear" ?)

I actually think it is a bug not a feature as we indicate how we use those fields – in a more restrictive way than upstream – in the lobid "application profile". schema.org just says, "use something ISO 8601 compliant". We clarify that we only use the [year format from ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Years). However, I think fixing the JSON-LD is more important than this and we could revisit this if necessary at another time. So, +1 from me.

acka47 (approved):

acka47:
Apparently, this is already deployed to production, see https://github.com/hbz/lobid-resources/issues/1317#issuecomment-987972247

dr0i:
Merged, closed.

Issue #1343, opened 2022-05-25
URI to lobid are always http instead of https in transformation


TobiasNx:
I suggest to change that.

e.g.:
https://github.com/hbz/lobid-resources/blob/4b14daf9302cc8cea455be7d7a9a6f98e53e1cf3/src/main/resources/alma/common/describedBy.xml#L62-L63

dr0i:
Done your suggestion.

dr0i:
Wait, there are a lot more.

TobiasNx:
Yes my example was only one of many.

dr0i:
Should be completed now.
@TobiasNx please check.

TobiasNx:
We need to update the schema too:
https://github.com/hbz/lobid-resources/pull/1346

dr0i:
As discussed in standup:
We can't just simply switch to https for RDF data.
We _can_ do this like in [4b6abb0](https://github.com/hbz/lobid-resources/commit/4b6abb06053b9b77a29580ac43408881db1dfbbf).
Other commits should be reverted.
Note to self: don't do code review and functional review in one person.

dr0i:
Reverted [bd88d02](https://github.com/hbz/lobid-resources/commit/bd88d0220b7cef473bdce8319bb1669703c01cf7).

TobiasNx:
This looks good now. +1

PR #1346, opened 2022-06-07
Update the schema to https instead of http URLs

TobiasNx:
Not only the transformation but also the schema needs to be updated.

https://github.com/hbz/lobid-resources/issues/1343


TobiasNx:
needed to add more changes to the transformation and the schema to get all `http` @dr0i could you have a look at the transformation why it fails? also @acka47 could you have a look at the schema changes?

src/test/resources/alma/(CKB)5280000000199164.json
```diff
@@ -85,8 +85,8 @@
       "label" : "Evangelisch-Lutherisches Missionswerk Leipzig"
     },
     "role" : {
-      "id" : "http://id.loc.gov/vocabulary/relators/isb",
-      "label" : "Herausgeber/in"
+      "id" : "https://id.loc.gov/vocabulary/relators/isb",
+      "label" : "isb"
```
dr0i, line 89: Why not keep the more verbose label?

src/test/resources/alma/(DE-605)HT000161712.json
```diff
@@ -166,8 +166,8 @@
       "id" : "https://d-nb.info/gnd/1023590484"
     },
     "role" : {
-      "id" : "http://id.loc.gov/vocabulary/relators/edt",
-      "label" : "Herausgeber/in"
+      "id" : "https://id.loc.gov/vocabulary/relators/edt",
+      "label" : "edt"
     }
```
dr0i, line 171: dito verbose

src/test/resources/alma/(DE-605)HT004285445.json
```diff
@@ -45,8 +45,8 @@
       "id" : "https://d-nb.info/gnd/134465202"
     },
     "role" : {
-      "id" : "http://id.loc.gov/vocabulary/relators/cmp",
-      "label" : "Komposition"
+      "id" : "https://id.loc.gov/vocabulary/relators/cmp",
+      "label" : "cmp"
     }
```
dr0i, line 50: better verbose

src/test/resources/alma/(DE-605)HT005207972.json
```diff
@@ -114,8 +114,8 @@
       "type" : [ "Person" ]
     },
     "role" : {
-      "id" : "http://id.loc.gov/vocabulary/relators/aut",
-      "label" : "Autor/in"
+      "id" : "https://id.loc.gov/vocabulary/relators/aut",
+      "label" : "aut"
     }
```
dr0i, line 119: dito

dr0i (changes requested):
See inline comments re verbosity. Haven't marked them all.

TobiasNx:
> See inline comments re verbosity. Haven't marked them all.

This seems to be due to the labelmaker not knowing about the changes from http -> https.

dr0i:
@UBmakla or @hagbeck : is the change of `http` to `https` in some way affecting you, or doesn't it matter?

acka47:
We should be careful about this. For DNB URIs and probably also for ZDB URIs this is fine as they switched to HTTPS URIs but MARC relators still use HTTP-only URIs (and do not provide a `sameAs` link or similar to the https URI), see e.g. https://id.loc.gov/vocabulary/relators/ill.html. So we should stick with that as this is important for the RDF representation.

acka47 (changes requested):
As noted in the other comment, we should continue using http-only URIs for MARC relator codes.

hagbeck:
The change to `https` is ok as long as the resulting URLs are not broken then ;-) 
We are using parts of some URL fields for selections and need to update these rules in some cases. But this is ok.

hagbeck:
Btw.: URLs to table of contents provided by hbz are currently `http`-only, too.

dr0i:
Conclusion so far:
Be aware of the generic URIs - even if e.g. `curl -L  --header "accept: text/plain" 'https://id.loc.gov/vocabulary/relators/aut'` brings up the proper `ntriples`, it's not reflected in the underlying data as [@acka47 noted ](https://github.com/hbz/lobid-resources/pull/1346#issuecomment-1152368724) as you can see in the data and in the missing `sameAs`. That's also true for our own [RPB-Skos](http://purl.org/lobid/rpb).
After you have checked `https` is ok, I would recommend to add this `https` variants to the shared label files (rather than have a copy of it ) - because it makes things easier to handle and consume and easier to exchange the label files).

Issue #1341, opened 2022-05-24
Move Alma transformation to Metafix


acka47:
- [x] Provide area for the fix transformation. A copy of the ALMA-Morph transformation. (@dr0i)
- [x] Start turning the morph transformation into fix.
- [x] alma.fix #1345 
- [x] titleRelatedFields.fix #1345 
- [x] contribution.fix #1351
- [x] item.fix
- [x] facets.fix
- [x] fields.fix
- [x] describedBy.fix
- [x] subjects.fix

Take in mind:
- #1340 there is a list of invalidations that need to be addressed in the new transformation
- #1339 hints missing elements that are important for NWBIB and are not in the ALMA morph transformation

Before we can start:
- #1345 must be closed

TobiasNx:
I reopen this ticket, since only the first part is done. The move of the alma transformation to fix will be done in steps over the next weeks.

TobiasNx:
One thing I noticed working with MARC in FIX is that the index-nr and the possible subfields can both have numeric path names. This should be taken in mind.

TobiasNx:
This can be closed since only details and frontend and backend implementations are needed to be done. Next steps are https://github.com/hbz/lobid-resources/issues/1435  #1342  #1434

PR #1354, opened 2022-06-21
1341-MoveAlmaTransformationToFix

TobiasNx:
Transform the describedBy morph into fix. Also move all other transformation to fix.

See #1341.


src/main/resources/alma/fix/subjects.fix
```diff
@@ -0,0 +1,397 @@
...
+        copy_field("$j", "subject[].$last.componentList[].$last.gndIdentifier")
+        replace_all("subject[].$last.componentList[].$last..gndIdentifier","^\\(DE-588\\)(.*)$","$1") 
+      end
+    end
+  end
+end
+
+# TODO: Find out why this does not work?
+# lookup("subject[].*.componentList[].*.type[].*","rswk-indicator")
+
```
TobiasNx, line 389: @blackwinter do you have an idea why this breaks?

TobiasNx:
```
Caused by: java.lang.IllegalArgumentException: Using ref, but can't find: 2 in: null
        at org.metafacture.metafix.FixPath.insertInto(FixPath.java:298)
        at org.metafacture.metafix.FixPath.insertInto(FixPath.java:265)
        at org.metafacture.metafix.FixPath.lambda$insertInto$20(FixPath.java:293)
        at org.metafacture.metafix.Value$TypeMatcher.match(Value.java:344)
        at org.metafacture.metafix.Value$TypeMatcher.ifArray(Value.java:317)
        at org.metafacture.metafix.FixPath.lambda$insertInto$22(FixPath.java:293)
```

TobiasNx: It runs with: `$ mvn clean install -DskipTests=true ; mvn -DgenerateTestData=true failsafe:integration-test -Dit.test=AlmaMarc21XmlToLobidJsonMetafixTest.java`

fsteeg: I'm afraid I have no quick fix here, it's actually surprising how that current path update logic works at all, I'll keep trying.

TobiasNx: Thanks this seems to work now!

TobiasNx:
Okay, this seems like a very huge branch and MR.
But besides the rest of TODOs the main construction for the new transformation seems done.

While changing the ALMA-transformation to fix I stumbled over some problems for that I needed to go back to the old ALEPH Morph especially with the `medium` and `type` elements. Here are some questions open concerning the mappings and the possibilities it seems that some info from Aleph is lost "forever". Others seem to have no proper mappings.

Things I suggest to do now is create an index from the new fix transformation. Repair all broken things with real data. Then start finishing and cleaning.

TobiasNx:
Added `hasItem` POR transformation also changed `oclcNumber` to array instead of `string`

Note to @hagbeck : when we move the ALMA Transformation to Fix, oclcNumber will be an array. Should we already implement this in the old ALMA-Morph, or is this change not relevant to you?

hagbeck:
@TobiasNx We are using the `oclcNumber` but the changes are not relevant for us now. So no changing of the ALMA-Morph is necessary.

dr0i:
Merged into master. We will go on using smaller PRs to solve expected issues. Closed.

PR #1369, opened 2022-08-09
1368 error fix

TobiasNx:
The error was due to the buggy behavior that `replace_all` with alterations demands multiple matching path otherwise it breaks.
See. https://github.com/metafacture/metafacture-fix/issues/249


dr0i (approved):

src/main/resources/alma/fix/describedBy.fix
```diff
@@ -29,9 +29,12 @@ set_array("describedBy.type[]", "BibliographicDescription")
 # 		<call-macro name="simplify-date" dname="dateModified" dsource="MNG  .d"/>
 copy_field("MNG  .b","describedBy.dateCreated")
 copy_field("MNG  .d","describedBy.dateModified")
-replace_all("describedBy.dateCreated|dateModified","-","")
-replace_all("describedBy.dateCreated|dateModified"," .*","")
-replace_all("describedBy.dateCreated|dateModified","c|©|\\s?|,|.|:|;|/|=","")
```
blackwinter, line 34: Any other potential bugs notwithstanding, I would strongly advise against relying on this wrong behaviour in the current Metafix implementation! See metafacture/metafacture-fix#143.

Issue #1390, opened 2022-08-16
ALMA-FIX: Links and relation metadata


TobiasNx:
- [ ] `sameAs`
- [x] `fulltextOnline[]`
- [ ] `hasVersion[]`
- [x] `similar[]`
- [x] `bibliographicCitation`
- [x] `containedIn[]`
- [x] `supplement[]`
- [x] `description[]`
- [x] `primaryForm[]`
- [x] `seeAlso[]`

These need to be added to the transformation.
These are quiet difficult to reconstruct from the old morph and many are buidling from the same Aleph elements `655`, `59?`, and `52?`

TobiasNx:
@acka47 primaryForm transformation in ALEPH MORPH say it links to primary and secondary form and vice versa. Isn't this contradicting the relation primaryForm and only true for the http://purl.org/dc/terms/isFormatOf which is the URI behind `primaryForm`.

https://github.com/hbz/lobid-resources/blob/69d0c8b011f83905e0114fc0a79069c86ab1e084/src/main/resources/morph-hbz01-to-lobid.xml#L1469-L1474

In ALMA I only used the link of the secondary to the primary, not the other way around. okay?

TobiasNx:
What is the difference between `sameAs` and `similar`?

dr0i:
From https://github.com/hbz/lobid-resources/issues/1475#issuecomment-1302141011: 
> Re ''seeAlso": I also can't find any examples, though I also couldn't make a comprehensive query. Just checked the 4 test files which formerly had that data, matched by:
```xml
 <data source="655[-eu][ -1].[x3]">
 <regexp match="^[zZ][uU][sS]"/>
</data>
```
> Seems to be gone from the catalog. I assume we can get rid of `seeAlso`

Depends on https://github.com/hbz/lobid-resources/issues/1509.

dr0i:
By configuring [index-config.json](https://github.com/hbz/lobid-resources/blob/master/src/main/resources/index-config.json) to index `seeAlso` I got a bunch of errors when ETLing the basedump, e.g.:
 > [1412]: index [resources-alma-fix-20221108-1045], type [resource], id [990219911120206441], message [MapperParsingException[failed to parse [seeAlso]]; nested: IllegalStateException[Can't get text on a START_OBJECT at 1:1711];]


So:
1. we have examples now (because of the error logs)
[...]

PR #1384, opened 2022-08-15
1383 spatial

TobiasNx:
See #1383

PR #1392, opened 2022-08-17
1390 links

TobiasNx:
Related to #1390 

Builds on #1384 and #1389

### DIFF
--- a/src/main/resources/alma/alma.fix
+++ b/src/main/resources/alma/alma.fix
@@ -56,6 +56,7 @@ retain(
   "related[]",
   "responsibilityStatement[]",
   "sameAs[]",
+  "seeAlso[]",
   "spatial[]",
   "subject[]",
   "subjectAltLabel[]",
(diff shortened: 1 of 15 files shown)


dr0i (changes requested):
+1 besides the missing test for `seeAlso`.
And the conflict, of course.

dr0i:
There _is_ a test file which has a `seeAlso` (potentially) in it:
```
[~/git/lobid-resources/src/test/resources/alma-fix] 1390-links
 grep 585270 '(DE-605)HT018295975.xml'
```
At least this was transformed to `seeAlso` using the aleph-sources , see `src/test/resources/jsonld/HT018295975.json`.

TobiasNx:
This seems to be an old test, this is not the productive version, see: http://lobid.org/resources/HT018295975.json

dr0i:
Right. I searched 14GB of marc xml and couldn't find any ` <datafield tag="856" ind1="4"` .  I guess we just leave it as it is. If someone complains, we will be provided with an example :P

dr0i:
I did a rebase on master and opened a [new PR](1393). I compared both branches and found them in sync, but @TobiasNx may want to check again. Than we could force push #1393 into this #1392 and close it.

PR #1397, opened 2022-08-22
1390-primaryForm

TobiasNx:
Both tests are needed since they include Reproduktion and Elektronische Reproduktion

Related to #1390


dr0i (approved):
+1 but if there is a chance to solve the "duplicate" question re "related" and "primary form" , it could made sense to discuss and solve that question now?

TobiasNx:
This is a question we need to solve with @acka47 especially with regard to all linked elements. I would suggest we open another ticket with regard to the differences of these elements.

dr0i:
ACK re new issue.

TobiasNx:
This PR is replaced by #1403

PR #1398, opened 2022-08-23
Add .editorconfig

TobiasNx:
The .editorconfig is borrowed from https://github.com/metafacture/metafacture-fix/blob/master/.editorconfig.

It can help to share format and style conventions across diffrent editors.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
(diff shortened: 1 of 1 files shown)


dr0i (approved):
Maybe it's a good idea to also point where you got that from, in case of updates from the other side from which we may benefit.

TobiasNx:
> Maybe it's a good idea to also point where you got that from, in case of updates from the other side from which we may benefit.

I added an inital description of the PR. Thanks.

PR #1399, opened 2022-08-23
1390-containedIn

TobiasNx:
See #1390. I added the  `containedIn`-array.
I transformed all 773 - Host Item Entry (R) with matching subfield $w .

### DIFF
--- a/src/main/resources/alma/fix/relatedRessourcesAndLinks.fix
+++ b/src/main/resources/alma/fix/relatedRessourcesAndLinks.fix
@@ -61,7 +61,7 @@ do list(path:"77008", "var":"$i")
   if any_match("$i.i", "Beil.*|Supp.*")
     add_field("supplement[].$append.test","")
     copy_field("$i.t","supplement[].$last.label")
-    copy_field("$i.w","supplement[].$last.id") 
+    copy_field("$i.w","supplement[].$last.id")
   end
 end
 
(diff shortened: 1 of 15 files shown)


dr0i (approved):
+1 but wouldn't that the point to solve the question first re redundance?

TobiasNx:
I need to reset this PR since it needs needs to include #1397 too

src/main/resources/alma/fix/relatedRessourcesAndLinks.fix
```diff
@@ -149,6 +149,25 @@ do list(path: "830??", "var": "$i")
   end
 end
 
+# containedIn
+# TODO: What is the difference between containedIn and isPartOf?
```
dr0i, line 153: Should this question be solved right now or do we want to postpone it ?

TobiasNx: Found some documentation on that: https://github.com/hbz/lobid-resources/issues/1400#issue-1347591143

`containedIn` is used with articles that are ressources depending on an umbrella publication as an journal issue or an anthology. The ressource is not published by itself.

`isPartOf`: is used with ressources that are published by themself but are part of an series, journal etc.

https://lobid.org/resources/api

TobiasNx:
This ticket is replaced by #1402

PR #1401, opened 2022-08-23
1372-fixError2

TobiasNx:
Prioritize this PR to the others. Will redo the other later.


dr0i (approved):
I think the big diff results from reformatting. I assume everything is ok (since the testfiles didn't change) but it would be good in the future to make first a sole commit for just reformatting so that one could see the diff which matters.

TobiasNx:
> I think the big diff results from reformatting. I assume everything is ok (since the testfiles didn't change) but it would be good in the future to make first a sole commit for just reformatting so that one could see the diff which matters.

This is due to the fact that I introduced a conditional and the code within the conditional needed to an additional indentation.

PR #1422, opened 2022-09-02
1409 do list urn error

TobiasNx:
I tested these they produce no error.


dr0i:
These errors may not be reproducible. They are most probably a result of https://github.com/metafacture/metafacture-fix/issues/255. You may close this PR @TobiasNx .

PR #1427, opened 2022-09-15
Do once schlagwortfolgen macro

TobiasNx:
@dr0i so lets try this. This should solve: https://github.com/metafacture/metafacture-fix/issues/258#issuecomment-1248185006

 I also deleted an alteration of subelements since that is not supported yet: https://github.com/metafacture/metafacture-fix/issues/249#issuecomment-1210945950


dr0i (approved):
yeah, that's it! Going to merge. May be reverted when https://github.com/metafacture/metafacture-fix/issues/258#issuecomment-1248266501 is fixed.

PR #1436, opened 2022-09-27
Add RPB-specific fields to index mapping config

fsteeg:
See https://jira.hbz-nrw.de/browse/RPB-27


dr0i:
Going to deploy at Wednesday.

dr0i:
It's part of `lobid-resources` index (`resources`).
However, `rpb` uses the index `resources-smalltest`, which is not updated with this setting.

PR #1457, opened 2022-10-05
copy_field in do list on the same field does not work

TobiasNx:


TobiasNx:
This was due to a wrong regexp.

Issue #1383, opened 2022-08-15
ALMA-Fix: Add `spatial`-element


TobiasNx:
This element is relevant for NWBib spatial ist yet missing.

TobiasNx:
@acka47 or @dr0i 

Is there a way I can identify if nwbib spatial concepts from subjects?

In ALMA is:
```
  <datafield tag="084" ind1=" " ind2=" ">
    <subfield code="a">03 ; Siebengebirge</subfield>
    <subfield code="2">nwbib</subfield>
  </datafield>
  ```

in Aleph:
```
<datafield tag="700" ind1="n" ind2="1">
<subfield code="a">03</subfield>
<subfield code="b">Siebengebirge</subfield>
</datafield>
```

This is related to this:
https://github.com/hbz/lobid-resources/pull/1045

TobiasNx:
For the moment excluded elements without URI.

Also `coverage` and geocoordinates need to be added at a later stage.

TobiasNx:
We will not add coverage. But spatial is now included: http://test.lobid.org/resources/99371527702306441.json

PR #1458, opened 2022-10-05
Add focus for spatial #1383

TobiasNx:
Added the `focus` part of `spatial`.

@dr0i is there still something missing?
@acka47 could you have a look if this is okay like this.

One question I still have there are two `label`s one in `spatial`, one in `focus`. Do they have the same value.
 Also can someone explain `coverage` to me? 

Example of the old morph:
https://github.com/hbz/lobid-resources/blob/f290c597648a9c38c0be24b53779c57387856b2d/src/test/resources/jsonld/BT000113394.json#L53

The `label`s and `notation`s of spatial are not the same as the string-value of `coverage`.

### DIFF
--- a/src/main/resources/alma/fix/subjects.fix
+++ b/src/main/resources/alma/fix/subjects.fix
@@ -204,8 +204,23 @@ do list(path:"084??", "var":"$i")
       set_array("spatial[].$last.type[]","Concept")
       add_field("spatial[].$last.source.id","https://nwbib.de/spatial")
       add_field("spatial[].$last.source.label","Raumsystematik der Nordrhein-Westfälischen Bibliographie")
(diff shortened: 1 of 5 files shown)


acka47 (approved):
+1

dr0i (approved):

dr0i:
Review after Sunday's ETL.

dr0i:
This errors with:
> [602]: index [resourcestest-alma-fix-20221010-1552], type [resource], id [990364455390206441], message [MapperParsingException[failed to parse]; nested: ElasticsearchParseException[[lat] and [lon] must be valid double values]; nested: NumberFormatException[For input string: "https://nwbib.de/spatial#Q2112"];]

So it seems that for `lat` or `lon` sometimes the URI is inserted, and ES denies indexing of this.

Issue #1471, opened 2022-10-13
Alma Fix: URIs falsely in lat/lon field


dr0i:
Results from https://github.com/hbz/lobid-resources/pull/1458:

>    [602]: index [resourcestest-alma-fix-20221010-1552], type [resource], id [990364455390206441], message [MapperParsingException[failed to parse]; nested: ElasticsearchParseException[[lat] and [lon] must be valid double values]; nested: NumberFormatException[For input string: "https://nwbib.de/spatial#Q2112"];]

So it seems that for lat or lon _often_ the URI is inserted, and ES denies indexing of this.

dr0i:
Wait, this may well be because `nwbib-spatial.tsv` couldn't be found at the server. Assigning myself.

dr0i:
After fixing the path (#1472), so that `nwbib-spatila.tsv` can be resolved, there are still some errors, like:

> [136]: index [resourcestest-alma-fix-20221010-1552], type [resource], id [99370768456106441], message [MapperParsingException[failed to parse]; nested: ElasticsearchParseException[[lat] and [lon] must be valid double values]; nested: NumberFormatException[For input string: "https://nwbib.de/spatial#N43"];]
[1650]: index [resourcestest-alma-fix-20221010-1552], type [resource], id [990110714900206441], message [MapperParsingException[failed to parse]; nested: ElasticsearchParseException[[lat] and [lon] must be valid double values]; nested: NumberFormatException[For input string: "http://www.wikidata.org/entity/Q1374337"];]

Do you have a clue @TobiasNx ?

TobiasNx:
This seems to be a problem since the values that should be looked up are not part of https://github.com/hbz/lobid-resources/blob/317e36a4e8cb4ba251c91673c9aa2e9558baf50d/src/main/resources/wd_itemLabelTypesCoordinates.tsv

TobiasNx:
Had a talk with @dr0i. We assumed that this is also broken in ALEPH-ALMA:
https://lobid.org/resources/BT000024700.json

It seems that https://github.com/hbz/lobid-resources/blob/317e36a4e8cb4ba251c91673c9aa2e9558baf50d/src/main/resources/wd_itemLabelTypesCoordinates.tsv has not been updated in a longer time.

dr0i:
See https://github.com/hbz/nwbib/issues/570#issuecomment-889746463 on how this file is build.

dr0i:
[...]

Issue #1473, opened 2022-10-13
ALMA-Fix: Use multi-column lookup for spatial


TobiasNx:
At  the moment we have a very weird lookup mechanism for spatial-lookups.

https://github.com/hbz/lobid-resources/blob/317e36a4e8cb4ba251c91673c9aa2e9558baf50d/src/main/resources/nwbib-spatial.tsv
```tsv
https://nwbib.de/spatial#Q890256	Boelerheide||http://www.wikidata.org/entity/Q890256
https://nwbib.de/spatial#Q2310026	Spiel (Titz)||http://www.wikidata.org/entity/Q2310026
https://nwbib.de/spatial#Q1968389	Pfalzdorf||http://www.wikidata.org/entity/Q1968389
https://nwbib.de/spatial#Q1621910	Hochkirchen||http://www.wikidata.org/entity/Q1621910
https://nwbib.de/spatial#Q5276	Altena|05962004|http://www.wikidata.org/entity/Q5276
```

https://github.com/hbz/lobid-resources/blob/317e36a4e8cb4ba251c91673c9aa2e9558baf50d/src/main/resources/wd_itemLabelTypesCoordinates.tsv

```tsv
http://www.wikidata.org/entity/Q61726052	Beek (Xanten),http://www.wikidata.org/entity/Q253019,Point(6.458805555 51.665583333)
http://www.wikidata.org/entity/Q20818804	Im Grafeld,http://www.wikidata.org/entity/Q253019,Point(7.092324 50.989545)
http://www.wikidata.org/entity/Q2809	Everswinkel,http://www.wikidata.org/entity/Q262166,Point(7.847777777 51.925)
http://www.wikidata.org/entity/Q1817500	Lendringsen,http://www.wikidata.org/entity/Q253019,Point(7.83056 51.4144)
http://www.wikidata.org/entity/Q382581	Soller,http://www.wikidata.org/entity/Q253019,Point(6.55583 50.7331)
```

To acess the later values we work with regexp.
It would be better  to use the multicolumn lookup that @blackwinter is preparing.

For that the created tsv  files should be created as multi-column tsv-files.

```tsv
https://nwbib.de/spatial#Q890256	Boelerheide		http://www.wikidata.org/entity/Q890256
https://nwbib.de/spatial#Q2310026	Spiel (Titz)		http://www.wikidata.org/entity/Q2310026
[...]

PR #1474, opened 2022-10-13
1471 update wd item label types coordinates.tsv

dr0i:
Se #1471.


dr0i:
@TobiasNx as discussed, the SPARQL results in the file `src/main/resources/nwbibWikidataLabelTypeCoords.csv`. Have a look if the file suits your needs. Note that the old file `wd_itemLabelTypesCoordinates.tsv` is deliberately _not_  overwritten with the new result because of structural integrity.

TobiasNx:
@dr0i could you document which elements we get from wikidata in plain text?

dr0i:
Not sure how to describe this better as looking at the first two lines of the tsv:
> nwbibId,item,itemLabel,types,coordinates
https://nwbib.de/spatial#Q1599076,http://www.wikidata.org/entity/Q1599076,Holthausen,http://www.wikidata.org/entity/Q1852178,Point(6.831666666 51.171944444)

where the first line is the name of the column and he second is an example.

TobiasNx:
The new lookupmechanism works, BUT it has some trouble with csv values that are protected by "....". Therefore I suggest to create a tsv instead of a csv. @dr0i could you change the fileformat and the sep-char?

src/test/resources/alma-fix/990110881770206441.json
```diff
@@ -85,10 +85,11 @@
     "focus" : {
       "id" : "http://www.wikidata.org/entity/Q6210",
       "label" : "Kreis Coesfeld",
-      "type" : [ "http://www.wikidata.org/entity/Q106658" ],
+      "type" : [ "\"http://www.wikidata.org/entity/Q106517174" ],
+      "@geoData" : " http://www.wikidata.org/entity/Q20738811\"",
```
TobiasNx, line 89: @dr0i This results because of the csv sep_char: `,`. MF does not know `"` as a protected value sign.

blackwinter:
> The new lookupmechanism works, BUT it has some trouble with csv values that are protected by "....".

FTR: This has nothing to do with the "new lookup mechanism". This has always been the case with Metafacture's file maps.

dr0i:
Done switching to `tsv`.

TobiasNx:
since getting the tsv some additional characters are added: I would need the tsv like this: `https://nwbib.de/spatial#Q2707	http://www.wikidata.org/entity/Q2707	Beckum	http://www.wikidata.org/entity/Q486972, http://www.wikidata.org/entity/Q54935786, http://www.wikidata.org/entity/Q42744322, http://www.wikidata.org/entity/Q707813	Point(8.040833333 51.755833333)`

TobiasNx:
@dr0i  could you review this. This seems to work now! It also includes also fix for https://github.com/hbz/lobid-resources/issues/1473 and works if geodata is missing.

blackwinter:
> Caused by: org.metafacture.metamorph.api.MorphExecutionException: File not found: /home/runner/work/lobid-resources/lobid-resources/alma/maps/classification.tsv

Note that `Metafix.resolvePath()` returns absolute paths now, so your previous "solution" to rely on `FileMaps`'s resource search doesn't work anymore.

Issue #1455, opened 2022-09-28
Add transformation for Article


TobiasNx:
https://github.com/hbz/lobid-resources/blob/ac8c976fa7f19b32adc2261d9623b0dc3a3190c7/src/main/resources/alma/fix/mediumAndType.fix#L406-L410

TobiasNx:
This seems to be an duplicate: #1314

TobiasNx:
Duplicate #1314

PR #1461, opened 2022-10-06
691-moreClassifications

TobiasNx:
I transformed the classifications to `subject`.
The structure would be there.

Question would be:
How do we proceed with classifications that do not have given labels?
How would we state the id of a concept?
What would be the label?
What are the id for the classification systems?

PR #1476, opened 2022-10-18
Add transformation for Articles in ALMA #1455 and #1314

TobiasNx:
Related to #1455 and #1314 

@acka47 : Could you help with the TODOs:
```

# TODO: What is Aufsatz/Artikel intended for only monograph, serial or continues ressource component?
# TODO: What about `leader07 a `and `b`? https://www.loc.gov/marc/bibliographic/bdleader.html
```

I only took up  `leader07 a` which is     
a - Monographic component part
but did not transformed:
    b - Serial component part

Also have a look why `"type" : [ "BibliographicResource", "Article", "Book" ],` article seems always accompanied by other `types`

### DIFF
--- a/src/main/resources/alma/fix/mediumAndType.fix
+++ b/src/main/resources/alma/fix/mediumAndType.fix
@@ -409,6 +409,9 @@ set_array("type[]","BibliographicResource")
 
 # TODO: Transformation needs to be added later.
 
+if any_match("@leaderPos06-07",".a")
+  add_field("type[].$append","Article")
+end
 
 # type: "Bibliography"
 
--- a/src/test/resources/alma-fix/990110509950206441.json
+++ b/src/test/resources/alma-fix/990110509950206441.json
@@ -152,6 +152,6 @@
     "label" : "Print",
     "id" : "http://rdaregistry.info/termList/RDAproductionMethod/1010"
   } ],
-  "type" : [ "BibliographicResource", "Map", "Book" ],
+  "type" : [ "BibliographicResource", "Article", "Map", "Book" ],
   "id" : "http://lobid.org/resources/990110509950206441#!"
 }
--- a/src/test/resources/alma-fix/990110714900206441.json
+++ b/src/test/resources/alma-fix/990110714900206441.json
@@ -114,7 +114,12 @@
   }, {
     "focus" : {
       "id" : "http://www.wikidata.org/entity/Q1374337",
-      "label" : "Euregio (Sitz: Gronau)"
+      "label" : "Euregio (Sitz: Gronau)",
+      "type" : [ "http://www.wikidata.org/entity/Q1374337" ],
+      "geo" : {
+        "lat" : "http://www.wikidata.org/entity/Q1374337",
+        "lon" : "http://www.wikidata.org/entity/Q1374337"
+      }
     }
   } ],
   "hasItem" : [ {
@@ -130,7 +135,7 @@
     "label" : "Print",
     "id" : "http://rdaregistry.info/termList/RDAproductionMethod/1010"
   } ],
-  "type" : [ "BibliographicResource", "Book" ],
+  "type" : [ "BibliographicResource", "Article", "Book" ],
   "contribution" : [ {
     "agent" : {
       "label" : "Prause, Klaus",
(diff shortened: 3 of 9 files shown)


acka47:
This is even harder for me to review than #1461  at this point as the PR includes commits from different unrelated issues (at least also the classification stuff and the reorganisation of the test files). As this would be taking too much of my time, I will be waiting for a cleaner PR to review.

TobiasNx:
Cleaned this up. Can you have another look @acka47 and help with the questions here:
https://github.com/hbz/lobid-resources/pull/1476#issue-1412984897

acka47:
Sorry, but this review led me down a rabbit hole. I think we have to take a step back and discuss the implementation also with regard to the `containedIn` property . Originally, we had created the property `containedIn` to be used only on resources of type `Article`. We created the `containedIn` property by ourselves to cover these use cases. From the property description (see http://purl.org/lobid/lv#containedIn):

> This property is used to link an article to a journal or an edited volume it is contained in.

Let's take a look at how `containedIn` is actually used.

## Problems in production lobid (Aleph, not to be fixed)

We apparently already got it wrong in the Aleph transformation as there are [354,375  resources with `containedIn` and without `"type": [ "Article" ]`](https://lobid.org/resources/search?q=_exists_%3AcontainedIn+AND+NOT+type%3AArticle&size=100). Skimming the results, I notice that `containedIn` is also used for musical scores in a collection of scores which also makes sense, I guess. 

So we take look at the [267,641 resources with `containedIn` and without type `Article` or `PublishedScore`](https://lobid.org/resources/search?q=_exists_%3AcontainedIn+AND+NOT+type%3A%28Article+OR+PublishedScore%29).  There are lots of resources in this list without an explicit type (i.e. [247,233 are of type `Miscellaneous`](https://lobid.org/resources/search?q=_exists_%3AcontainedIn+AND+NOT+type%3A%28Article+OR+PublishedScore%29+AND+type%3AMiscellaneous)). Looking at some examples, it seems like one could add type `Article` to them which would improve the data significantly. However, as we won't touch the old API, let's move on to the test API.

## The status in test.lobid.org/resources

Now, let's take a look at the status in the Alma-Fix transformation:

1. [639,455 resources with `containedIn` and without type `Article` or `PublishedScore`](https://test.lobid.org/resources/search?q=_exists_%3AcontainedIn+AND+NOT+type%3A(Article+OR+PublishedScore))
2. Many of the resources in 1.) are incorrectly typed as Book, see the list at [_exists_:containedIn AND NOT type:(Article OR PublishedScore)](https://test.lobid.org/resources/search?q=_exists_%3AcontainedIn+AND+NOT+type%3A%28Article+OR+PublishedScore%29+AND+type%3ABook)

There are some other types where `containedIn` also seems to make sense: `Map` ([example](https://test.lobid.org/resources/990197160910206441)), `Image` ([example](https://test.lobid.org/resources/990364910310206441)), `Report` ([example](https://test.lobid.org/resources/990110770560206441)). Filtering these out together with `Article` and `PublishedScore`, we get a list of [625,366 resources](https://test.lobid.org/resources/search?q=_exists_%3AcontainedIn+AND+NOT+type%3A%28Article+OR+PublishedScore+OR+Image+OR+PublishedScore+OR+Report%29) we should take a deeper look at.

## To Do

If I haven't overseen anything, we should adjust the transformation so that it checks that **all resources with a `containedIn` statement are either of type `Article`, `PublishedScore`, `Map`, `Image` or `Report`**. (I might have missed a type here.)

## Some other strange things I noticed and we will probably need extra tickets for

* This resource is typed `EditedVolume` though it looks like an aticle to me: https://test.lobid.org/resources/990194221820206441 Actually, there are [38,834 resources of type `EditedVolume` with a `containedIn` statement](https://test.lobid.org/resources/search?q=_exists_%3AcontainedIn+AND+type%3AEditedVolume+AND+NOT+type%3A%28Article+OR+PublishedScore%29+) where we should take a look at.
* This volume used `containedIn` where `isPartOf` would be the correct relation: https://test.lobid.org/resources/99371066229606441
* There are  [ resources of type `Standard`]() where a larger amount should rather have another type, for example there are lots of Dictionaries in there, see [`type:Standard AND title:Wörterbuch`](https://test.lobid.org/resources/search?q=type%3AStandard+AND+title%3AW%C3%B6rterbuch)
*

TobiasNx:
Replaced by by https://github.com/hbz/lobid-resources/pull/1563

Issue #1313, opened 2021-09-06
Use same Alma-resources as test resources that are used in Aleph


acka47:
To compare and monitor progress of the Alma2lobid ETL it would be nice to easily compare the same test resources (i.e. resource with the same hbz ID) coming from Aleph and Alma.

TobiasNx:
I added the Aleph-Testfiles in the branch https://github.com/hbz/lobid-resources/tree/1313-copyAlephTestfilesToAlma

Then I compared the ALMA and Aleph-Transforamtion-Results of the test:

ALMA is missing:

- [ ] contribution[].*.agent.altLabel[].*
- [ ] hasItem[].*.note[].*
- [ ] contribution[].*.agent.dateOfBirth
- [ ] spatial[].*.focus.type[].*
- [ ] isPartOf[].*.type[].*
- [ ] contribution[].*.agent.dateOfDeath
- [ ] fulltextOnline[].*.id
- [ ] fulltextOnline[].*.label
- [ ] contribution[].*.agent.altLabel
- [ ] subject[].*.componentList[].*.altLabel[].*
- [ ] spatial[].*.id
- [ ] spatial[].*.label
- [ ] spatial[].*.source.id
- [ ] spatial[].*.source.label
- [ ] spatial[].*.type[].*
- [ ] spatial[].*.notation
- [ ] coverage[].*
- [ ] spatial[].*.focus.id
- [ ] hasVersion[].*.id
- [ ] hasVersion[].*.label
- [ ] similar[].*.id
- [ ] similar[].*.label
- [ ] spatial[].*.focus.geo.lat
- [ ] spatial[].*.focus.geo.lon
- [ ] bibliographicCitation
- [ ] spatial[].*.focus.label
- [ ] containedIn[].*.id
- [ ] containedIn[].*.label
- [ ] urn[].*
- [ ] exampleOfWork.label
- [ ] exampleOfWork.type[].*
- [ ] publication[].*.note[].*
- [ ] publication[].*.publishedBy[].*
- [ ] subject[].*.componentList[].*.dateOfBirth
- [ ] contribution[].*.agent.source.id
- [ ] contribution[].*.agent.source.label
- [ ] containsExampleOfWork[].*.label
- [ ] containsExampleOfWork[].*.type[].*
- [ ] oclcNumber[].*
- [ ] subject[].*.componentList[].*.dateOfDeath
- [ ] containsExampleOfWork[].*.creatorOfWork
[...]

PR #1477, opened 2022-10-18
1471-addDeleteForLookup

TobiasNx:
This should stop elastic search from breaking, see https://github.com/hbz/lobid-resources/issues/1471 . But does not fix the problems with the missing geo-data. This would need the solution in #1473


TobiasNx:
This branch is now upbranched from #1313

TobiasNx:
@dr0i you can go on with this one too.

dr0i (changes requested):
src/test/resources/alma-fix/almaMarcXmlTestFiles is only a temporary file (it is tar.bz2 archived). I will add the removal of this file in src/test/resources/alma/updateAlmaTestFiles.sh  after the archive is created so that no traces are in the file system.

dr0i (approved):
Very good that you did this in two commits - one can see that before your change the lat/lon was part of the ETL.

PR #1479, opened 2022-10-18
1313 compare test results of aleph json with alma json

dr0i:
See #1313.


src/test/resources/alma-fix/almaFilesFromIdsOfAlephTestFiles/grepAlephTestIdsAndLookupAlmaJsonAndXml.sh
```diff
@@ -0,0 +1,18 @@
...
+
+for hbzId in $(grep hbzId ../../jsonld/*.json | cut -d '"' -f4 |grep -v _); do
+  echo $hbzId
+  json=$(curl "http://test.lobid.org/resources/search?q=hbzId:${hbzId}&format=json"| jq .member[0])
+  almaMmsId=$(echo "$json" | jq .almaMmsId | tr -d '"')
+  echo "almaID:'$almaMmsId'"
+  if [ -n "${almaMmsId}" -a "${almaMmsId}" != "null" ]; then
+    xml=$(curl https://indexes.devel.digibib.net/export/${almaMmsId} | xmllint --format -)
+    echo "$json" > "${almaMmsId}_${hbzId}.json"
+    echo "$xml" > "${almaMmsId}_${hbzId}.xml"
```
TobiasNx, line 13:
```
    echo "$json" > "${almaMmsId}_${hbzId}_aleph.json"
    echo "$xml" > "${almaMmsId}_${hbzId}_aleph.xml"
```
perhaps make more explicit that this should be the aleph file

TobiasNx:
I tested the new update script for fix and the script for the aleph files of the ALMA tests. This seems to work great :)

TobiasNx:
@dr0i you can go on!

TobiasNx (approved):

dr0i:
I woul vote against commiting all these (>290) files:
- we can get them if we need to by running the script
- they are not part of an automatic process
- they may change often (every time you improve the data to become more like the blue prints of ALEPH's files) so you need to run the script anyway
- they become obsolete if ALMA's ETL outcom is in sync with ALEPH's ETL outcome

Or do you have an actual need in them to be commited and revisioned?

TobiasNx:
> I woul vote against commiting all these (>290) files:
> 
>     * we can get them if we need to by running the script

Takes time

> 
>     * they are not part of an automatic process

valid reason

> 
>     * they may change often (every time you improve the data to become more like the blue prints of ALEPH's files) so you need to run the script anyway

ALEPH-Files do not change often, since there are no new changes to the ALEPH-Morph. Updates can be added if new Alma-test files are added. Or if needed.

> 
>     * they become obsolete if ALMA's ETL outcom is in sync with ALEPH's ETL outcome
Yes, you can kick them out when files are more in sync but you can kick out the Aleph-Morph-Test and the ALMA-Morph when obsolete.


> 
> 
> Or do you have an actual need in them to be commited and revisioned?
They help the reviewer (@acka47) and me to compare the files easily.

I would vote to keep them, and get rid of them at a later stage.

dr0i:
Thinking about it, it also makes even less sense to update `src/test/resources/alma-fix/*.[json|xml]` . What is the purpose of this? They mostly  only differ in holding data (which is highly fluent anyway), while sometimes even possibly tested data like (`subjectAltLabel` in 990011470300206441) disappears. Have you checked if all the test cases are still covered? Note that these test cases may be made up by purpose, not necessarily reflecting the data of the catalog. And this is fine (see "tests covered").

> They help the reviewer (@acka47) and me to compare the files easily.

@acka47 would just run the script to get the data. You have 4 computers and would have to run the script 4 times locally . I like this much more that commiting 150 * 10kb files. Mind that this "ephemeral/tmp" data will _always_ be part the repo's git history. Is this worth it? If you like, I could make an archive and publish it on a tmp directory at our webserver instead, if you really want not to execute the script.

TobiasNx:
updating testfiles was introduced purposely with ALMA-Morph.

TobiasNx:
Talked with @dr0i off board decided not only to revert but to reset the commits.

dr0i:
Thx! Merged.

Issue #1482, opened 2022-10-20
ALMA-Fix: Add `corporateBodyForTitle`


TobiasNx:
Seems to be based on. https://github.com/hbz/lobid-resources/issues/33

Mapping ALEPH -> ALMA is:

333# $a [nur RAK] | 245 10 $a ... / < 333# $a > | Zu ergänzende Urheber zum Haupttitel [nur RAK]Hz: Neu
-- | -- | --

https://service-wiki.hbz-nrw.de/pages/viewpage.action?pageId=483000321

TobiasNx:
It seems that the info is old RAK metadata. See [HT007015768](https://lobid.org/resources/HT007015768):

```
 "corporateBodyForTitle": [
    "Kommission der Europäischen Gemeinschaften, Zentraldienst Archiv, Dokumentation, Abteilung IX/C/3"
  ],
```

=

```
          <datafield tag="331" ind1="-" ind2="1">
            <subfield code="a">Dokumentationsbulletin</subfield>
          </datafield>
          <datafield tag="333" ind1="-" ind2="1">
            <subfield code="a">Kommission der Europäischen Gemeinschaften, Zentraldienst Archiv, Dokumentation, Abteilung IX/C/3</subfield>
          </datafield>
```

In [ALMA](https://alma.lobid.org/marcxml/990055713370206441):

```
<datafield tag="245" ind1="1" ind2="0">
    <subfield code="a">Dokumentationsbulletin / Kommission der Europ&#228;ischen Gemeinschaften, Zentraldienst Archiv, Dokumentation, Abteilung IX/C/3</subfield>
    <subfield code="b"> = Bulletin de renseignement documentaire / Commission des Communaut&#233;s Europ&#233;ennes, Service Central Archives-Documentation (SCAD), Division IX/C/3 = Bollettino di informazione documentaria / Commissione delle Comunit&#224; Europee, Servizio Centrale Archivi-Documentazione, Divisone IX/C/3</subfield>
    <subfield code="n">[...]</subfield>
    <subfield code="p">A</subfield>
    <subfield code="n">[...]</subfield>
    <subfield code="p">Sondernummer</subfield>
  </datafield>
```

@acka47 should we drop `corporateBodyForTitle` then?

acka47:
As there is no MARC21 equivalent to `333` (see below), we won't need `corporateBodyForTitle` any more.

[...]

PR #1483, opened 2022-10-20
1482-corporateBodyForTitle

TobiasNx:


TobiasNx:
Due to decision in #1482 closed.

Issue #1419, opened 2022-09-01
publication.publicationHistory is recorded for too many resources


acka47:
E.g. all monographs currently have this information although it should only be found in descriptions of series and periodicals: `$ curl "https://alma.lobid.org/resources/search?q=type:Book&format=json" | jq .member[].publication[].publicationHistory`

The question is how to filter for resources that might contain this field. One way might be to only add this for ZDB resources.

TobiasNx:
@acka47 hinted to problems with publicationHistory:

> I just took a short look at it. At least the `publicationHistory` needs some improvement. Currently it only reads "anfangs" for two resource. I guess, we will have to include more (all?) subfields from Marc 264 when building `publicationHistory` and currently wea re only seem to be using subfield `a`.
> 
> See both test files with 264:
> 
> https://github.com/hbz/lobid-resources/blob/c3425a1dd047fd44403e15a643c299e507797781/src/test/resources/alma-fix/990108740950206441.xml#L54-L66
> 
> https://github.com/hbz/lobid-resources/blob/c3425a1dd047fd44403e15a643c299e507797781/src/test/resources/alma-fix/990053976760206441.xml#L86-L95
> 
> It is always good to also check how the ZDB RDF looks (here e.g. http://ld.zdb-services.de/data/123550-3.ttl) but in the case of MARC 26, they are not transforming it to RDF. Looking at the HTML view, though, one sees that they just show the contents of 362a for Publication History/Erscheinungsverlauf, see https://ld.zdb-services.de/resource/123550-3 (Click on "More title information") and https://ld.zdb-services.de/resource/1484300-6.
> 
> https://github.com/hbz/lobid-resources/blob/c3425a1dd047fd44403e15a643c299e507797781/src/test/resources/alma-fix/990053976760206441.xml#L96-L98
> 
> https://github.com/hbz/lobid-resources/blob/c3425a1dd047fd44403e15a643c299e507797781/src/test/resources/alma-fix/990108740950206441.xml#L67-L69
> 
> So we should do the same and skip 264.

https://github.com/hbz/lobid-resources/pull/1488#issuecomment-1290593062

I should fix this within this ticket.

TobiasNx:
This should be fixed with #1631

dr0i:
This can be checked beginning with 20.2.2023.

TobiasNx:
[...]

PR #1488, opened 2022-10-21
1484-publicationEnhancement

TobiasNx:


dr0i (approved):

TobiasNx:
@acka47  can you help this is still a draft, but to advance your help would be appreciated

acka47:
I just took a short look at it. At least the `pulicationHistory` needs some improvement. Currently it only reads "anfangs" for two resource. I guess, we will have to include more (all?) subfields from Marc 264 when building `publicationHistory` and currently wea re only seem to be using subfield `a`.

See both test files with 264:

https://github.com/hbz/lobid-resources/blob/c3425a1dd047fd44403e15a643c299e507797781/src/test/resources/alma-fix/990108740950206441.xml#L54-L66

https://github.com/hbz/lobid-resources/blob/c3425a1dd047fd44403e15a643c299e507797781/src/test/resources/alma-fix/990053976760206441.xml#L86-L95

It is always good to also check how the ZDB RDF looks (here e.g. http://ld.zdb-services.de/data/123550-3.ttl)  but in the case of MARC 26, they are not transforming it to RDF. Looking at the HTML view, though, one sees that they just show the contents of 362a for Publication History/Erscheinungsverlauf, see https://ld.zdb-services.de/resource/123550-3 (Click on "More title information") and https://ld.zdb-services.de/resource/1484300-6.

https://github.com/hbz/lobid-resources/blob/c3425a1dd047fd44403e15a643c299e507797781/src/test/resources/alma-fix/990053976760206441.xml#L96-L98

https://github.com/hbz/lobid-resources/blob/c3425a1dd047fd44403e15a643c299e507797781/src/test/resources/alma-fix/990108740950206441.xml#L67-L69

So we should do the same and skip 264.

acka47:
I will take another look when you have cmomitted a fix for the `publicationHistory`, @TobiasNx .

TobiasNx:
I would suggest to move publicationHistory to this ticket: `https://github.com/hbz/lobid-resources/issues/1419`
Here I only want to discuss `publication.note` and `publication.frequency` here, since they are new.

acka47 (approved):
+1 Looks perfect. Thanks!

Issue #1491, opened 2022-10-24
ALMA-Fix: Add license


TobiasNx:

dr0i:
Looks good, see e.g. http://test.lobid.org/resources/990210781980206441.json.

PR #1492, opened 2022-10-24
Add license element #1491

TobiasNx:
#1491

### DIFF
--- a/src/main/resources/alma/alma.fix
+++ b/src/main/resources/alma/alma.fix
@@ -40,6 +40,7 @@ retain(
   "ismn[]",
   "isPartOf[]",
   "language[]",
+  "license[]",
   "medium[]",
   "natureOfContent[]",
   "note[]",
--- a/src/main/resources/alma/fix/otherFields.fix
+++ b/src/main/resources/alma/fix/otherFields.fix
@@ -130,3 +130,14 @@ replace_all("bibliographicCitation.*","<<|>>","")
 replace_all("bibliographicCitation.*","^http","Siehe: http")
 
 join_field("bibliographicCitation", "; ")
+
+# 542 - Information Relating to Copyright Status (R) - $n - Note (R)  $u - Uniform Resource Identifier (R)
+
+set_array("license[]")
+do list(path: "542??", "var": "$i")
+  copy_field("$i.u", "license[].$append.id")
+  if exists("$i.u")
+    set_array("license[].$last.note[]")
+    copy_field("$i.n", "license[].$last.note[].$append")
+  end
+end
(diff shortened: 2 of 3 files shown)


TobiasNx:
@droi label seems strange.

acka47 (changes requested):
Currently, the information from 543n is added as `note` statement to the license object but the note is actually about the work. So we should either leave it out or move this information to the bibliographic resource using a property like`https://creativecommons.org/ns#useGuidelines` or similar.

dr0i:
Yes it's the EtikettMaker, which works as a fallback if no `label` is explicitly given, to guarante a `label` for every `id` . So the `label` is derived from `https://creativecommons.org/licenses/by-nd/4.0/"` by taking the last path segment of the URI.

dr0i:
Some proposals re 'label':
-Maybe you could rename the field "note" to "label" - "label: 4.0" will then not be generated. I don't know if this is feasible, e.g if all licenses have a "note" field in the data. That the `label` would be very long is not such a problem in my view, because I think seldom a UI would show this field as default.
- You also make a lookup table to provide e.g. "CC-BY-ND 4.0" or so  - but then you have to know all kind of licenses resp. those which are not part of the lookup table may have strange labels
- use the URI as label

TobiasNx:
> Currently, the information from 543n is added as `note` statement to the license object but the note is actually about the work. So we should either leave it out or move this information to the bibliographic resource using a property like`https://creativecommons.org/ns#useGuidelines` or similar.

I am okay with leaving out the `note` element. It is already used as this in the old ALEPH-Moprh and provides the same info:
https://lobid.org/resources/HT018907266.json

acka47:
As just discussed in the meeting, we will remove `license.note` and use the URL – without `http(s)` as label.

TobiasNx:
@dr0i do you set the label? or should this be done by the transformation-fix?

dr0i:
No, [EtikettMaker won't set the label as you want it](https://github.com/hbz/lobid-resources/pull/1492#issuecomment-1290256998). You have to set it yourself.

TobiasNx:
> No, [EtikettMaker won't set the label as you want it](https://github.com/hbz/lobid-resources/pull/1492#issuecomment-1290256998). You have to set it yourself.

but somehow this works with webPageArchived?

https://github.com/hbz/lobid-resources/blob/b4adf01e9ac2f58f3ccbfbfc3f1f7bdcd4a07008/src/main/resources/alma/fix/mediumAndType.fix#L388-L396

https://github.com/hbz/lobid-resources/blob/59e70be11115475439d28ee7dca8dbb76fa33421/src/test/resources/alma-fix/990217495840206441.json#L149-L152

dr0i:
There is no URI path in `http://www.klimaatlas.koblenz.de/`. Hence it fallbacks to generate a label from the (sub)domain, leaving out the protocol.

dr0i (approved):

acka47 (approved):
+1

dr0i:
I grab this, as discussed with @acka47 , and merge and deploy it.

PR #1493, opened 2022-10-25
Add element "titleKeyword"

TobiasNx:
Sidenote: HT017551955 and CT003043468 are those dublicates?


acka47 (approved):
+1

dr0i:
Hip-shot re: duplicates:
All CT* are digitized versions from their physical originals. So they librarians don't  qualify these as duplicates.

Issue #1373, opened 2022-08-10
Properly model MBD, POR, H52, ITM, etc.


TobiasNx:
At the moment these are all modeled as `hasItem` but they are not all `Item`s therefore we should remodel these.
At first we need better documentation what the elements represent. #1363

TobiasNx:
@hagbeck and @UBmakla : at the moment we dump all holding and portfolio related [ALMA-hbz-specific elements](https://service-wiki.hbz-nrw.de/pages/viewpage.action?pageId=612532300) into `hasItem` differentiated by a messy `type`-Value:

- MNG Management Information
- HOL Additional Holdings Fields (HOL)
- ITM Items Information (ITM)
- POR Electronic Portfolio Information (POR)
- POC Portfolio Coverage Information (POC)
- POE Portfolio Embargo Information (POE)
- MBD Members Information (MBD)
- Holdings information (Hxx):
- H16 	MARC 016 	National Bibliographic Agency Control Number |  
- H16 	[MARC 016](https://www.loc.gov/marc/holdings/hd016.html) 	National Bibliographic Agency Control Number	
- H24 	MARC 024 	Other Standard Identifier	
- H35 	MARC 035 	System Control Number	
- H92 	LOCAL 092 	
- H93 	LOCAL 093 	
- H94 	LOCAL 094 	
- H62 	MARC 562 	Copy and Version Identification Note	
- H83 	MARC 583 	Action Note	
- H89 	LOCAL 689 	
- H95 	LOCAL 695 	
- H42 	MARC 842 	Textual Physical Form Designator	
- H43 	MARC 843 	Reproduction Note	
- H52 	MARC 852 	Location	
- H56 	MARC 856 	Electronic Location and Access	
- H59 	LOCAL 859 	
- H66 	MARC 866 	Textual Holdings - Basic Bibliographic Unit	
- H69 	LOCAL 869 	
- H80 	LOCAL 980 	
- HAD 	LOCAL 992 	Alte Drucke/Provenienz


But we intend to clean up und remodel the holding/portfolio data better. Could you help us to specify which elements you as prime customer no1 need, for which purpose and what they entail ?

hagbeck:
Hi,
at first: the link to the wiki page result in "page not found".
The dump of all these "items" was a workaround because of metafacture-morph limitations. For me it would make sense to rethink the goals of the data and try to make it better with metafacture-fix.
We suggest to discuss this in a video call, perhaps on November 11th?

[...]

PR #1519, opened 2022-11-21
Exclude POR, HOL, MDB from hasItem #1373

TobiasNx:
#1373

POR should be excluded, but still is in the tests. Also MDB and HOL items are still in the tests.


dr0i:
- restored the corrupted test archive (lots of missing test file there)
- rebased using the master
<= force pushed
(please ignore the newly structured XMLs - indentation is the only change here)

PR #1528, opened 2022-11-25
Rename resources.conf (#1524)

### LINKED ISSUE #1524: Rename resources.conf to avoid overwriting it when pulling
We should rather use a template then use this conf directly, because a `git pull` may overwrite it and thus breaks our specific configs.

Helps to avoid https://github.com/hbz/lobid-resources/issues/1513#issuecomment-1326008647.

dr0i:
This avoids overwriting of your specific web/resources.conf.
Resolves #1524.


dr0i:
Already rolles out to servers to have a quick functional fix.

fsteeg (approved):

Issue #1526, opened 2022-11-24
Institution Lookup for ITM is broken.


TobiasNx:
Uh, doesnot work on server:
>  Error while executing Fix expression (at file:/home/sol/git/lobid-resources-test/web/target/universal/stage/conf/alma/fix/item.fix, line 58): lookup("hasItem[].$last.heldBy.id", "alma-institution-code-to-isil")
org.metafacture.metamorph.api.MorphExecutionException: File not found: conf/alma/./maps/institutions.tsv

_Originally posted by @dr0i in https://github.com/hbz/lobid-resources/issues/1520#issuecomment-1326142143_
      

Diskussed this, for now we should use a variable. Next time we should have a look why our FIX does not work with the dot-notation for the path.

TobiasNx:
Seems to be live: http://test.lobid.org/resources/990112775700206441.json

dr0i:
Great, the let's close this.

Issue #1527, opened 2022-11-24
Lookup file refrence does not work with dot notation


TobiasNx:
https://github.com/hbz/lobid-resources/issues/1526

```

put_map("alma-iz-code-to-isil",
put_filemap("./maps/institutions.tsv","alma-iz-code-to-isil", sep_char:"\t",key_column:"2",value_column:"0",expected_columns:"-1")
  "49HBZ_UBD": "DE-290",  # Universitätsbibliothek Dortmund -->
put_filemap("./maps/institutions.tsv","alma-institution-code-to-isil", sep_char:"\t",key_column:"1",value_column:"0",expected_columns:"-1")
  "49HBZ_BIE": "DE-361",  # Universitätsbibliothek Bielefeld -->


```

This works locally but does not work on the server. Returns:

```
    Error while executing Fix expression (at file:/home/sol/git/lobid-resources-test/web/target/universal/stage/conf/alma/fix/item.fix, line 58): lookup("hasItem[].$last.heldBy.id", "alma-institution-code-to-isil")
    org.metafacture.metamorph.api.MorphExecutionException: File not found: conf/alma/./maps/institutions.tsv
```

We should investigate the problem. So that we do not need the workaround with the variables.

blackwinter:
You're [using](https://github.com/hbz/lobid-resources/blob/3230d48a340bbe92d3e73bb67c6c0e6a1cfa0a90/web/app/controllers/resources/WebhookAlmaFix.java#L49) `conf/alma/alma.fix` instead of `src/main/resources/alma/alma.fix` as `MORPH_FILENAME`. No idea how the Fix file ends up there, but it means that all relative paths are resolved based on the new location.

dr0i:
That's because the play app uses lobid-resources as library. This setting:
> unmanagedResourceDirectories in Compile += baseDirectory.value / "../src/main/resources/"
copies all the files found in `src/main/resources` to `./web/target/universal/stage/conf/`. So e.g. `./web/target/universal/stage/conf/alma/maps/institutions.tsv` is in place, relative to the fix.
I think this was working with morph, but with fix it isn't. Not sure, have to investigate.

TobiasNx:
> That's because the play app uses lobid-resources as library. [This setting](https://github.com/hbz/lobid-resources/blob/master/web/build.sbt#L8):
> 
> > unmanagedResourceDirectories in Compile += baseDirectory.value / "../src/main/resources/"
> 
[...]

PR #1529, opened 2022-11-25
Fix resolving file "institutions.tsv" (#1526)

dr0i:
Similar to f59c5d09ea28e20775910bcc1c5356d75fb8fda2.

See #1526.

### DIFF
--- a/src/main/java/org/lobid/resources/run/AlmaMarcXmlFix2lobidJsonEs.java
+++ b/src/main/java/org/lobid/resources/run/AlmaMarcXmlFix2lobidJsonEs.java
@@ -113,6 +113,7 @@ public class AlmaMarcXmlFix2lobidJsonEs {
                 fixVariables.put("deweyLabels", "deweyLabels.tsv");
                 fixVariables.put("nwbib-spatial", "nwbib-spatial.tsv");
                 fixVariables.put("wd_itemLabelTypesCoordinates", "wd_itemLabelTypesCoordinates.tsv");
+                fixVariables.put("maps-institutions.tsv", "alma/maps/institutions.tsv");
 
                 XmlElementSplitter xmlElementSplitter = new XmlElementSplitter();
                 xmlElementSplitter.setElementName("record");
--- a/src/main/resources/alma/fix/maps.fix
+++ b/src/main/resources/alma/fix/maps.fix
@@ -1,5 +1,5 @@
-put_filemap("./maps/institutions.tsv","alma-iz-code-to-isil", sep_char:"\t",key_column:"2",value_column:"0",expected_columns:"-1")
-put_filemap("./maps/institutions.tsv","alma-institution-code-to-isil", sep_char:"\t",key_column:"1",value_column:"0",expected_columns:"-1")
+put_filemap("$[maps-institutions.tsv]","alma-iz-code-to-isil", sep_char:"\t",key_column:"2",value_column:"0",expected_columns:"-1")
+put_filemap("$[maps-institutions.tsv]","alma-institution-code-to-isil", sep_char:"\t",key_column:"1",value_column:"0",expected_columns:"-1")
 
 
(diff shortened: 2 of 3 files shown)


blackwinter:
Sorry, but this doesn't explain why _relative_ paths (relative to the Fix file, that is) are resolved differently in CLI and web app. The working directory is different, fine; but the directory structure is not (as you explained in #1527). And why does `include()` behave differently still? :confused: 

Also, you set both variables to the same value. How does this work?

dr0i:
> Sorry, but this doesn't explain ...

Ack. This should be solved in [another issue](https://github.com/hbz/lobid-resources/issues/1527).
This PR here is just a quick working workaraound.

> Also, you set both variables to the same value. How does this work?

You mean
https://github.com/hbz/lobid-resources/pull/1529/files#diff-99a8a9f4045726c27c432ada7d278e20550881c5babd5ec4b2b1fcafa3197b13L2 ?
That's _two_ variables set to one location.

blackwinter:
> Ack. This should be solved in [another issue](https://github.com/hbz/lobid-resources/issues/1527).

Okay.

> You mean
> https://github.com/hbz/lobid-resources/pull/1529/files#diff-99a8a9f4045726c27c432ada7d278e20550881c5babd5ec4b2b1fcafa3197b13L2 ?
> That's _two_ variables set to one location.

No, I meant [AlmaMarcXmlFix2lobidJsonEs](https://github.com/hbz/lobid-resources/pull/1529/files#diff-8057d2f6f68f3245d1e3843c6323177281a5f79498da99191ef40c016dcf019bR116) and [AlmaMarc21XmlToLobidJsonMetafixTest](https://github.com/hbz/lobid-resources/pull/1529/files#diff-28eb094d191f895ff3c63cee0c419ae1e0f50919e52c7809757ef129058e4743R64). But that's not even CLI vs. web app, that's CLI vs. test, right? I don't understand how this fixes anything w.r.t. the web app. But it's none of my business anyway, so I'll leave you in peace now ;)

TobiasNx (approved):
+1 runs locally without changes.

dr0i:
> So you effectively only changed the path, the introduction of the variable is inconsequential?

funnily, no , it _is_ of consequence - if only the path is changed in `src/main/resources/alma/fix/maps.fix` the  commands  `mvn clean install -DskipTests=true; mvn -DgenerateTestData=true failsafe:integration-test -Dit.test=AlmaMarc21XmlToLobidJsonMetafixTest` result in:

> Caused by: org.metafacture.metamorph.api.MorphExecutionException: File not found: src/main/resources/alma/./alma/maps/institutions.tsv
        at org.metafacture.metamorph.maps.FileMap.lambda$null$0(FileMap.java:211)
        at java.util.Optional.orElseThrow(Optional.java:290)

blackwinter:
> funnily, no , it _is_ of consequence - if only the path is changed

So you changed the path from `./maps/institutions.tsv` to `alma/maps/institutions.tsv` (on both lines) and get `File not found: src/main/resources/alma/./alma/maps/institutions.tsv`? How is this possible? Where does the `./` part come from then? It looks like we must have a serious problem somewhere...

dr0i:
Agreed. May have also affected https://github.com/hbz/lobid-resources/actions/runs/3564501298/jobs/5989728745#step:7:557 , where the build breaks (and it does so only at GitHub, locally `mvn clean install` runs fine). This is weird.

blackwinter:
1. I didn't understand why changing the path to `alma/maps/institutions.tsv` would work at all. But that's due to `FileMap` trying to open the path as a resource if opening it as a file fails. That's also why you saw different outcomes for maps and includes.
2. I still don't understand why your build failed. But metafacture/metafacture-fix@7635572 changed `Metafix.resolvePath()` to return an absolute path and so you should get consistent behaviour now. Of course, this also means that you have to revert your changes relying on resource search.

Issue #1470, opened 2022-10-11
Fix error in how we model provenance data?


acka47:
Currently we have `dateCreated` and `dateModified` statements about the RDF description, e.g.:

```json
{
   "describedBy":{
      "license":[
         {
            "id":"http://creativecommons.org/publicdomain/zero/1.0",
            "label":"Creative Commons-Lizenz CC0 1.0 Universal"
         }
      ],
      "inDataset":{
         "id":"http://lobid.org/resources/dataset#!",
         "label":"lobid-resources – Der hbz-Verbundkatalog als Linked Open Data"
      },
      "dateCreated":"20210405",
      "sourceOrganization":{
         "id":"http://lobid.org/organisations/DE-294#!",
         "label":"lobid Organisation"
      },
      "resultOf":{
         "instrument":{
            "id":"https://github.com/hbz/lobid-resources",
            "label":"Software lobid-resources",
            "type":[
               "SoftwareApplication"
            ]
         },
         "endTime":"2022-10-08T12:04:14",
         "type":[
            "CreateAction"
         ],
         "object":{
            "inDataset":{
               "id":"https://datahub.io/dataset/hbz_unioncatalog",
               "label":"hbz_unioncatalog"
            },
            "id":"https://lobid.org/hbz01/990209683790206441",
            "label":"hbz-Ressource 990209683790206441 im Exportformat MARC21 XML",
            "type":[
               "DataFeedItem"
            ]
         }
      }
   }
}
```

But actually this information is about the underlying MARC record, it should thus sit in the `object` object like this:

```json
{
   "describedBy":{
      "license":[
         {
            "id":"http://creativecommons.org/publicdomain/zero/1.0",
            "label":"Creative Commons-Lizenz CC0 1.0 Universal"
         }
      ],
      "inDataset":{
         "id":"http://lobid.org/resources/dataset#!",
         "label":"lobid-resources – Der hbz-Verbundkatalog als Linked Open Data"
      },
      "sourceOrganization":{
         "id":"http://lobid.org/organisations/DE-294#!",
         "label":"lobid Organisation"
      },
      "resultOf":{
         "instrument":{
            "id":"https://github.com/hbz/lobid-resources",
            "label":"Software lobid-resources",
            "type":[
               "SoftwareApplication"
            ]
         },
         "endTime":"2022-10-08T12:04:14",
         "type":[
            "CreateAction"
         ],
         "object":{
            "inDataset":{
[...]

PR #1532, opened 2022-11-30
Move date into `describedBy.resultOf.Object` #1470

TobiasNx:
I moved the dates as wished by @acka47 in #1470 one question though why do we format dates differently in `describedBy.resultOf.endTime`

`"endTime" : "0000-00-00T00:00:00"` is dateTime while `dateCreated` and `dateModified`: `YYYYMMDD` without `-`.

### DIFF
--- a/src/main/resources/alma/fix/describedBy.fix
+++ b/src/main/resources/alma/fix/describedBy.fix
@@ -40,13 +40,17 @@ replace_all("describedBy.resultOf.object.dateCreated"," .*","")
 replace_all("describedBy.resultOf.object.dateCreated","c|©|\\s?|,|.|:|;|/|=","")
 replace_all("describedBy.resultOf.object.dateModified","-","")
 replace_all("describedBy.resultOf.object.dateModified"," .*","")
-replace_all("describedBy.resultOf.object.+dateModified","c|©|\\s?|,|.|:|;|/|=","")
+replace_all("describedBy.resultOf.object.dateModified","c|©|\\s?|,|.|:|;|/|=","")
 unless any_match("describedBy.resultOf.object.dateCreated","\\d{8}|\\d{4}")
 	remove_field("describedBy.resultOf.object.dateCreated")
 end
 unless any_match("describedBy.resultOf.object.dateModified","\\d{8}|\\d{4}")
 	remove_field("describedBy.resultOf.object.dateModified")
 end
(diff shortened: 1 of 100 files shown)


acka47:
> why do we format dates differently in `describedBy.resultOf.endTime`
> 
> `"endTime" : "0000-00-00T00:00:00"` is dateTime while `dateCreated` and `dateModified`: `YYYYMMDD` without `-`.

This has historical reasons, I think. `YYYYMMDD` was used in the Aleph data. I agree it would be better to use ISO 8601 everywhere but I think this might break some clients...

TobiasNx:
> > why do we format dates differently in `describedBy.resultOf.endTime`
> > `"endTime" : "0000-00-00T00:00:00"` is dateTime while `dateCreated` and `dateModified`: `YYYYMMDD` without `-`.
> 
> This has historical reasons, I think. `YYYYMMDD` was used in the Aleph data. I agree it would be better to use ISO 8601 everywhere but I think this might break some clients...

@acka47 : If we already break the API by moving the date why not change the date pattern anyway then?

acka47:
>  If we already break the API by moving the date for ALMA FIX why not change the date pattern anyway then?

Of course, you are right. I forgot that we decided to move the date but this PR's title states it loud an clear. So please make it all ISO 8601.

dr0i:
Breaking the pattern `YYYYMMDD` will probably break our deletions API and also maybe something like this:

*Working*:
http://alma.lobid.org/resources/search?q=describedBy.dateCreated:%3E20220110&format=json
http://lobid.org/resources/search?q=publication.startDate%3A%5B*+TO+1500%5D&format=json
http://lobid.org/resources/deletions/search?q=describedBy.deleted%3A%5B20180328+TO+20180405%5D

*Not working*:
http://alma.lobid.org/resources/search?q=describedBy.dateCreated:%3E%222022-01-10%22&format=json
http://lobid.org/resources/deletions/search?q=describedBy.deleted%3A[2018-03-28+TO+2018-04-05]
http://lobid.org/resources/search?q=publication.startDate%3A[*+TO+1500-01-01]&format=json

Maybe this is just a syntax thing, maybe an elasticsearch setting. Idk.

dr0i (approved):

dr0i (changes requested):
Deletions API may  break if the field is moved.
Also API documentation would have to be updated.
Do we have an "API-break" tag? We could/should handle these questions later when switching this to our main product.

Issue #1505, opened 2022-11-03
ALMA-Fix: Use GND enrichment for altLabel


TobiasNx:
At the moment the GND enrichment is inactive. But altLabels are provided there. We could use them to enrich `subject` and `contribution` objects with GND id.

dr0i:
Note that GND enrichment in ALMA is enabled meanwhile.

dr0i:
@TobiasNx : unassigned me here, don't know what you intend me to do.

TobiasNx:
https://github.com/hbz/lobid-resources/pull/1536#issuecomment-1334943188

Comment by @dr0i:
> Just to note:
> This adds data to contribution: altLabel, which is good. It brings often a lot of new data , which also is [indexed](https://github.com/hbz/lobid-resources/blob/master/src/main/resources/index-config.json#L436). Don't want to criticize this, just to make aware that the ES index will be a increased, also ETL time. Let's see how much:
> Landmarks, atm:
> Index size: 75 GB (this varies - updates increases this. This is basedump + 5 daily updates )
> ETL time: ~ 7 h

Not just to `contribution`, but also to `subject`

Two questions still open:

It should be discussed if subjectAltLabel can be deleted.
Nature of content is also based on GND should we enrich them too?

PR #1536, opened 2022-12-01
1505-gndEnrichment

TobiasNx:
First step to reuse GND altLabels 1. contribution 2. subject 

Questions:
After both steps are done. It should be discussed if subjectAltLabel can be deleted.
Nature of content is also based on GND should we enrich them too?

#1505

### DIFF
--- a/src/main/resources/alma/fix/contribution.fix
+++ b/src/main/resources/alma/fix/contribution.fix
@@ -307,7 +307,7 @@ do list (path: "contribution[]", "var": "$i")
     remove_field("$i.agent.dateOfBirthAndDeath")
   end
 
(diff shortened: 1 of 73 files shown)


dr0i:
Just to note:
This adds data to `contribution: altLabel`, which is good. It brings often a lot of new data , which also is [indexed](https://github.com/hbz/lobid-resources/blob/master/src/main/resources/index-config.json#L436). Don't want to criticize this, just to make aware that the ES index will be a increased, also ETL time. Let's see how much:
Landmarks, atm:
Index size: 75 GB (this varies - updates increases this. This is basedump + 5 daily updates )
ETL time: ~ 7 h

dr0i (approved):

Issue #1307, opened 2021-07-30
Add ORCID


dr0i:
@UBmakla asks for ORCID integration.

In GND exists sometimes a sameAs to ORCID, e.g. https://lobid.org/gnd/121546985 .

These ORCIDs exists also in Alma, see e.g. in https://indexes.devel.digibib.net/export/990181637940206441 (accessible only in hbz internal network):
```
<datafield tag="GSI" ind1="7" ind2=" ">
<subfield code="a">0000-0003-3071-9772</subfield>
<subfield code="2">orcid</subfield>
<subfield code="A">GND</subfield>
<subfield code="B">GND-121546985</subfield>
<subfield code="C">024</subfield>
</datafield>
```

If it would be integrated it would appear in https://alma.lobid.org/resources/search?q=121546985&format=json .

TobiasNx:
Okay, that seems to be tricky since the author information resides in 100/110 and 700/710. In there is a GND refrence to the GSI element, which also exist twice:

```
<datafield tag="700" ind1="1" ind2=" ">
<subfield code="a">Tinnemann, Peter</subfield>
<subfield code="d">1967-</subfield>
<subfield code="0">(DE-588)121546985</subfield>
<subfield code="4">oth</subfield>
<subfield code="0">
(uri) https://portal.dnb.de/opac.htm?method=simpleSearch&cqlMode=true&query=idn=121546985
</subfield>
<subfield code="0">(uri) http://viaf.org/viaf/sourceID/DNB|121546985</subfield>
<subfield code="B">GND-121546985</subfield>
</datafield>
...
<datafield tag="GSI" ind1="7" ind2=" ">
<subfield code="a">121546985</subfield>
<subfield code="0">http://d-nb.info/gnd/121546985</subfield>
<subfield code="2">gnd</subfield>
<subfield code="A">GND</subfield>
<subfield code="B">GND-121546985</subfield>
<subfield code="C">024</subfield>
</datafield>
<datafield tag="GSI" ind1="7" ind2=" ">
<subfield code="a">0000-0003-3071-9772</subfield>
<subfield code="2">orcid</subfield>
<subfield code="A">GND</subfield>
<subfield code="B">GND-121546985</subfield>
<subfield code="C">024</subfield>
</datafield>
</record>
```

dr0i:
AFAIK we use `sameAs` only at top level. So it may be easier for both, producer (us!) and consumer to enrich agents by simply adding `orcid` like this:

[...]

PR #1537, opened 2022-12-02
Add orcid as gnd enrichment #1307

TobiasNx:
Added the orcid as wished by @UBmakla and @hagbeck 

Seolves #1307


dr0i:
I added `orcid` to `index-settings.json` to make it searchable.

dr0i (changes requested):
You introduced a new field `orcid`, so we need it added to the context.
 https://en.wikipedia.org/wiki/ORCID#Identifiers may be a good start how to do that, i.a. what kind of data the value could look like.  I am unsure if an `agent` can have _two_ @id .

TobiasNx:
Replaced by #1539

PR #1541, opened 2022-12-06
Introduce deprecatedUri #1535

### LINKED ISSUE #1535: Add deprecatedUri statements for hbzIds
Originates from https://github.com/hbz/lobid-resources/issues/1433.

@acka47 [wrote](https://github.com/hbz/lobid-resources/issues/1433#issuecomment-1326562109):
> we should add a [dnb:deprecatedUri](https://d-nb.info/standards/elementset/dnb#deprecatedUri) statement to each resource that indicates the old URI e.g. for the example resource (in N-Triples): <http://lobid.org/resources/99370849894106441#!> <https://d-nb.info/standards/elementset/dnb#deprecatedUri> "https://test.lobid.org/resources/HT020968842#!" .

@TobiasNx [wrote](https://github.com/hbz/lobid-resources/issues/1433#issuecomment-1327280320):
>Should we add deprecatedUri to all ressources that have HT-Numbers? In my opinion It would be difficult to differentiate between records prior and after the switch to ALMA-Fix. And this would be the easier way.

TobiasNx:
Resolves #1535

### DIFF
--- a/src/main/resources/alma/alma.fix
+++ b/src/main/resources/alma/alma.fix
@@ -28,6 +28,7 @@ retain(
   "containedIn[]",
   "containsExampleOfWork[]",
   "contribution[]",
+  "deprecatedUri",
   "doi[]",
   "extent",
   "edition[]",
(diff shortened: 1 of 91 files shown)


dr0i (commented):
+1

PR #1550, opened 2022-12-15
Add HOL as collection #1174

TobiasNx:
#1174

### DIFF
--- a/src/main/resources/alma/fix/relatedRessourcesAndLinks.fix
+++ b/src/main/resources/alma/fix/relatedRessourcesAndLinks.fix
@@ -400,6 +400,17 @@ do list(path:"084  ", "var":"$i")
 end
 
 
+# Holding Collection
+
+do list(path:"HOL??", "var":"$i")
+  copy_field("$i.8", "inCollection[].$append.id")
+  replace_all("inCollection[].$last.id",".*(\\d{4})$","$1")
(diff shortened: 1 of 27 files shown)


TobiasNx:
@hagbeck is this change what you had in mind? (https://github.com/hbz/lobid-resources/issues/1174?#issuecomment-1346297758) -> every HOL gets an inCollection object.

@acka47 you wanted an example?

dr0i (approved):
To be honest, I don't like this:
as @acka47 noted, this bloats the data with lots of redundant entries - and `inCollection.id` is even indexed in the Elasticssearch, bloating that index too.
... but if it serves the good, go with it.

dr0i (changes requested):
One thing is not correct: the id  needs a hasbang at the end like in: `http://lobid.org/organisations/DE-468#!"`

hagbeck:
> @hagbeck is this change what you had in mind? (#1174) -> every HOL gets an inCollection object.

Yes, that the idea.

> To be honest, I don't like this: as @acka47 noted, this bloats the data with lots of redundant entries - and `inCollection.id` is even indexed in the Elasticssearch, bloating that index too. ... but if it serves the good, go with it.

You're right. This solution bloats the data. But what could be an alternative way?

TobiasNx:
Replaced by https://github.com/hbz/lobid-resources/pull/1546

PR #1564, opened 2022-12-23
Replace Lokaldatensatztypen

TobiasNx:


dr0i (approved):
The test shows that the id of the new item got the callnumber of a an precursor's item - but that may well be ignorable I guess.

Issue #1557, opened 2022-12-19
ALMA-Fix: No map when Karte in 300 but is game


TobiasNx:
```
elsif any_match("300  .a", ".*(Kt.|[kK]arte).*")
  add_field("type[].$append","Map")
elsif any_match("300  .*.a", ".*(Kt.|[kK]arte).*")
```

```
    <datafield tag="300" ind1=" " ind2=" ">
      <subfield code="a">1 Spiel (32 Karten, 1 Karte Spielanleitung) ; in Beh&#228;ltnis 10 x 7 x 1,5 cm</subfield>
    </datafield>
```

At the moment it mappes this as Map but is Spiel, should exclude if `Spiel`

TobiasNx:
Still 133 Karten + Spiel exist: http://test.lobid.org/resources/search?t=Game%2CMap%2CAND&aggregations=type

TobiasNx:
Still 22 but this seems to be okay.

PR #1569, opened 2023-01-03
1557-gameNotMap

TobiasNx:
This version is more encompassing.


dr0i (approved):

dr0i:
> This version is more encompassing.

More encompassing than what?
Also there is no mentioning of the issue in this PR so that the issue in github's standard way to (#1557) so that GH couldn't reference automatically.

PR #1579, opened 2023-01-04
Delete old TODO comments

TobiasNx:
Clean up some old comments.


dr0i:
There is a conflict which I am not bold enough to resolve ad hoc.

PR #1598, opened 2023-01-13
Add MBD with HOL and POR as hasItem #1174

TobiasNx:
Also introduce a new workaround type "NurTitel" for MBD without HOL or POR and "PhysikalischerTitel" for HOL without ITM

### DIFF
--- a/src/main/resources/alma/fix/item.fix
+++ b/src/main/resources/alma/fix/item.fix
@@ -24,7 +24,7 @@ do list(path: "HOL  ", "var": "$i")
     unless in("$i.8", "ITM  .*.H")
       add_field( "hasItem[].$append.test","")
       add_field("hasItem[].$last.label", "lobid Bestandsressource")
-      set_array("hasItem[].$last.type[]", "Item","NoItem")
+      set_array("hasItem[].$last.type[]", "Item","PhysikalischerTitel")
       copy_field("$i.8", "hasItem[].$last.heldBy.id")
       replace_all("hasItem[].$last.heldBy.id",".*(\\d{4})$","$1")
       lookup("hasItem[].$last.heldBy.id", "alma-institution-code-to-isil")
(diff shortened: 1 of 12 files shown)


TobiasNx:
FIY: @blackwinter @UBmakla @hagbeck

src/main/resources/alma/fix/item.fix
```diff
@@ -24,7 +24,7 @@ do list(path: "HOL  ", "var": "$i")
     unless in("$i.8", "ITM  .*.H")
       add_field( "hasItem[].$append.test","")
       add_field("hasItem[].$last.label", "lobid Bestandsressource")
-      set_array("hasItem[].$last.type[]", "Item","NoItem")
+      set_array("hasItem[].$last.type[]", "Item","PhysikalischerTitel")
       copy_field("$i.8", "hasItem[].$last.heldBy.id")
```
dr0i, line 28: I know, this is preliminary, and my pedantry my be annoying  - but I think it would be nonetheless better to use also an English nomen to not mix languages.

src/main/resources/alma/fix/item.fix
```diff
@@ -42,6 +42,29 @@ do list(path: "HOL  ", "var": "$i")
...
+do list(path: "MBD  ", "var": "$i")
+  unless any_match("$i.M","49HBZ_NETWORK")
+    unless in("$i.M", "HOL  .M")
+      unless in("$i.M", "HOL  .*.M")
+        unless in("$i.M", "POR  .M")
+          unless in("$i.M", "POR  .*.M")
+            add_field( "hasItem[].$append.test","")
+            add_field("hasItem[].$last.label", "lobid Bestandsressource")
+            set_array("hasItem[].$last.type[]", "Item","NurTitel")
+            copy_field("$i.i", "hasItem[].$last.heldBy.id")
```
dr0i, line 54: same re language

dr0i (commented):
looks good otherwise

TobiasNx:
I see your point, since this will be workaround placeholders I would keep the german type-values but open an ticket for replacement. Also the translation would be odd: OnlyWork and PhysicalWork / Onyl Title and PhysicalTitle.
Lets move foreward and change it in the next step.

PR #1621, opened 2023-02-13
1478-rpb

TobiasNx:
@acka47 @fsteeg can you help with the construction of the notations for "Systematik der Rheinland-Pfälzischen Bibliographie" and "RPB-Raumsystematik".

How do I model these?


TobiasNx:
And should RPB Spatial also be mapped to `"spatial"` ?

TobiasNx:
We decided to halt this until the Verbund data is enriched with RPB and we have the proper uris.

Issue #1618, opened 2023-01-30
Enable owner facets


dr0i:
We need (not as URIs, but as simple String-IDs ("plain Isil")):
- [x]  owner facets

At at an even later date (June or Juli):
 - [x]  owner facets of sub ISILs ("Institutionszugehörigkeit")
 
In Alma-Fix-lobid-resources we don't have an owner facet atm in the GUI - but the data provides ES-`aggs` so that one can use that to build a `facet`.
~~We need to enable URL-encoding so this depends on https://github.com/hbz/lobid-resources/issues/1521.~~
~~We need then to set `heldBy.sublocation` to have the owner.~~ ~~Depends on https://github.com/hbz/lobid-resources/issues/1602~~
@acka47  noted that it would be nice to be able to filter on cities.

TobiasNx:
@dr0i is this a duplicate to : #1602 

btw. missing url encoding of item-`id` is not blocking this anymore since the id-pattern changed: #https://github.com/hbz/lobid-resources/pull/1647

acka47:
Duplicate of #1602. Closing.

acka47:
Reopening

TobiasNx:
Isn't this query based on the owner facette: http://lobid.org/resources/search?owner=http%3A%2F%2Flobid.org%2Forganisations%2FDE-38%23%21&aggregations=owner

dr0i:
First question is: was this issue properly named? @blackwinter do you need the GUI enabled to show owner facets or di you need ES-`aggs`  to build your own facets- the latter is working like this:
```
curl -XGET  'http://$esServer:9200/resources/_search?q=linux&pretty=true' -d '
{
  "aggs": {
            "top-isil": {
              "terms": {
                "field": "hasItem.heldBy.id",
                "size": 50
              }
            }
      }
}
'
```

blackwinter:
> do you need the GUI enabled to show owner facets or di you need ES-`aggs` to build your own facets

The latter. We don't care about Lobid GUI in this context.

dr0i:
@TobiasNx we need:
>  owner facets of sub ISILs ("Institutionszugehörigkeit")

We already have `hasItem.sublocation.id` (with values like e.g. `https://hbz-network.userservices.exlibrisgroup.com/view/uresolver/49HBZ_FUH
[...]

PR #1661, opened 2023-03-14
Enable hasItem.heldBy.isil as aggs (#1618)

dr0i:
Complements #1640 (as isil was introduced there but the index-config was not configured).

See #1618.


blackwinter (approved):
Thanks!

dr0i:
Will be deployed at Monday.

Issue #1567, opened 2023-01-02
Facette mapping with `300` should exclude types and medium behind `:`


TobiasNx:
Type Mapping with 300 should exclude those that are behind a `:` e.g.:

```
    <datafield tag="300" ind1=" " ind2=" ">
      <subfield code="a">S. [129] - 524 : Ill., graph. Darst, Kt.</subfield>
    </datafield>
```

This has a map as a part of the book but it is no map in it self: http://test.lobid.org/resources/HT004764408#!

dr0i:
Not sure why I am assigned here. If it's about checking wether your commit works it seems not (http://alma.lobid.org/resources/990065341720206441) but then I don't know if I understand the issue coorectly.

TobiasNx:
Seems to work now.

PR #1662, opened 2023-03-14
Only map main extend info #1567

TobiasNx:
Related to #15671


dr0i:
Will be deployed at Monday.

Issue #1553, opened 2022-12-16
ALMA-Fix: Only use content not carrier subjects for `natureOfContent`


TobiasNx:
`natureOfContent` describes the content and not the carrier. But `655` also provides info for `carrier`. Also some old Formschlagwörter are carrier instead of content describing e.g. Film 16mm.

We should get a list of all values and filter out, what carriers we want to exclude.

TobiasNx:
I could introduce this  conditional but I have no example:      `unless any_match("$i.a",".*(Audiovisuelles Material|Bildplatte|CD|CD-ROM|Dia|Diskette|DVD-Audio|DVD-ROM|DVD-Video|Elektronische Publikation|Film|Medienkombination|Mikroform|Musikdruck|Online-Publikation|Schallplatte|Text|Tonbildreihe|Tonkassette|Tonträger|Videokassette).*")`

TobiasNx:
therefore would close if #1663 is deployes

dr0i:
Unassigning me. If https://github.com/hbz/lobid-resources/pull/1663 is deployed and everything is fine @TobiasNx will close this issue.

TobiasNx:
Yes done. Will open another ticket to reuse gnd carrier and old carrirer Formschlagwörter for medium

PR #1663, opened 2023-03-14
Exclude gnd-carrier #1553

TobiasNx:


src/test/resources/alma-fix/990124590390206441.json
```diff
@@ -69,10 +69,6 @@
     "label" : "Deutsch"
   } ],
   "extent" : "1 CD-ROM ; 12 cm + 1 Beih. [6] S.",
-  "natureOfContent" : [ {
-    "label" : "CD-ROM",
-    "id" : "https://d-nb.info/gnd/4139307-7"
-  } ],
   "subject" : [ {
```
dr0i, line 72: How about moving this to `medium` instead of just removing it?

TobiasNx: I agree, but I would do this in context of: #1549

dr0i:
Will be deployed at Monday.

PR #1668, opened 2023-03-22
EtlError300repeatable

TobiasNx:
Instead of only copying a simple field we need to iterate twice over 300 (r) and $a (r).

Error was reported by the monitoring of ALMA ETL.

The error is the result from the first PR in #1567


dr0i (approved):
Two commits are very fine to see what difference the ETL changes make. But you would not only to have to add a test source (xml) but also run the process and add the corresponding json to make this possible.
Otherwise (I think) it's ok.

dr0i:
Be deployed Monday.

PR #1674, opened 2023-03-22
1672-meshInd7

TobiasNx:
Add Medical Subject Headings (MeSH) if they are listed with indicator 2=7


blackwinter (approved):
That was quick :) Thanks!

I'm just wondering, wouldn't it be easier/more performant to include it in the "FAST" block above? Just branch on `$2` for every target vocabulary (`if ... elsif ... end`).

dr0i (approved):

PR #1675, opened 2023-03-22
1485-secondaryPublication

TobiasNx:
Now it also catches endDates if there is a space between the year and the -


dr0i (approved):

dr0i:
Be deployed Monday.

PR #1684, opened 2023-03-27
Add `digiBibSubjectFacet` #1617

### LINKED ISSUE #1617: Make more `labels` working as aggregations to provide facets
Lobid facets use IDs. DigiBib uses labels. Thus we need labels to be typed as `keyword` to  enable ES' aggregations.

We should start with:
- [x]  `language` and`medium`
and later also:
- [x] `subject`

@blackwinter noted that it woul be nice to have ISIL as a field on its own.

@dr0i expects not a huge decline in overall ETL processing time. ES-cluster should be fine doing the extra work.

TobiasNx:
I suggest this version.

Resolves #1617.

### DIFF
--- a/src/main/resources/alma/fix/subjects.fix
+++ b/src/main/resources/alma/fix/subjects.fix
@@ -545,7 +545,7 @@ end
 
 set_array("subjectslabels[]")
 do list(path:"subject[]","var":"$i")
-  if any_match("$i.type[]",".*(Concept|Keyword).*")
+  if any_match("$i.type[]","Concept")
     unless exists("$i.notation")
       copy_field("$i.label","subjectslabels[].$append")
     end
--- a/src/main/resources/alma/fix/subjects.fix
+++ b/src/main/resources/alma/fix/subjects.fix
@@ -73,7 +73,7 @@ set_array("subject[]")
 
 
 do list(path:"600?4|610?4|611?4|630?4|648?4|650?4|651?4|653??", "var":"$i")
-  set_array("subject[].$append.type[]","Keyword")
+  set_array("subject[].$append.type[]","Concept")
   add_field("subject[].$last.source.label","Freie Verschlagwortung")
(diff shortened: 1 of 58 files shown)


src/main/resources/alma/fix/subjects.fix
```diff
@@ -542,3 +542,14 @@ do list(path:"spatial[]", "var":"$i")
   replace_all("$i.focus.geo.lon","^Point\\((.*) .*\\)","$1")
   replace_all("$i.focus.geo.lat","^Point\\(.* (.*)\\)","$1")
 end
+
+set_array("digiBibSubjectFacet[]")
+do list(path:"subject[]","var":"$i")
+  if any_equal("$i.type","Concept")
```
blackwinter, line 548: Shouldn't this be `$i.type[]`?

Looking at test case `990197293880206441.json`, it should also have "Computer science", "Software engineering" and "Operating systems (Computers)" in the new field.

blackwinter (commented):
Fine by me :) Thank you!

But I just noticed that there's also `type=Keyword`. Would you mind adding those as well?

src/main/resources/alma/fix/subjects.fix
TobiasNx: You are right

TobiasNx:
Like this?

blackwinter (approved):
Great, thanks!

Except that local (IZ) fields should not be included at all :( E.g. "FB 03" in `990156027740206441.json`. Will create a new issue for this (see #1687).

TobiasNx:
@acka47 are you okay 1) with the name of this element 2) should we set this element public or exclude this from the public index

dr0i:
Note: we once (when?) agreed to overtly serve the data as it is - with the advantage that useres can see what fieled exists and to implicitly know (well, most of the times) what queries can be build to get their stuff done.

acka47 (changes requested):
See comment

TobiasNx:
We use subjectslabels now

PR #1690, opened 2023-03-30
Exclude DigiBib subject facet from document source (public API).

blackwinter:
In case the bespoke field `digiBibSubjectFacet` (see #1684) should be excluded from the public API, this might be an option at index level ([documentation](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/mapping-source-field.html#include-exclude)).

### DIFF
--- a/src/main/resources/index-config.json
+++ b/src/main/resources/index-config.json
@@ -127,6 +127,11 @@
       },
       "resource" : {
          "dynamic": false,
+         "_source": {
+           "excludes": [
+             "digiBibSubjectFacet"
+           ]
+         },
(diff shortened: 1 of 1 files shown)


dr0i:
+1 . This is a great alternative to to filter the source because we wouldn't have to adjust queries.
Oth we once (when?) agreed to overtly serve the data _as it is_ - with the advantage that useres can see what fieled exists and to implicitly know (well, most of the times) what queries can be build to get their stuff done.

dr0i:
Closing as we will make the field public.
Can be opened if decide later otherwise.

Issue #1574, opened 2023-01-04
Introduce type for "Teil eines Werkes"


TobiasNx:
Limetrans has a mapping for Teil eines Werkes, which translates to the component parts of a publication:

https://github.com/hbz/limetrans/blob/433b0eeffa59dcf31e49ebd662e0826db9d0c02f/src/main/resources/transformation/alma.fix#L727

While we have this for Articles we should introduce this as a general type in my opinion:
Map if `leader` 07 is `a` and/or if `245` has subfield "n" or "p"

acka47:
Today's discussion:

- @TobiasNx and @acka47 have reservations of putting information into `type` that is already expressed by a relation (`isPartOf`)
- Another approach would be to display a facet for `bibliographicLevel`. @blackwinter will investiugate whether this could be an alternative route.

PR #1692, opened 2023-03-30
Introduce bibliographicLevel

TobiasNx:
Instead of a type "PartOfWork" I think we should introduce a new element `bibliographicLevel` derived from [Marc Leader Pos 07](https://www.loc.gov/marc/bibliographic/bdleader.html).

PartOfWork is not a reference to the content  as type provides.

Related to #1574


src/main/resources/alma/fix/mediumAndType.fix
```diff
@@ -37,7 +53,10 @@ paste("@leaderTyp+008","@leaderTyp","008", join_char:"")
 # TODO: Check introx concerning possible mappings of 340 https://github.com/hbz/limetrans/blob/d2dff10e1b5cdf5699239a8f3474f8b652d582a3/src/main/resources/transformation/alma.fix#L629
 
 set_array("@300a_mainExtend")
-do list(path:"300  ", "var":"$i")
+do list(path:"300  ", "var":"$iContains fields that describe the subunit and data that identify the host item.
+
+Subunits may be items, folders, boxes, archival series, subgroups, or subcollections.
+")
```
blackwinter, line 59: There seem to be multiple copy & paste (?) errors. Two more below.

TobiasNx: Thanks

acka47:
We should use controlled values not only strings or at east leave open the possibility to do so in the future. As justed discussed, we will use an object with `label` for now an add IDs for controlled values later.

acka47 (changes requested):
See comment

dr0i:
There are conflicts.

TobiasNx:
I redid the last commit and merged the master into this branch before that.

dr0i:
Will be deployed on Monday.

Issue #1693, opened 2023-03-30
containedIn holds non alma ids


dr0i:
See e.g. http://alma.lobid.org/resources/990221359400206441.json

These old aleph ids are going to be obsolete. They are also problematic for https://github.com/hbz/lobid-resources/issues/1686 . Can they be substituded by `almaMmsId`?

blackwinter:
> These old aleph ids are going to be obsolete.

What do you mean? hbz IDs are here to stay. Only ZDB titles are going to lose them.

```json
"containedIn" : [ {
  "id" : "http://lobid.org/resources/HT012034605#!",
  "label" : "lobid Ressource"
} ]
```

dr0i:
Uh, gosh, wasn't (anymore?) aware of this. It means, in essential, that  it would even be more neccessarye to provide resolvables URIs like http://lobid.org/resources/HT012034605 in the future.

However - why can't we just link to http://alma.lobid.org/resources/990101871580206441 ? It is not only a duplicate of the old deprecatedUri but has even itself `hbzId: "HT012034605"` so the data should be there.
Ah - cheeze - is it true that this data is missing in [990221359400206441](http://alma.lobid.org/resources/990221359400206441.json) ? Looks like: https://alma.lobid.org/marcxml/990221359400206441

This doesn't make much sense.
We _could_ repair this by making a lookup. We could generate a lookup table.


> Only ZDB titles are going to lose them.

What does this mean? The referenced `containedIn` resource has a `zdbId`.

TobiasNx:
> Uh, gosh, wasn't (anymore?) aware of this. It means, in essential, that it would even be more neccessarye to provide resolvables URIs like http://lobid.org/resources/HT012034605 in the future.
We also have to provide resolvables URL because the old URIs are out in the web. Because of that we introduced solutions for:
https://github.com/hbz/lobid-resources/issues/1433
https://github.com/hbz/lobid-resources/issues/1535

> However - why can't we just link to http://alma.lobid.org/resources/990101871580206441 ? It is not only a duplicate of the old deprecatedUri but has even itself `hbzId: "HT012034605"` so the data should be there. Ah - cheeze - is it true that this data is missing in [990221359400206441](http://alma.lobid.org/resources/990221359400206441.json) ? Looks like: https://alma.lobid.org/marcxml/990221359400206441
Nope linked resource data is not there.

> 
> This doesn't make much sense. We _could_ repair this by making a lookup. We could generate a lookup table.
[...]

PR #1695, opened 2023-03-31
ChangeIds_WIP

TobiasNx:
I drafted a PR for another id handling. We would follow the approach from DigiBib (@blackwinter thats your approach right?):

> BTW: In DigiBib, we're using the hbz ID as record ID where available, otherwise the ZDB ID, falling back to the MMS ID as a last resort.

https://github.com/hbz/lobid-resources/issues/1693#issuecomment-1491781235

I also changed the item ids alike.

@acka47 @fsteeg @dr0i what do you think? can we change this. It would be a our last chance before ALMA-lobid replaces Aleph-lobid. what about RPBs reconcile activities?

@hagbeck  @UBmakla would this change be okay with you?

### DIFF
--- a/src/main/resources/alma/fix/identifiers.fix
+++ b/src/main/resources/alma/fix/identifiers.fix
@@ -1,7 +1,5 @@
 copy_field("001","almaMmsId")
 
-paste("id", "~http://lobid.org/resources/", "001", "~#!", join_char: "")
-
 
 # 024 - Other Standard Identifier (R) Subfield: $a (NR) $2 (NR)
 # urn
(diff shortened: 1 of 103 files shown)


dr0i (approved):

blackwinter (approved):
LGTM.

Just to be sure, your index documents keep the MMS ID as their document ID, right? Because hbz IDs aren't guaranteed to be unique.

dr0i (commented):
URIs must be unique. This approach won't do.

blackwinter:
I'm sorry, I shouldn't even have brought it up in the first place. It's no different than with your current Aleph setup, anyway.

fsteeg (approved):

TobiasNx:
Got info from the Verbund:
Linking via hbzId or zdbId will be done at least until ALEPH is turned off.
How linking is handled afterward is not decided:
- it could be that they will replace all links and use ALMA MMS IDs, or
- it could be that they will use ALMA MMS IDs only for new catalogue entries.

Verbund will decided after NZ Phase II is done.

Since the linking was the main motivation for the changes in this PR I would vote for keeping the ALMA-MMS-IDs as part of the lobid ids and for not merging this PR.

TobiasNx:
I withdraw this PR since hbzId is not unique and Verbund is not sure if they continue to keep hbzIds for linking after ALEPH will be shut down.

Issue #1686, opened 2023-03-28
Switch play app testing" Aleph Morph" to "Alma Fix" 


dr0i:
As we are switching from Aleph Morph based ETL to Alma Fix ETL we don't need anymore testing the play app Aleph ETL but instead need testing of Alma Fix.
This is especially useful #1614.

PR #1696, opened 2023-03-31
Add testing of  alma API

dr0i:
See #1686.

### DIFF
--- a/web/app/controllers/resources/LocalIndex.java
+++ b/web/app/controllers/resources/LocalIndex.java
@@ -44,24 +44,25 @@ import play.libs.Json;
  */
 public class LocalIndex {
 
-	private static String TEST_CONFIG =
+	private static final String TEST_CONFIG =
 			"../src/main/resources/index-config.json";
 	private static final String TEST_DATA = "../src/test/resources/jsonld";
 	private static final String TEST_DATA_ITEMS = TEST_DATA + "/items";
 
-    private Node node;
+	private Node node;
 	private Client client;
 	private static final URI CONTEXT_URI =
 			new File("conf/context.jsonld").toURI();
 
-    public LocalIndex() {
-        this(TEST_CONFIG,TEST_DATA,TEST_DATA_ITEMS);
-    }
-    /**
+	public LocalIndex() {
+		this(TEST_CONFIG,TEST_DATA,TEST_DATA_ITEMS);
+	}
+
+	/**
 	 * Create a new local index based on our test set
 	 */
 	public LocalIndex(String testConfig, String testData, String testDataItems) {
-            node = new Node(Settings.builder()
+		node = new Node(Settings.builder()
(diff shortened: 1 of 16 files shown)


web/app/controllers/resources/LocalIndex.java
```diff
@@ -38,27 +38,30 @@
...
 public class LocalIndex {
 
-	private static final String TEST_CONFIG =
+	private static String TEST_CONFIG =
 			"../src/main/resources/index-config.json";
 	private static final String TEST_DATA = "../src/test/resources/jsonld";
 	private static final String TEST_DATA_ITEMS = TEST_DATA + "/items";
 
-	private Node node;
+    private Node node;
```
fsteeg, line 52: Formatting seems off here, and in some other places below.

web/test/tests/almafix/LocalIndexSetup.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Setup for the alma fix search tests. Creates a local ES index with test data.
+ *
+ * @author Fabian Steeg (fsteeg)
+ */
+@SuppressWarnings("javadoc")
+public abstract class LocalIndexSetup {
+
+    private static LocalIndex index;
+    private static String TEST_CONFIG =
```
fsteeg, line 20: This could be `final`, right? If not, as above, it should  be called `testConfig`.

fsteeg (commented):
Commented on some minor issues inline.

And a general question: the `almafix` package duplicates the original tests? Isn't the API basically the same? I would have expected something like a switch for setting where the data comes from (morph/aleph or fix/alma), but the tests remaining the same. I'm probably missing something here.

web/test/tests/almafix/LocalIndexSetup.java
dr0i: right. fixed that.

web/test/tests/IndexIntegrationTest.java
```diff
@@ -98,7 +98,6 @@ private static List<Object[]> queries(Object[][] objects) {
 			String s = (String) testCase[0];
 			Integer hits = (Integer) testCase[1];
 			result.add(new Object[] { new Queries.Builder().q(s), /*->*/ hits });
-			result.add(new Object[] { new Queries.Builder().word(s), /*->*/ hits });
```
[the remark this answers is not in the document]
dr0i: Like the commit describes -it's redundant. It just tests the exact input/output all over again.  Maybe I missed something.

fsteeg (approved):

web/test/tests/IndexIntegrationTest.java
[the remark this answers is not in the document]
fsteeg: Yes, I guess it's alright, we should probably come up with a test in the future that actually shows how these methods return different results.

PR #1713, opened 2023-04-24
Add Alma sublibrary codes for FIZ (libraries without central cataloging).

blackwinter:
Required for [DigiBib](https://github.com/hbz/lobid-resources/milestone/3) (see hbz/lookup-tables#13).


blackwinter:
E.g. [HT005072105](https://alma.lobid.org/resources/990054518430206441.json): `hasItem.heldBy.id=http://lobid.org/organisations/7816#!` which should actually be `DE-Bi12`.

blackwinter:
Should also resolve #1714 (in this particular case) since all items of the title in question will receive actual ISILs.

PR #1732, opened 2023-05-05
Introduce hasFormat/secondaryForm for links to reproductions #1577

TobiasNx:
This illustrates my suggestion from #1577

For links to Reproductions use property `hasFormat`


TobiasNx:
Changed the propertyName to secondaryForm.

PR #1733, opened 2023-05-05
Draft macro for provenance organisations

TobiasNx:
@blackwinter can you identify why the macro is not working?


blackwinter:
The syntax for variables is `$[field]`, not `[$field]` (see also [the documentation](https://github.com/metafacture/metafacture-fix#do-put_macro)).

Issue #1466, opened 2022-10-10
Retrieve the "Buchungsnummer" in hasItem


gregorbg:
In our library (DE-38-459), we assign a unique integer ID (we call it "Mediennummer") to every physical resource. This ID is registered as part of our "Lokaldaten" and it is also physically present in the resource, i.e. we glue a sticker on the inside of the book cover. This sticker contains a barcode and I see a lot of potential here to snap it with a smartphone (or even a proper barcode scanner) and quickly retrieve all relevant information, for example during our annual inventory.

Example: http://lobid.org/resources/HT016717822
Desired information: `JAP/0001835`

The information is currently not present on [lobid-resources](http://lobid.org/resources/HT016717822.json) (Ctrl+F for `JAP/` yields only the signature `JAP/Ls4-2-1CK3`) however it is obtainable through a direct search at http://okeanos-www.hbz-nrw.de/F/.
The Okeanos search doesn't support permalinks afaik, so search for `Nihongo no bunkei to hyōgen 82` to obtain the example item. Clicking on the "Bestand" details for `38/459 K: Ostasiat.Sem./Japan.` yields the following information (as of writing this issue):
![image](https://user-images.githubusercontent.com/6136469/194873952-cc76b41e-2fe4-43f6-95e1-165a08089a3c.png)

The fact that this screenshot contains the desired information leads me to believe that it might(?) be available for lobid to consume. Investigating [the MAB21 export](http://lobid.org/hbz01/HT016717822) for the example item shows field `088` subfields `a`, `b`, `c` and `e`, but for some reason the `d` is missing. Is there a connection here?

TobiasNx:
Since we are preparing the migration to ALMA I checked if this info in the MARC record. This info is not provided: https://alma.lobid.org/marcxml/990184218340206441

blackwinter:
DE-38 is not in Alma yet, they're migrating with Wave 3 next year.

gregorbg:
Here's another example: https://lobid.org/resources/HT018217583
This book is also present in DE-61 which I believe already runs on ALMA.

For reference, here's the screenshot from the Okeanos "Bestandsanzeige": 
![image](https://user-images.githubusercontent.com/6136469/194973323-e85681e7-0412-4ee0-bb21-d9986963f288.png)

Again, I am interested in the information `JAP/0001286` from the row `38/459 K`. This corresponds to the information `131724901` from the row `61 D`.

gregorbg:
[...]

Issue #1614, opened 2023-01-30
Solve differences in analyzing search fields


dr0i:
There are differences in analyzing search fields. DigiBib uses stopwords and in part different analyzers.
We gather some examples and try to find a compromise/solution for both worlds.

blackwinter:
Here are some differences between DigiBib and Lobid index settings:

| Setting | DigiBib | Lobid |
|-|-|-|
| Analyzer | [Custom](https://github.com/hbz/limetrans/blob/a1566c5e257d942e9c9b3f4b710e16b800f26592/src/main/resources/elasticsearch/hbztitle-settings.json#L98)¹ | [Custom](https://github.com/hbz/lobid-resources/blob/4e3cb45c2d5ff36b0f9bbf31011945e9fa3c1fb4/src/main/resources/index-config.json#L7) & [Standard](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/analysis-standard-analyzer.html) |
| Tokenizer | [Hyphen](https://github.com/jprante/elasticsearch-plugin-bundle/blob/master/src/docs/asciidoc/hyphen.adoc) | [Standard](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/analysis-standard-tokenizer.html) |
| Search analyzer | [w/o Hyphen](https://github.com/hbz/limetrans/blob/a1566c5e257d942e9c9b3f4b710e16b800f26592/src/main/resources/elasticsearch/hbztitle-settings.json#L114) | n/a |
| "All fields" search | [`cql.allIndexes`](https://github.com/hbz/limetrans/blob/a1566c5e257d942e9c9b3f4b710e16b800f26592/src/main/resources/elasticsearch/hbztitle-mapping.json#L68) (via [`copy_to`](https://www.elastic.co/guide/en/elasticsearch/reference/2.2/copy-to.html)) | [`_all`](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/mapping-all-field.html)² |

¹ Additional filters in DigiBib analyzer:

- [german_stop_words](https://github.com/hbz/limetrans/blob/a1566c5e257d942e9c9b3f4b710e16b800f26592/src/main/resources/elasticsearch/hbztitle-settings.json#L39)
- [keyword_repeat](https://www.elastic.co/guide/en/elasticsearch/reference/2.2/analysis-keyword-repeat-tokenfilter.html)
[...]

PR #1737, opened 2023-05-05
Add serialNumber to ElasticSearch index

gregorbg:
Context: We glue barcodes into each and every book (so-called "Mediennummern") which are published in Lobid-Resources as `hasItem.serialNumber` (cf. https://github.com/hbz/lobid-resources/issues/1466). Perspectively, it would be cool to offer information services where people can zap the barcode with their smartphone camera and receive information on the book they're holding in their hands.

Bonus: Can also be very useful for our annual inventory, but digitising that is a long way :innocent: 

**DISCLAIMER**: I was not able to find instructions on how to run the Integration tests (folder structure looks like Scala Build Tools, but the general project is build using Maven??) so the tests that I provided are written blind using some `serialNumber` and `callNumber` that I picked via CTRL+F'ing the test resources folder.


dr0i:
Looks good but for a test case, see f519dc7 .

If you want to build and test the scala web app have a look at https://github.com/hbz/lobid-resources#build. You need `activator` or even better `sbt` (and do then `sbt clean test`).

However, I don't understand why/if you actually need to _search_ a `serialNumber` - you describe the issue as if it's just about _reading_ it. So we could configure the index to  `"index": "false"` . That would ease the Elasticsearch.

gregorbg:
Thanks for fixing the test case and pointing me in the right direction for running the integration tests. I was aware of the section in your README but I didn't realise that this part is what actually runs the integration tests. I will look into it when I have more free time on my hands this week.

> However, I don't understand why/if you actually need to search a serialNumber - you describe the issue as if it's just about reading it. So we could configure the index to "index": "false" . That would ease the Elasticsearch.

I literally just copied the index definition from `callNumber` straight above, without giving it too much thought. Your argument is perfectly valid and we can set `"index": false`, but consequentially we then should consider doing the same for `callNumber`. What do you think?

dr0i:
You are right. I've done this with [96eff74](https://github.com/hbz/lobid-resources/pull/1737/commits/96eff742501e3af50ebfc9dcde7a8f0715b11ec9) and force pushed that.

src/main/resources/alma/index-config.json
```diff
@@ -382,7 +382,12 @@
                      "index" : "false"
                   },
                   "callNumber" : {
-                     "type" : "keyword"
+                     "type" : "keyword",
+                     "index" : "false"
```
blackwinter, line 386: Sorry, but I have to "veto" this change. In DigiBib, we need to be able to search for call numbers.

There might be a different solution coming up in the context of #1614, but for now it's indispensable.

blackwinter:
> left `hasItem.callNumber` as it was. Introduced an API test to ensure that bevaviour.

Great, thanks!

dr0i:
Will be deployed next Monday.

PR #1738, opened 2023-05-05
Fix `describedBy.resultOf` URLs to point to `/marcxml/` instead of `hbz01`

gregorbg:
As of writing this issue, the live data for Alma still points to `lobid.org/hbz01/${almaMmsId}` which throws an error.
Example at https://lobid.org/resources/99371014448006441?format=json:
```json
"id": "https://lobid.org/hbz01/99371014448006441",
"type": [
  "DataFeedItem"
],
"label": "hbz-Ressource 99371014448006441 im Exportformat MARC21 XML",
```

I updated all test resources accordingly. The script at https://github.com/hbz/lobid-resources/blob/master/src/test/resources/alma-fix/alephFilesFromIdsOfAlmaTestFiles/grepAlmaTestIdsAndLookupAlephJsonAndXml.sh#L10 likely also needs to be updated to point to `aleph.lobid.org` but I was unsure how to proceed there because (a) I don't know when this script is being used and (b) CURL likes throwing a tantrum when the SSL certificates are not properly configured (and you claimed that you won't maintain SSL for aleph.lobid.org)

### DIFF
--- a/src/main/resources/alma/fix/describedBy.fix
+++ b/src/main/resources/alma/fix/describedBy.fix
@@ -29,7 +29,7 @@ set_array("describedBy.resultOf.instrument.type[]", "SoftwareApplication")
 add_field("describedBy.resultOf.instrument.label","Software lobid-resources")
 
 copy_field("almaMmsId","describedBy.resultOf.object.id")
-prepend("describedBy.resultOf.object.id","https://lobid.org/hbz01/")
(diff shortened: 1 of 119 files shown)


dr0i:
Good catch - thx! We totally forgot about that...
Will be deployed next Monday.

dr0i:
Also: removed old script , see https://github.com/hbz/lobid-resources/commit/03536d9e66a7988675a20b599c92178912be714e.

PR #1747, opened 2023-05-11
Use Sigil Mapping for describedBy Organisations

dr0i:
See https://github.com/hbz/lobid-resources/pull/1742.


TobiasNx:
@gregorbg have a look at my solution. should suit your needs.

dr0i (commented):
An example showing the solving of the DE-38-459 problem would be good.

gregorbg:
:star_struck: Thank you for providing such an extensive solution by bringing back the entire Siegel map!

To be honest, there was no real problem to begin with. I don't even need the links or data under `describedBy`. I simply noticed while glancing over the data that the URL is corrupt, but I have no specific test case that would cover any "use case", because there is none :sweat_smile:

dr0i:
Will be deployed next Monday.

PR #1755, opened 2023-05-12
Exclude IZ-activated CZ resources from union catalogue

### LINKED ISSUE #1754: Exclude IZ-activated CZ resources from union catalogue
The filter condition from #1677 needs to be refined: IZs can activate resources from the CZ, which then automatically generate records in the NZ for technical reasons; these are different from centrally managed CZ resources and should be excluded from the union catalogue.

Example: [99370694196806441](https://lobid.org/resources/99370694196806441)

blackwinter:
Resolves #1754.


src/test/resources/alma-fix/99370694196806441.json
```diff
@@ -0,0 +1,242 @@
+{
```
dr0i, line 1: is it necessary to add this resource or does it structurally just reflects what other test resources already do? If you just added this resource because you mentioned it in the issue I suggest to change the example in the issue and remove this test file here to keep the files as small as reasonable.

dr0i (changes requested):

src/test/resources/alma-fix/99370694196806441.json
blackwinter: I appreciate the rationale and, no, it's technically not necessary in the context of this pull request (there are other test files illustrating the effect). But it's the only test record, AFAICS, that shows the effect of the new NWBib filter (3a84d88) in combination with this change.

dr0i: good then :+1:

dr0i (approved):

dr0i:
Will be deployed next Monday.

blackwinter:
Thanks!

PR #1767, opened 2023-05-17
1761-gndAndMacro

TobiasNx:
Events and meetings have a specific structure for the label. In order to construct them from the Marc Subfields we had to introduce a separate transformation for events.

Also I have put all macros in a separate fix. @blackwinter can you have a look at this. Since I deleted all do once for the single macros but have put it in the maps do once in alma.fix

The macro.fix allows for the usage of the macros in subject.fix and contribution.fix.

### DIFF
--- a/src/main/resources/alma/alma.fix
+++ b/src/main/resources/alma/alma.fix
@@ -1,8 +1,9 @@
 add_field("@context","http://lobid.org/resources/context.jsonld")
 
 # Add maps
-do once("maps")
+do once("mapsAndMacros")
   include ("./fix/maps.fix")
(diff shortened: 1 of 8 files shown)


blackwinter (approved):
> Also I have put all macros in a separate fix.

Yes, I would recommend this approach.

> Since I deleted all do once for the single macros but have put it in the maps do once in alma.fix

You could simplify it even further by merging `mapsAndMacros` and `vars` into a single `setup` block.

dr0i (approved):

dr0i:
Will be deployed next Monday.

PR #1774, opened 2023-05-30
Add missing zdbId

TobiasNx:


blackwinter (approved):
I suppose it's okay, so :+1:

(For future reference, `035$$a` with prefix `(DE-599)` is the [EKI](https://verbundwiki.gbv.de/display/VZG/Erstkatalogisierungs-ID).)

But while you're at it, you might also want to map `016$7#$a` with `$$2=DE-600`.

TobiasNx:
Added the requested transformation which indeed catched a missing zdbId. I also added an dnbId and an object for dnb resources since they are provided in 016 too.

TobiasNx:
@acka47 is it okay to map the dnbId of the record too?

src/test/resources/alma-fix/990103899140206441.json
```diff
@@ -3,6 +3,7 @@
   "almaMmsId" : "990103899140206441",
   "oclcNumber" : [ "1367753905" ],
   "zdbId" : "590016-5",
+  "dnbId" : "01335941X",
```
dr0i, line 6: This field `dnbId` has no entry in the `context.jsonld`.
Can you provide one ?

TobiasNx: I added it now

dr0i (approved):

TobiasNx:
@acka47 approved this

dr0i:
Will be deployed next Monday.

Issue #1606, opened 2023-01-17
Set up basic reconciliation service for lobid-resources


fsteeg:
Originating from [RPB-50](https://jira.hbz-nrw.de/browse/RPB-50), we decided to set up a basic reconciliation endpoint for lobid-resources. Here is the data set for that use case: [rpb-36sm.json.zip](https://github.com/hbz/lobid-resources/files/10434548/rpb-36sm.json.zip), see [instructions on importing the data](https://github.com/hbz/rpb/blob/rpb-50/README.md#reconciliation) into OpenRefine. The goal here is to reconcile that data with lobid-resources (using values from multiple fields in the source data), and to output mappings from the source data ID `#00 ` to `hbzId` and `zdbId` in lobid-resources.

PR #1777, opened 2023-05-31
Reconciliation service

fsteeg:
Current state of the reconciliation service for #1606, as it has been used for [RPB-50](https://jira.hbz-nrw.de/browse/RPB-50) (in https://github.com/hbz/rpb/pull/26).

For merging this as a first version of the reconciliation service, we should at least:

- [ ] Remove rpbId filter
- [ ] Add documentation


TobiasNx:
@fsteeg In context of http://openrefine.labs.lobid.org/project?project=2114080802133 the lobid resources reconciliation was used, now it seems to be non existing.

fsteeg:
Now running again at https://test.lobid.org/resources/reconcile

PR #1782, opened 2023-06-02
Nwbib616-missingFullTextOnline

TobiasNx:
Are there more variations of the nbn-resolving link that i missed?


dr0i (changes requested):
Can you point me to the transformed filed in json? I cannot find it adhoc.

TobiasNx:
added the change in the test

dr0i (approved):

PR #1794, opened 2023-06-15
Skip all deleted records

TobiasNx:
Deleted records have DEL in their data. We should exclude them from the index.

### DIFF
--- a/src/main/resources/alma/alma.fix
+++ b/src/main/resources/alma/alma.fix
@@ -5,6 +5,11 @@ do once("setup")
   put_var("member", "-")
 end
 
+# Skip all deleted records
+if any_equal("DEL??.a", "Y")
+  reject()
+end
+
 # drop all local fields #1687
 de.hbz.lobid.helper.DropLocal()
 


dr0i:
Also, see https://github.com/hbz/lobid-resources/issues/1795 - we need to create an index to query which resources were deleted.

dr0i (approved):

blackwinter (approved):
Unfortunately, I can't recall off the top of my head why we're also checking `leader/5` in [Limetrans](https://github.com/hbz/limetrans/blob/9e4f700f0fc593edda79fa280469380ba0978018/src/main/resources/transformation/alma.fix#L1585). Otherwise, this should do it.

Also, just for the record: You're only skipping deleted records here; they're not actually deleted from the index during an update.

dr0i:
> Also, just for the record: You're only skipping deleted records here; they're not actually deleted from the index during an update.

Hm, thx for the intervention - it's possible to do that. Just as a note. If demand arises.

PR #1798, opened 2023-06-19
1797-zdbIdSpace

TobiasNx:
Sometimes zdbId can have values like `2598795 1`. The whitespace should be substituted with an `-`.


dr0i (changes requested):
What is the purpose of this PR -  order is afaik nothing we support. Order is arbitrary.
And why intoducing a new test file?

TobiasNx:
Have a look at the commits not just the diff, the new test fixes the problem with the space.

dr0i:
Ok, but I beg for a simple description - it's easier for you who has understanding of the problem.
Leaving the order: why change it?

TobiasNx:
The reason is it enables to group the `replace_all` fixes for `zdbId` together in the last commit. It is a preparation for the last commit.

Issue #1671, opened 2023-03-22
Implement MARC `546$$a,b` (Language Note)


blackwinter:
[HT018856429](https://alma.lobid.org/resources/990210273970206441) and [HT019472708](https://alma.lobid.org/resources/990219775100206441) should include the "Language Note" ([MARC 546$$a,b](https://www.loc.gov/marc/bibliographic/bd546.html)).

"Language note":

```xml
<datafield tag="546" ind1=" " ind2=" ">
  <subfield code="a">Beiträge teilweise deutsch, teilweise englisch</subfield>
</datafield>
```

"Information code or alphabet":

```xml
<datafield tag="546" ind1=" " ind2=" ">
  <subfield code="b">In chinesischer Schrift</subfield>
</datafield>
```

acka47:
From https://www.loc.gov/bibframe/mtbf/ it looks like Bibframe uses a general note field and adds a type information to the different notes. This apporach would mean an API break so we can not follow this path. Thus, we should either use this approach for newly added notes or just define a property `lv:langNote` or similar in lobid vocab.

blackwinter:
`langNote` is available now.

PR #1800, opened 2023-06-20
Add langNote mapping #1671

TobiasNx:
Introduced a mapping for language note from Marc, see #1671

Also added the new property `langNote`. I have to introduce this to the vocab too.

### DIFF
--- a/src/main/resources/alma/fix/otherFields.fix
+++ b/src/main/resources/alma/fix/otherFields.fix
@@ -162,6 +162,10 @@ end
 
 set_array("langNote[]")
 do list(path: "546??", "var": "$i")
-  copy_field("$i.a", "langNote[].$append")
-  copy_field("$i.b", "langNote[].$last")
+  set_array("$i.@langNote")
+  copy_field("$i.a", "$i.@langNote.$append")
+  copy_field("$i.b", "$i.@langNote.$append")
+  replace_all("$i.@langNote.*", "[.;]$", "")
+  join_field("$i.@langNote", "; ")
+  move_field("$i.@langNote", "langNote[].$append")
 end
--- a/src/main/resources/alma/fix/otherFields.fix
+++ b/src/main/resources/alma/fix/otherFields.fix
@@ -159,3 +159,9 @@ do list(path: "542??", "var": "$i")
   copy_field("$i.u", "license[].$last.label")
   replace_all("license[].$last.label", "^(http://|https://)(.*)$","$2")
 end
+
+set_array("langNote[]")
+do list(path: "546??", "var": "$i")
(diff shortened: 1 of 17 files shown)


dr0i (approved):

src/main/resources/alma/fix/otherFields.fix
```diff
@@ -159,3 +159,9 @@ do list(path: "542??", "var": "$i")
   copy_field("$i.u", "license[].$last.label")
   replace_all("license[].$last.label", "^(http://|https://)(.*)$","$2")
 end
+
+set_array("langNote[]")
+do list(path: "546??", "var": "$i")
+  copy_field("$i.a", "langNote[].$append")
+  copy_field("$i.b", "langNote[].$last")
```
blackwinter, line 166: So this is _either_ `$$a` or `$$b`? I'm not sure if this is the way to go, but I don't have a counterexample either.

According to the [MARC standard](https://www.loc.gov/marc/bibliographic/bd546.html), there _can_ be both at the same time. So maybe something like this?

```suggestion
  set_array("$i.@langNote")
  copy_field("$i.a", "$i.@langNote.$append")
  copy_field("$i.b", "$i.@langNote.$append")
  replace_all("$i.@langNote.*", "[.;]$", "")
  join_field("$i.@langNote", "; ")
  move_field("$i.@langNote", "langNote[].$append")
```

blackwinter (approved):
This would be the minimum viable solution for now, but I'm not sure if it's sufficient long-term.

src/main/resources/alma/fix/otherFields.fix
blackwinter, line 166: This would also overwrite a previous `546` in case the field is repeated, so I guess I'd have to veto this approach.

blackwinter (changes requested):

src/main/resources/alma/fix/otherFields.fix
TobiasNx: This approach was not correct. I have to fix this.
But what do you mean by `546`

blackwinter: > But what do you mean by `546`

MARC field `546`. If there are multiple occurrences, the last `546$$b` wins; previous ones will be overwritten.

TobiasNx: Thanks. Should be fixed now.

web/conf/context.jsonld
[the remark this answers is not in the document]
TobiasNx: na, I will delete them.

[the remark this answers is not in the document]
TobiasNx: or we can leave them. They are created new every time. And will not commit next time

PR #1805, opened 2023-06-22
Introduce lookup for deprecatedUri based on mapping from zdbId

TobiasNx:
In ALMA all zdb records loose their hbzId (aka HT Nummer). To keep at least their old record Id as deprecatedUri . I created a mapping file from the last aleph transformation and map the zdbId to the old hbzId. With this I create the deprecatedUri element.

The lookup holds 2.000.000 + key value pairs. I hope this is not to heavy on the performance.


TobiasNx:
I created a mapping file from my tsv and added the old hbzId to the zdb titles.

Perhaps also we could reuse the file from Jens/Verbund.

blackwinter:
> added the old hbzId to the zdb titles.

Great, thanks!

fsteeg (approved):

PR #1810, opened 2023-06-28
UI support for RPB-only resources

### LINKED ISSUE #1809: Update UI for records with rpbId only
In preparation for adding RPB records to the productive lobid-resources index, the UI needs to be updated to work with records that contain neither `hbzId`, `almaMmsId`, nor `zdbId`, only an `rpbId`.

We already have these in the staging index, see:
http://stage.lobid.org/resources/search?q=NOT+_exists_:(hbzId+OR+almaMmsId+OR+zdbId)

This is a UI-only issue, but adding the records to the productive index might cause similar issues for other clients, e.g. DigiBib (do you use stage for testing @blackwinter? If so, this should be fine).

fsteeg:
Will resolve #1809.

Deployed to stage:
http://stage.lobid.org/resources/search?q=NOT+_exists_:(hbzId+OR+almaMmsId+OR+zdbId)

### DIFF
--- a/web/app/controllers/resources/Application.java
+++ b/web/app/controllers/resources/Application.java
@@ -471,15 +471,23 @@ public class Application extends Controller {
 		if (cachedResult != null)
 			return cachedResult;
 		Promise<Result> promise = Promise.promise(() -> {
-			JsonNode result = getJsonForId(id);
-			if (result == null) {
+			JsonNode result = new Search.Builder().build().getResource(id).getResult();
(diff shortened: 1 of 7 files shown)


dr0i:
Will be automatically deployed at lobid.org/resources at Monday.

fsteeg:
> Will be automatically deployed at lobid.org/resources at Monday.

Can I deploy it manually and index RPB data into the production index?

dr0i:
Yes, you can do that.

Issue #1814, opened 2023-07-07
Missing rpbId / inCollection


fsteeg:
For http://lobid.org/resources/990186699340206441

According to LBZ, it's catalogued like other RPB titles.

fsteeg:
Looks good for http://lobid.org/resources/990186699340206441 now.

But now there is one record in RPB which should not be `inCollection` RPB:

http://lobid.org/resources/990174158630206441

PR #1815, opened 2023-07-07
Fix inCollection for RPB #1814

TobiasNx:
The mapping was wrong. Fixed it.


fsteeg (approved):
Oh great, that was quick, thanks!

PR #1826, opened 2023-07-24
1824 missing cz journals

TobiasNx:
@hagbeck : Have a look here.

The mapping with concern to the `"POR"` Element seems to create some confusion. Any suggestion how to proceed here? It seems that the same record can have `POR $$f` BOOK and JOURNAL.

https://github.com/hbz/lobid-resources/blob/7ba4a070546d5775330925cb88d56c0814ce5c88/src/test/resources/alma-fix/990210093550206441.xml#L245-L246

https://github.com/hbz/lobid-resources/blob/7ba4a070546d5775330925cb88d56c0814ce5c88/src/test/resources/alma-fix/990210093550206441.xml#L226-L227

### DIFF
--- a/src/main/resources/alma/fix/mediumAndType.fix
+++ b/src/main/resources/alma/fix/mediumAndType.fix
@@ -563,8 +563,12 @@ elsif any_match("natureOfContent[].*.label", ".*(Zeitschrift|Werkzeitschrift|eri
 elsif any_equal("906", "JOURNAL")
   add_field("type[].$append","Periodical")
 else 
+  set_array("@PORtype")
   do list(path: "POR  ","var":"$i")
-    if any_equal("$i.f","JOURNAL")
+    copy_field("$i.f","@PORtype.$append")
+  end
+  if any_equal("@PORtype","JOURNAL")
+    unless any_equal("@PORtype","BOOK")
       add_field("type[].$append","Periodical")
     end
   end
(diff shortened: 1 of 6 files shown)


hagbeck:
Hm, it looks like, that this depends on the selected portfolio by the institution. But in this case the record is managed by the NZ P2E_JOB and I would expect the same value for all portfolios.

TobiasNx:
I searched deep in the issues: It seems that Perodical is not the same as Journal but also includes Series that are part of the ZDB At least this was like this.I don't know. Perhaps we could open a specific issue and keep it like that. Or I could exclude Periodical if it is stated as Book:
https://github.com/hbz/lobid-resources/issues/121

Also this is at least at the moment a minor problem: http://lobid.org/resources/search?q=type%3ASeries+AND+type%3APeriodical

src/main/resources/alma/fix/mediumAndType.fix
```diff
@@ -558,8 +558,20 @@ if any_match("@leaderTyp+008", "(Continuing Resources)(.{21})p.*")  # Pos21
   add_field("type[].$append","Periodical")
 elsif any_match("006", "^[s](.{3})p.*") # Pos00+04
   add_field("type[].$append","Periodical")
-elsif any_match("natureOfContent[].*.label", ".*(Zeitschrift|Werkzeitschrift).*")
+elsif any_match("natureOfContent[].*.label", ".*(Zeitschrift|Werkzeitschrift|eriodical).*")
```
blackwinter, line 561: Was this intentional or is the "P" missing?

blackwinter (approved):

src/main/resources/alma/fix/mediumAndType.fix
TobiasNx: This is intentional to catch Periodical and periodical.

blackwinter: Maybe make it explicit then? `[Pp]`

TobiasNx: thanks for the suggestion went with `(?i).*(Zeitschrift|Periodical).*`

Issue #1830, opened 2023-07-25
GUI differences and Metadata differences ALMA/ALEPH


TobiasNx:
Der Spiegel is an example that shows differences between old ALEPH GUI and ALMA GUI, also it highlights some of the metadata problems:

http://lobid.org/resources/990054089950206441
http://aleph.lobid.org/resources/HT001387709

Also Spiegel is an big example, that breaks the testing. It is 3,5 MB in contrast to some kb.

- [ ] #2009
- [ ] #1991 
- [x] Has multiple publication-objects [IMprovement in ALMA](https://github.com/hbz/lobid-resources/pull/1834)
- [x] << and >> in alternativeTitel ALEPH // better in ALMA, therefore okay.
- [ ] #2012
- [x] ~strange containsExampleOfWork in ALEPH~
- [x] Augstein Begründer: Beitragender in ALEPH und Other in ALMA // seems to be no issue anymore.
- [ ]       "label": "lobid-resources – Der hbz-Verbundkatalog als Linked Open Data" in ALMA schein falsch
- [x] ISSN in ALMA not uniq
- [ ] #2011
- [x] #1989
- [x] issn in related has still - ALMA
- [x] culturegraph links are missing in ALMA, should be able to reconstruct in ALMA now that HT number are available too
- [ ] #2014
- [x] ~Formschlagwörter have been dropped in ALEPH, but this seems to be due to changes in the RSWK rules~
- [x] DDC is missing in ALMA
- [x] #1994
- [ ] #1992 GUI shows subordinates and containedIn links, in ALEPH but not in ALMA GUI.
 - [ ] not sure if holdings are missing in GUI

TobiasNx:
The medium is print in ALEPH and microformat in ALMA.

TobiasNx:
Closing this ticket, since almost all issues are solved or follow up tickets are provided.

PR #1834, opened 2023-07-26
1830-publicationImprovement

TobiasNx:
Solves an issue from #1830 and solves #1886 

In Aleph it seems, that there is only one object for PrimaryPublicationEvent. Because of the different sources of the publicationInfo we now have multiple and all have the info outside `260` and `264`.

I suggest to keep the multiple `publication`-objects, but prefer `264` over `260`.
And only add the info for `frequency`,`publicationHistory`,`note` and the mapping of indian pub dates, to the first `object` of the `publication`-Array.
The first ideally is the current (indicator1: 3) publication info, if not the intermediate  (indicator1: 2) and if not the first/undefined (indicator1: #).

This PR does:
- introduce a macro for publication
- prefer 264 over 260 publication info
- only add publicationHistory and frequency to the first/current publication info
- add some fallbacks for publication date and location

### DIFF
--- a/src/main/resources/alma/fix/macros.fix
+++ b/src/main/resources/alma/fix/macros.fix
@@ -329,7 +329,6 @@ end
 do put_macro("publication")
   do list(path:"$[field]", "var":"$i")
     add_field("publication[].$append.test","")
-    copy_field("362??.a","publication[].$last.publicationHistory")
     do list(path: "$i.c", "var":"$j")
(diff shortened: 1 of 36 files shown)


TobiasNx:
The Spiegel Example can help to further improve the publication mapping.

TobiasNx:
Adrian gave +1 of board.

dr0i (approved):

Issue #1325, opened 2021-12-16
Support for multi-language alphabets/symbols in all applicable fields


gregorbg:
This is a feature request. I would like to be able to retrieve information from any book or bibliographical resource that is written in a foreign language in the original script / alphabet.

Context: When cataloguing Japanese books, we usually enter the Japanese details in a romanized transcript, for example `さむらい` becomes `samurai`. At the same time, we _also_ fill so-called "parallel fields" in Aleph, that contain the same information in the original script.

The "parallel fields" are derived by replacing the first digit N of a MAB2 field number by the N-th letter of the Latin alphabet (where 0=Z) like so:
- **1**00 --> A00 (`A` is letter number `1` of the Latin alphabet)
- **3**31 --> C31 (`C` is letter number `3`)
- **4**19 --> D19 (you get the idea :smile:)
- **0**89 --> Z89 (minor exception, see above)

I know that other linguistic scripts can be implemented the same way (think Arabic, Korean, Sanskrit, etc...) and in fact all of the "parallel fields" need to contain sub-field `6` (the red thing in Aleph, not sure whether "sub-field" is the correct terminology) which specifies a four-letter code like `Jpan` for Japanese script.

You can see an example at http://lobid.org/hbz01/HT020582812. We generally provide "parallel fields" with original script information wherever possible, in >95% of the cases this affects `A00`, `A04`, ..., `B00`, `B04`, ..., `C31`, `C59`, `D03`, `D19` and perhaps `Z89` or `D51` and `D55`. But there are even more fields which do in theory support parallel values, like `370` for alternative titles etc.

I currently don't have a preference for whether to provide one JSON per script or all scripts in one big JSON. Just being able to retrieve the information would be great :partying_face:

acka47:
As we definitely won't have the time to implement this with Aleph/MAB-based data before the migration to Alma, we need to know how this looks in MARC21. To easily do this by myself, #1316 needs to be resolved.

dr0i:
In MARC it looks like:
```
...
<datafield tag="880" ind1="1" ind2=" ">
<subfield code="6">100-02/Jpan</subfield>
<subfield code="a">黒田, 清揚</subfield>
</datafield>
<datafield tag="245" ind1="1" ind2="0">
<subfield code="6">880-01</subfield>
<subfield code="a">Sora kara mita nihon</subfield>
<subfield code="c">Kuroda, Kiyoaki ; Sasaki, Atsurô kyôcho</subfield>
[...]

PR #1854, opened 2023-08-02
1325 alternate graphic representation

TobiasNx:
For arraysOfObjects I thought about something like this.
For the other elements, we have to use another approach.

Limetrans only maps alternat graphic representation for:
100 -> array of objects `contribution`
110 -> array of objects `contribution`
245 -> string `title`
246 -> array of strings `alternativeTitle`
250 -> array of strings `edition`
264 ->  array of objects `publication` 
490 -> object  `isPartOf.hasSuperordinate`
500 -> array of strings `note`
610 -> array of objects `subject` 
650 -> array of objects `subject` 
700 -> array of objects `contribution`
710 -> array of objects `contribution`

While `contribution`, `isPartOf.hasSuperordinate`, `subject` should be relatively easy to implement since I could add 
I am not sure how we model the Alternate Graphic Representation for `title`, `alternativeTitle`, `edition` and `note` since they are "simple" string elements or array of strings.

Also publication probably needs another approach.


src/test/resources/alma-fix/990367761810206441.json
```diff
@@ -0,0 +1,162 @@
...
+  "responsibilityStatement" : [ "Kuroda, Kiyoaki ; Sasaki, Atsurô kyôcho" ],
+  "contribution" : [ {
+    "agent" : {
+      "label" : "Kuroda, Kiyoaki",
+      "type" : [ "Person" ],
+      "alternateGraphicRepresentation" : [ {
+        "script" : "Japanisch (Kanji, Hiragana und Katakana)"
+      }, {
+        "label" : "黒田, 清揚"
+      } ]
```
blackwinter, line 138: I don't think this is the correct data structure. It should probably be:

```json
"alternateGraphicRepresentation" : [ {
  "script" : "Japanisch (Kanji, Hiragana und Katakana)",
  "label" : "黒田, 清揚"
} ]
```

src/main/resources/alma/alma.fix
```diff
@@ -15,6 +15,11 @@ de.hbz.lobid.helper.DropLocal()
 
 add_field("@context","http://lobid.org/resources/context.jsonld")
 
+do list(path:"880??","var":"$i")
+  copy_field("$i.6","$i.linkageTest")
+  replace_all("$i.linkageTest","\\d{3}(-\\d{2}).*","880$1")
+end
```
blackwinter, line 21: What is this supposed to do?

blackwinter (changes requested):

src/main/resources/alma/alma.fix
TobiasNx: This prepares the comparison in the linkage field and allows it to be angostic about the field name:

https://github.com/hbz/lobid-resources/blob/57f3f7916ae7c4f3930f2efd9672600339ea986f/src/main/resources/alma/fix/macros.fix#L4-L9

TobiasNx: ignore the outcomment line it can be deleted.

blackwinter: I see, thanks.

src/test/resources/alma-fix/990367761810206441.json
```diff
@@ -4,6 +4,21 @@
...
   "deprecatedUri" : "http://lobid.org/resources/HT020582812#!",
   "title" : "Sora kara mita nihon",
+  "alternateGraphicRepresentation" : [ {
+    "$append" : {
+      "script" : {
+        "id" : "https://unicode.org/iso15924/iso15924.txt#Jpan",
+        "label" : "iso15924.txt#Jpan"
+      }
+    },
+    "$last" : {
```
TobiasNx, line 14: @blackwinter can you identify why the resultsare not written in the same object and the script is separated?

blackwinter: Not off the top of my head, no. I can look at it later.

blackwinter: Also note that the object keys are wrong: `$append` and `$last`.

TobiasNx: The object keys seems to be the key problem. I narrowed down the problem: https://github.com/hbz/lobid-resources/pull/1854/commits/d510f56925aa32cbe25e6651e32a67e3d46bb1f5

It is not the merging of additional elements.
Also it is not the conditional or the list-bind:
```
do list(path:"alternateGraphicRepresentation[]","var":"$AGR")
              if in ("$880.@script.id","$AGR.script.id")
```

blackwinter: You need to put `set_array("alternateGraphicRepresentation[]")` after any `vacuum()`, otherwise the empty array will be removed before it's used.

TobiasNx: thanks, i knew that using vacuum in the middle of the transformation was a bad idea. Will fix this.

TobiasNx:
@blackwinter and @acka47  can you have a look.
This is my suggestion. It is not very consistently modeled, since all array of objects and objects have their own `"alternateGraphicRepresentation"`-Element and all others are dumped in "`alternateGraphicRepresentation`".

I would prefer dumping all alternative graphical representation into the top level element: `"alternateGraphicRepresentation"` But this would make it difficult to match values that belong together of the array of objects as hinted by @gregorbg : https://github.com/hbz/lobid-resources/issues/1325#issuecomment-1343788666. Only the paths would be the same as I handled the array of strings e.g. `"note"`

src/main/resources/alma/fix/otherFields.fix
```diff
@@ -95,6 +95,36 @@ end
...
+        if in ("$i.linkageTest","$880.linkageTest")
+          if in ("$880.@script.id","alternateGraphicRepresentation[].*.script.id")   
+            do list(path:"alternateGraphicRepresentation[]","var":"$AGR")
+              if in ("$880.@script.id","$AGR.script.id")
+                if exists("$AGR.note[]")     
+                  copy_field("$880.a","$AGR.note[].$append")     
+                else
+                  set_array("$AGR.note[]")
+                  copy_field("$880.a","$AGR.note[].$append")
+                end
```
blackwinter, line 112:
```suggestion
                unless exists("$AGR.note[]")
                  set_array("$AGR.note[]")
                end
                copy_field("$880.a","$AGR.note[].$append")
```

blackwinter (changes requested):
Looks good so far, apart from minor issues.

I still need to check suitability for use in DigiBib before I can approve.

blackwinter:
Of course, but that might not be sufficient to solve #1325 as well.

blackwinter (approved):
We could work with both top-level and nested `alternateGraphicRepresentation` in DigiBib. Top-level (only) would be simpler implementation-wise, but nested would probably be more future-proof.

Just so we're on the same page, you're talking about this data structure for top-level, right?

```json
{
  "alternateGraphicRepresentation": [
    {   
      "contribution": [
        {
          "agent": {
            "label": "佐々木, 敦朗"
          }
        }
      ],
      "edition": [
        "初版" 
      ],
      "publication": [
        {
          "location": [
            "大阪"
          ],
          "publishedBy": [
            "保育社"
          ]
        }
      ],
      "script": {
        "id": "https://unicode.org/iso15924/iso15924.txt#Jpan",
        "label": "Japanisch (Kanji, Hiragana und Katakana)"
      },
      "title": "空から見た日本"
    }
  ] 
}
```

TobiasNx:
@gregorbg : this would help you with the alternative script versions  would this version be good for you?

gregorbg:
Thanks for your effort to implement this feature! Even though the name `alternateGraphicRepresentation` doesn't exactly roll off the tip of your tongue, it does cover everything that I need.

I would prefer the top-level version because then I only have to change one type definition -- the project is using a strongly typed language so the current nested approach requires me to update _every_ type schema which feels a bit overpowered. However, I can also live with the nested version if there are stronger arguments in favor of it.

acka47:
I discussed this yesterday with @TobiasNx and we are npot yet satisfied with the data structure exemplified in https://github.com/hbz/lobid-resources/pull/1854#pullrequestreview-1569558540.

For certain, the `script` should not just be listed on the same level as the fields in the language, so we need something like this:

```json
{
   "alternateGraphicRepresentation":{
      "script":{
         "id":"https://unicode.org/iso15924/iso15924.txt#Jpan",
         "label":"Japanisch (Kanji, Hiragana und Katakana)"
      },
      "fields":[
         {
            "contribution":[
               {
                  "agent":{
                     "label":"佐々木, 敦朗"
                  }
               }
            ]
         },
         {
            "edition":[
               "初版"
            ]
         },
         {
            "publication":[
               {
                  "location":[
                     "大阪"
                  ],
                  "publishedBy":[
                     "保育社"
                  ]
               }
            ]
         },
         {
            "title":"空から見た日本"
         }
      ]
   }
}
``` 

Furthermore, I am are unsure whether there is a need for an array with `alternateGraphicRepresentation` in case there is more than one script covered in the data. I don't know whether this is possible or desired in MARC. If yes, it should look something like this:

```json
{
   "alternateGraphicRepresentation":[
      {
         "script":{
            "id":"https://unicode.org/iso15924/iso15924.txt#Jpan",
            "label":"Japanisch (Kanji, Hiragana und Katakana)"
         },
         "fields":[
            {
               "contribution":[
                  {
                     "agent":{
                        "label":"佐々木, 敦朗"
                     }
                  }
               ]
            },
            {
               "edition":[
                  "初版"
               ]
            },
            {
               "publication":[
                  {
                     "location":[
                        "大阪"
                     ],
                     "publishedBy":[
                        "保育社"
                     ]
                  }
               ]
            },
            {
               "title":"空から見た日本"
            }
         ]
      }
   ]
}
```

What do you thin, @gregorbg & @blackwinter ?

acka47:
Generally, this whole addition messes up the – currently rather clean – semantic structure of lobid-resources as we add information about strings – their alternative representations – on a level where we usually have information about the bibliographic resource. A semantically clean approach would look something like this – but we can not implement this as it means lots of API breaks (and also we shouldn't add that kind of complexity for a small fraction of lobid records even if it wasn't an API break):

### Example (**just an illustration no suggestion for implementation**)

```json
{
      "@context":"http://lobid.org/resources/context.jsonld",
      "title":[
         {
            "value":"Sora kara mita nihon",
            "alternateGraphicRepresentation":{
               "script":{
                  "id":"https://unicode.org/iso15924/iso15924.txt#Jpan",
                  "label":"Japanisch (Kanji, Hiragana und Katakana)"
               },
               "value":"空から見た日本"
            }
         }
      ],
      "edition":[
         {
            "value":"Shohan",
            "alternateGraphicRepresentation":{
               "script":{
                  "id":"https://unicode.org/iso15924/iso15924.txt#Jpan",
                  "label":"Japanisch (Kanji, Hiragana und Katakana)"
               },
               "value":"初版"
            }
         }
      ],
      "publication":[
         {
            "type":[
               "PublicationEvent"
            ],
            "location":[
               {
                  "value":"Ôsaka",
                  "alternateGraphicRepresentation":{
                     "script":{
                        "id":"https://unicode.org/iso15924/iso15924.txt#Jpan",
                        "label":"Japanisch (Kanji, Hiragana und Katakana)"
                     },
                     "value":"大阪"
                  }
               }
            ],
            "publishedBy":[
               {
                  "value":"Hoikusha",
                  "alternateGraphicRepresentation":{
                     "script":{
                        "id":"https://unicode.org/iso15924/iso15924.txt#Jpan",
                        "label":"Japanisch (Kanji, Hiragana und Katakana)"
                     },
                     "value":"保育社"
                  }
               }
            ]
         }
      ],
      "contribution":[
         {
            "agent":{
               "label":{
                  "value":"Kuroda, Kiyoaki",
                  "alternateGraphicRepresentation":{
                     "script":{
                        "id":"https://unicode.org/iso15924/iso15924.txt#Jpan",
                        "label":"Japanisch (Kanji, Hiragana und Katakana)"
                     },
                     "value":"佐々木, 敦朗"
                  }
               },
               "type":[
                  "Person"
               ]
            },
            "role":{
               "id":"http://id.loc.gov/vocabulary/relators/aut",
               "label":"Autor/in"
            },
            "type":[
               "Contribution"
            ]
         },
         {
            "agent":{
               "label":{
                  "value":"Sasaki, Atsurô"
               },
               "type":[
                  "Person"
               ]
            },
            "role":{
               "id":"http://id.loc.gov/vocabulary/relators/aut",
               "label":"Autor/in"
            },
            "type":[
               "Contribution"
            ]
         }
      ],
      "id":"http://lobid.org/resources/990367761810206441#!"
   }
```

However, @TobiasNx just had a good idea how to balance this: by naming the field something like `describedInAlternativeGraphics`. Thinking about this, another option might be to put it in the `describedBy` object that contains data about the bibliographic description.

### Example for the `describedBy` approach

```json
{
   "id":"http://lobid.org/resources/990367761810206441#!",
   "describedBy":{
      "id":"http://lobid.org/resources/990367761810206441",
      "type":[
         "BibliographicDescription"
      ],
      "inDataset":{
         "id":"http://lobid.org/resources/dataset#!",
         "label":"lobid-resources – Der hbz-Verbundkatalog als Linked Open Data"
      },
      "alternateGraphicRepresentation":{
         "script":{
            "id":"https://unicode.org/iso15924/iso15924.txt#Jpan",
            "label":"Japanisch (Kanji, Hiragana und Katakana)"
         },
         "record":{
            "contribution":[
               {
                  "agent":{
                     "label":"佐々木, 敦朗"
                  }
               }
            ],
            "edition":[
               "初版"
            ],
            "publication":[
               {
                  "location":[
                     "大阪"
                  ],
                  "publishedBy":[
                     "保育社"
                  ]
               }
            ],
            "title":"空から見た日本"
         }
      }
   }
}
```

In summary, I see three options to go forward:
1. the first one in https://github.com/hbz/lobid-resources/pull/1854#issuecomment-1674282421 (@TobiasNx let me know that an array is not needed as there aren't multiple alternatvie graphic representations for one record)
2. like 1.) but with another name for the field
3. putting the alternative description in `describedBy` as in the example above.

blackwinter:
Either option would work for us (except the API breaking one), but I don't think I particularly like the `describedBy` approach.

> 1. @TobiasNx let me know that an array is not needed as there aren't multiple alternatvie graphic representations for one record

This doesn't seem to be correct; there _can_ be [multiple scripts](https://www.loc.gov/marc/bibliographic/ecbdmulti.html#modelb) in a record.

So, if you're not going to associate the alternate graphic representations with their respective source fields anyway, why not simply treat them as entities of their own? Each `880` then becomes an element in the `alternateGraphicRepresentation` array (see also the implementation in [Limetrans](https://github.com/hbz/limetrans/blob/b84c20790d2bf07ed16f9c5fc947d116fb6d5f75/src/main/resources/transformation/alma.fix#L1868)):

```json
{     
  "alternateGraphicRepresentation": [
    { 
      "script": {
        "id": "https://unicode.org/iso15924/iso15924.txt#Jpan",
        "label": "Japanisch (Kanji, Hiragana und Katakana)"
      }, 
      "field": "contribution.agent",
      "value": "佐々木, 敦朗"
    },
    { 
      "script": {
        "id": "https://unicode.org/iso15924/iso15924.txt#Jpan",
        "label": "Japanisch (Kanji, Hiragana und Katakana)"
      }, 
      "field": "edition",
      "value": "初版"
    },
    { 
      "script": {
        "id": "https://unicode.org/iso15924/iso15924.txt#Jpan",
        "label": "Japanisch (Kanji, Hiragana und Katakana)"
      }, 
      "field": "publication.location",
      "value": "大阪"
    },
    { 
      "script": {
        "id": "https://unicode.org/iso15924/iso15924.txt#Jpan",
        "label": "Japanisch (Kanji, Hiragana und Katakana)"
      }, 
      "field": "publication.publishedBy",
      "value": "保育社"
    },
    { 
      "script": {
        "id": "https://unicode.org/iso15924/iso15924.txt#Jpan",
        "label": "Japanisch (Kanji, Hiragana und Katakana)"
      },
      "field": "title",
      "value": "空から見た日本"
    }
  ]
}
```

TobiasNx:
@acka47 
an array is needed. @gregorbg gave the example: 
https://github.com/hbz/lobid-resources/issues/1325#issuecomment-1342254898

acka47:
> @acka47
> an array is needed. @gregorbg gave the example:
https://github.com/hbz/lobid-resources/issues/1325#issuecomment-1342254898

Thanks for the pointer. I misread your message and have now edited my comment at https://github.com/hbz/lobid-resources/pull/1854#issuecomment-1674363427

acka47:
@gregorbg Would the solution in https://github.com/hbz/lobid-resources/pull/1854#issuecomment-1674524346 be ok for your use case?

acka47:
We just discussed in the meeting whether the "alternate" scripts are always those that are actually written on the book. If yes, we might/should as well call the field `originalGraphicRepresentation`. @gregorbg can you enlighten us here?

gregorbg:
Sorry for the delay, I was in Asia...

> We just discussed in the meeting whether the "alternate" scripts are always those that are actually written on the book. If yes, we might/should as well call the field `originalGraphicRepresentation`. @gregorbg can you enlighten us here?

The answer here is a clear "Jain". The basic rule is that we only record what is written on the book, so by definition you will always have one main script (and the Latin alphabet transliteration). However, there are cases like Traditional Chinese and Simplified Chinese, where some book might have Traditional Chinese characters on its cover (especially for literature from Hong Kong) which is then accurately recorded in the Katalogisat. At the same time, for searchability and discoverability, the Simplified Chinese characters may be recorded as well. So it is not 100% guaranteed that there is always just "at most one" alternate representation

(Slightly more convoluted example: Japanese Kanji can in theory have multiple readings. In written texts for vocabulary, it is always clear from the context which word has which reading. But for place names and human names (the latter especially relevant for Verfasserangaben) the reading can be ambiguous and is sometimes even deliberately used in non-trivial and non-standard readings to set the name of the child apart. So, it is good practice for Japanese authors to add the transliteration, the original Japanese Kanji writing as appears on the book as `Jpan` transliteration, and then **also** another `Kana` transliteration that clarifies the reading in Japanese)

gregorbg:
> @gregorbg Would the solution in [#1854 (comment)](https://github.com/hbz/lobid-resources/pull/1854#issuecomment-1674524346) be ok for your use case?

Definitely works, as in "I can figure out which information belongs to which field". However, it feels "too technical". One thing that I appreciate about the Lobid API is that its technical implementation is very sophisticated on one hand, but the end result is also very intuitive on the other hand. You can almost read the JSON representation without any technical knowledge, barring the basic sense for digital Katalogisatformate that would also be required by any librarian creating MARC entries.

In that sense, introducing `jq`-esque notations that require dedicated parsing just "feel wrong". I can't be any more precise than that, and I see that there are semantic arguments in favor of this solution. But the `describedBy` approach just feels more natural because it references the same object in the same schema just in a different script (which makes more sense as to what an alternate representation actually is).

What about going with an `alternateGraphicRepresentations` approach similar to https://github.com/hbz/lobid-resources/pull/1854#issuecomment-1674363427, but putting it on top level rather than under `describedBy`?

acka47:
> What about going with an `alternateGraphicRepresentations` approach similar to https://github.com/hbz/lobid-resources/pull/1854#issuecomment-1674363427, but putting it on top level rather than under `describedBy`? 

This would be ok for me.

acka47:
LGTM. +1 We will have to think about whether we will actually add `alternateGraphicRepresentation`, `script` and `record` to the JSON-LD context or whether this is "just in the JSON".

acka47 (approved):

src/main/resources/alma/fix/relatedRessourcesAndLinks.fix
```diff
@@ -130,9 +131,9 @@ do list(path: "830??", "var": "$i")
       copy_field("$j", "isPartOf[].$last.hasSuperordinate[].$last.label.$append")
     end
     join_field("isPartOf[].$last.hasSuperordinate[].$last.label", " / ")
-    # call_macro("alternateGraphicRepresationSimple",targetField:"isPartOf[].$last.hasSuperordinate[].$last",variable:"$i")
+    # call_macro("alternateGraphicRepresationArrayOfObjects",targetField:"isPartOf[].$last.hasSuperordinate[].$last",variable:"$i")
```
blackwinter, line 134: This outcommented line looks wrong. Better delete it?

src/main/resources/alma/fix/macros.fix
```diff
@@ -1,20 +1,56 @@
 # alternateGraphicRepresation
-do put_macro("alternateGraphicRepresationSimple")
-  if exists("$[variable].6")
```
blackwinter: If `variable` isn't used anymore, it should be deleted from all call sites.

src/main/resources/alma/fix/relatedRessourcesAndLinks.fix
```diff
@@ -130,9 +131,9 @@ do list(path: "830??", "var": "$i")
...
     do list(path: "4901?", "var": "$j")
-      call_macro("alternateGraphicRepresationSimple",targetField:"isPartOf[].$last.hasSuperordinate[].$last",variable:"$j")
+      call_macro("alternateGraphicRepresationIsPartOf",targetField:"isPartOf[].$last.hasSuperordinate[].$append",variable:"$j")
```
blackwinter, line 136: But `variable` is required here.

src/main/resources/alma/fix/macros.fix
```diff
@@ -1,20 +1,56 @@
...
+                unless exists("$AGR.record.$[targetArray]")
+                  set_array("$AGR.record.$[targetArray]")
+                end
+                copy_field("$880.a","$AGR.record.$[targetField].label")
+              end
+            end
+          else    
+            copy_field("$880.@script.id","alternateGraphicRepresentation[].$append.script.id")
+            copy_field("$880.@script.label","alternateGraphicRepresentation[].$last.script.label")        
+            set_array("alternateGraphicRepresentation[].$last.record.$[targetField]")
```
blackwinter, line 20: Shouldn't this use `targetArray`?

blackwinter (approved):
`IndexIntegrationTest` fails. And transformation test coverage looks a bit weak. But overall :+1:

gregorbg:
> Any good? @blackwinter @gregorbg @acka47

Wonderful! :heart_eyes: Seems like the object(s) beneath `alternateGraphicRepresentation.record` all follow the same JSON schema as the top-level Resource object, which is all I could ever ask for!

Nitpick: The entry we're using as an example throughout this issue (https://lobid.org/resources/990367761810206441.json) has an alternate representation for `responsibilityStatement` in the MARC-XML

```xml
<datafield tag="880" ind1="1" ind2="0">
  <subfield code="6">245-01/Jpan</subfield>
  <subfield code="c">黒田清揚 ;佐々木敦朗 共著</subfield>
  <subfield code="9">F:359</subfield>
  <subfield code="M">49HBZ_BRIDGE_UBK</subfield>
</datafield>
```

TobiasNx:
@gregorbg I have added `responsibilityStatement`.

dr0i (approved):

dr0i:
Will be deployed next Monday.

PR #1877, opened 2023-08-28
1876 fix rpb spatial

TobiasNx:
I am not sure what the notation is for the values with 99_o as prefix:

`99_o13207076`
`o13207076`
or `13207076`

Are there more Variants of the Spatial notations that I dont know and that are not covered by our testsfiles `01` and `99_o13207076`?


acka47:
> I am not sure what the notation is for the values with 99_o as prefix:
>
> `99_o13207076`
> `o13207076`
> or `13207076`

You can look them up in https://github.com/hbz/lobid-vocabs/blob/master/rpb/rpb-spatial.ttl 
We have not defined a notation in the vocab, so I would omiz it for now and ask LBZ whether one is needed.

> Are there more Variants of the Spatial notations that I dont know and that are not covered by our testsfiles `01` and `99_o13207076`?

I don't know. We will see an might fix those later I guess.

acka47 (approved):
+1

PR #1884, opened 2023-09-06
1882 rpb notation

TobiasNx:
Related to #1882 not sure if necessary.


TobiasNx:
closing since entry error of lbz

PR #1890, opened 2023-09-13
Update fix version

TobiasNx:
I will open a ticket and have a look at the change in extend. Seems to be due to a repeated element.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
       run: |
         git clone https://github.com/metafacture/metafacture-fix.git
         cd metafacture-fix
-        git reset --hard 0.5.1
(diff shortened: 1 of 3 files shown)


blackwinter (approved):

dr0i:
I feel a bit uncomfortable because of the change of `extent`. I mean, shouldn't be a bug or something? Opened the issue #1894.

TobiasNx:
> I feel a bit uncomfortable because of the change of `extent`. I mean, shouldn't be a bug or something? Opened the issue #1894.

This is due to the fact, that one of the subfields of `300` is repeated. The transformation that I borrowed from introx did not expect this to be repeated. The new destructive behaviour of add_field, copy_field and move_field is based on this.

Will have a look at this yes.

PR #1892, opened 2023-09-14
Add index fields for DigiBib.

### LINKED ISSUE #1614: Solve differences in analyzing search fields

blackwinter:
Resolves #1614.

### DIFF
--- a/src/main/resources/alma/index-config.json
+++ b/src/main/resources/alma/index-config.json
@@ -1,891 +1,891 @@
 {
-   "settings" : {
-      "index" : {
-         "number_of_shards" : 3,
-         "analysis" : {
-            "analyzer" : {
-               "german_analyzer" : {
-                  "filter" : [
-                     "lowercase",
-                     "snowball"
-                  ],
-                  "tokenizer" : "standard",
-                  "type" : "custom"
-               },
-               "german_analyzer_unstemmed" : {
-                  "filter" : [
-                     "lowercase",
-                     "german_normalization"
-                  ],
-                  "tokenizer" : "standard",
-                  "type" : "custom"
-               },
-               "id_analyzer" : {
-                  "tokenizer" : "keyword",
-                  "filter" : [
-                     "lowercase"
-                  ]
-               },
-               "ngram_analyzer" : {
-                  "filter" : [
-                     "lowercase",
-                     "stop",
-                     "ngram",
-                     "snowball"
-                  ],
-                  "tokenizer" : "letter",
-                  "type" : "custom"
-               },
-               "hyphen_analyzer": {
-                  "tokenizer": "standard",
-                  "char_filter": "remove-hyphens"
-               }
-            },
-            "filter" : {
-               "ngram" : {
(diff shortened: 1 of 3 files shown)


dr0i (commented):
re 7d1b06765cba5ebd61a752dd39b0c67dc85ca41b : Hm. We didn't talked about it much. This _could_ be done combining querying these fields with an OR query.

blackwinter:
> re [7d1b067](https://github.com/hbz/lobid-resources/commit/7d1b06765cba5ebd61a752dd39b0c67dc85ca41b) : This also _could_ be done combining querying these fields with an OR query.

Not for our min/max facets, AFAIK.

src/main/resources/alma/index-config.json
```diff
@@ -39,6 +39,37 @@
...
+              "lowercase",
+              "digibib_german_stop_words",
+              "keyword_repeat",
+              "icu_normalizer",
+              "icu_folding",
+              "hyphen",
+              "digibib_string_normalizer",
+              "snowball",
+              "digibib_standardnumber",
+              "unique"
```
dr0i, line 56: Here I don't know if we should take this "digibib_standardnumber" to the  already present `isbn` field.

blackwinter: AFAICS, this shouldn't make a difference since you already provide ISBN-10 and ISBN-13 variants through the transformation.

dr0i (approved):
I don't have any objections. I have a feeling that it _could_ be that some of these changes would also made sense for lobid-resources in general, so that we wouldn't all these special fields here.

src/main/resources/alma/index-config.json
dr0i: ok then

web/test/tests/IndexIntegrationTest.java
```diff
@@ -98,7 +98,15 @@ public static Collection<Object[]> data() {
...
-			{ "q.publisher:Quedenfeld*", /*->*/ 2 }
+			{ "q.publisher:Quedenfeld*", /*->*/ 2 },
+			{ "q.subject:Düsseldorf", /*->*/ 1 },
+			{ "q.subject:Duesseldorf", /*->*/ 1 },
+			{ "q.subject:Dusseldorf", /*->*/ 1 },
+			{ "q.subject:Düsseldorfer", /*->*/ 1 },
+			{ "q.subject:Westfalen", /*->*/ 10 },
+			{ "q.subject:Westfälen", /*->*/ 10 },
+			{ "q.subject:Lithuania", /*->*/ 1 },
+			{ "q.subject:Baukem", /*->*/ 1 }
```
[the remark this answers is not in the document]
blackwinter: No, `q.subject` is a query field, `subjectslabels` is a facet field.

dr0i:
Will be deployed next Monday.

Issue #1893, opened 2023-09-15
Separate lookup-tables for test and prod and use gitsubmodule for prod


TobiasNx:
As already done for DDC.

dr0i:
Separation is done with #1902. 
We don't use anymore cronjob to update th lookupt tables but gitsubmodule (as proposed by @blackwinter ).
It's already deployed.
We shall await some live tests with real data before closing this issue.

dr0i:
Seems to be good as there are no errors in the logs.
Closing.

PR #1902, opened 2023-09-22
Add lookup tables as git submodules

dr0i:
Do [edited:] `git config --local submodule.recurse true` once in your checked out repo. Any further `git pull` will automatically update the submodules.

See  #1893.


blackwinter (approved):
> Do "git config --global submodule.recurse true" once in your checked out repo.

I don't think that's such a good recommendation. I'd suggest `--local` instead (which is the default). Or just use `git pull --recurse-submodules` explicitly.

dr0i:
`--local` is good. (I like to use the "established" commands  (here `git pull`) to work with the new setting - not only because of cronjobs and scripts (which will work without adapting them).

TobiasNx:
` src/test/resources/deweyLabels.tsv` should also be moved into the maps folder.

TobiasNx:
Also we do not need a specific `src/test/resources/alma/maps` folder a folder `src/test/resources/maps` would be good enough.

dr0i:
We have `src/main/resources/alma/maps/` and thus `src/test/resources/alma/maps/` is symmetric - which is a very nice property and should be kept.

Re dewey: done that
Re further discussion of where to put test resources: that's not the scope of this PR/issue so  we should open a new one if needed.

dr0i:
I've deleted now `src/main/resources/alma/maps/generatedAlmaSuppressedLocations.tsv`.
Re documentation: I don't understand why we should put a readme into the `maps` directory when these lookup tables are not read from this directory.

TobiasNx:
> I've deleted now `src/main/resources/alma/maps/generatedAlmaSuppressedLocations.tsv`. Re documentation: I don't understand why we should put a readme into the `maps` directory when these lookup tables are not read from this directory.

okay, then go for it.
I just thought that it helps to find missing map files when somebody does not find it in the folder. But then go on. +1

TobiasNx (approved):

README.textile
```diff
@@ -102,6 +102,10 @@ If you've added and commited everything, check again if all is ok:
 
 This should result in *BUILD SUCCESS*. Push your changes. You're done :)
 
+h2. Tables as gitsubmodules
+The lookup tables are provided through gitsubmodules. To make a @git pull@ also
```
TobiasNx, line 106: Oh, perhaps say "Some lookup tables" since not all of them are provided via submodule.

dr0i: Resoved with [8338ca9](https://github.com/hbz/lobid-resources/pull/1902/commits/8338ca93ec74af0db9c9ebc687ff32aea5b04856).

dr0i:
Unfortunatley it seem not possible to have the submodules not in a detached state (may depend on the git version) but you have to do ` git submodule update --recursive --remote` after `git pull`. Will update all scripts and README.

PR #1909, opened 2023-09-26
LogErrorInAlternativeGraphicRepresentation

TobiasNx:
Reported in logs

The examples seem non sensical  since they provide latin  alternativeGraphicRepresentation for a latin script record / ublicaation.

But they are an example for the fixed error.


dr0i (approved):

dr0i:
Will be deployed next Monday.

PR #1906, opened 2023-09-25
1888-opacLink

TobiasNx:
This PR does the following:
- Refactoring the mapping `heldBy` so that the ISIL of the IZ is provided, that changes the order of `heldBy.isil` and `heldBy.id`
- Adding a macro for adding a `opacLink` as `sameAs`
- Adding mapping files (tsv) for `opacLink` tusing a submodule for prod and a local copy for test
- Adding `sameAs` for all mapped `hasItem`-objects

@acka47 do you have a better property than `opacLink`?

-----
Outdated:
~First draft of opac link enrichment mechanism~

~@blackwinter: my first design runs into an error, could you have a look at it:~

<details>
<summary>~Error-message~</summary>
```
[main] ERROR org.lobid.resources.AlmaMarc21XmlToLobidJsonMetafixTest - Errored when transforming 
org.metafacture.metafix.FixProcessException: Error while executing Fix expression (at file:/home/tobias/git/lobid-resources/src/main/resources/alma/fix/macros.fix, line 267): split_field("$i.@opacLink","{hbzid}")
        at org.metafacture.metafix.RecordTransformer.tryRun(RecordTransformer.java:249)
        at org.metafacture.metafix.RecordTransformer.lambda$transform$2(RecordTransformer.java:125)
        at java.lang.Iterable.forEach(Iterable.java:75)
        at org.metafacture.metafix.RecordTransformer.transform(RecordTransformer.java:124)
        at org.metafacture.metafix.RecordTransformer.lambda$null$14(RecordTransformer.java:183)
        at org.metafacture.metafix.RecordTransformer.lambda$null$22(RecordTransformer.java:226)
        at org.metafacture.metafix.RecordTransformer.lambda$null$1(RecordTransformer.java:125)
        at org.metafacture.metafix.RecordTransformer.tryRun(RecordTransformer.java:237)
        at org.metafacture.metafix.RecordTransformer.lambda$transform$2(RecordTransformer.java:125)
        at java.lang.Iterable.forEach(Iterable.java:75)
        at org.metafacture.metafix.RecordTransformer.transform(RecordTransformer.java:124)
        at org.metafacture.metafix.RecordTransformer.transform(RecordTransformer.java:116)
        at org.metafacture.metafix.FixMethod$10.apply(FixMethod.java:155)
        at org.metafacture.metafix.RecordTransformer.lambda$null$20(RecordTransformer.java:206)
[...]

Issue #1888, opened 2023-09-08
Provide links into local catalogues.


blackwinter:
It would be nice if individual holdings (`hasItem`) could include links into the respective library's catalogue (OPAC). The Lobid web app already [does this](https://github.com/hbz/lobid-resources/blob/f7fc49c6dae16f92b60b12189db81f2b8506387b/web/app/controllers/resources/Lobid.java#L393) based on [Link Templates](https://github.com/hbz/link-templates). If the URLs would be provided via the API, any downstream consumers could reuse them.

It needs to support both hbz IDs and ZDB IDs; possibly also MMS IDs as fallback.

Originally requested by FEx Daten (W1/W2).

TobiasNx:
@blackwinter you mentioned that we should prefer hbzId over zdbId.
Are you sure that local catalogues still have hbzId in their metadata for ZDB records, since they got dropped in the NZ?
But we still add the old hbzId to the pre-migration ZDB record via a mapping.

TobiasNx:
zdbId and hbzId mappings seem to work.

How do we proceed with ISBNs, Alma MMS Ids and ISSNs?
1) There is no map for opac links with ALMA MMS ID.
2) There are multiple ISBN and ISSN possible. Should we somehow join them before adding them?

blackwinter:
> There is no map for opac links with ALMA MMS ID.

Needs to be added in lookup-tables. But you can already reference it even if it doesn't exist yet.

> There are multiple ISBN and ISSN possible. Should we somehow join them before adding them?

Depends on the individual catalogue. Maybe just use the first one for starters?

Issue #1868, opened 2023-08-18
Cache lookup table


dr0i:
Results from https://github.com/hbz/lobid-resources/issues/1862:
We should not lookup `isil2opac_hbzid.json` but put it into the filesystem and make it hot-loadable (automatically reload this conf if it changes). We should then crontab (or webhook) to update `isil2opac_hbzid.json`.

Details:
`web/conf/resources.conf:isil2opac_hbzid = "https://raw.githubusercontent.com/hbz/link-templates/master/isil2opac_hbzid.json"`

Note: this is used in other projects , e.g. nwbib. Grep githubusercontent. Should be treated accordingly.

TobiasNx:
@dr0i I am not sure if we need to do this, since we are enriching the source data with this information, see: #1888

dr0i:
Ah , good to hear! The old lookup of websites every time was really not good.
Closing.

PR #1911, opened 2023-09-26
Cache table "isil2opac_hbzid"

dr0i:
Don't make a URL lookup to get the concordance every time an item is shown. Load thie concordance from a gitsubmodule table and cache it as HashMap.

Follows #1893.

### DIFF
--- a/README.textile
+++ b/README.textile
@@ -105,6 +105,7 @@ This should result in *BUILD SUCCESS*. Push your changes. You're done :)
 h2. Tables as gitsubmodules
 Some lookup tables are provided through gitsubmodules. To make a @git pull@ also
 update these tables you can e.g. do @git config --local submodule.recurse true@ once.
+If you update an already cloned repo you need to do @git submodule update --init --recursive@ first.
 
 h2. Propagate the context.json to lobid-resources-web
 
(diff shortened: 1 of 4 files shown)


fsteeg:
Loading the file in `Global` and setting a static field on the `Lobid` controller seems more complicated than needed. Play has a built-in cache which I think would make sense to use here. It's already used elsewhere in the controller, see: https://github.com/hbz/lobid-resources/blob/880219bc02d119c5422ab6412dc33364440a6663/web/app/controllers/resources/Lobid.java#L148

dr0i:
I don't think this makes sense as it would invoke the loading of the file hundreds of time.
I think play's cache makes sense for caching websites, not files. Or have I missed something?

fsteeg:
Well it would load the file only once, and after that use the cached version. It's not about a browser cache it that's what you are thinking of. It's a server side cache.

TobiasNx:
@dr0i @fsteeg I am not sure I we should proceed with this PR because we are about to add this info to the record metadata: #1888 and #1906

dr0i:
Definitely not :) . See #1868.
Closing.

Issue #1917, opened 2023-09-29
Map sublibraries/libraryCode for POR if `POR  $g` exists


TobiasNx:
At the moment we only map sublibraries (aka the librarycode in ALMA) only in ITM and HOL with H52. They also can be mapped if `POR $g` exists.

blackwinter:
Resolved by #1918.

PR #1918, opened 2023-09-29
1917 por g sublibrary location

TobiasNx:
Related to #1917

This PR does four things:
- add a testfile 990366624790206441
- updates the local sublibrary / libraryCode map
- adds a mapping for libraryCodes/sublibraries based on `POR $g`
- introduces a macro for adding holdingLibraries and opac links that can be used for `ITM`, `HOL` + `H52` and `POR`


blackwinter (changes requested):
Wow, that was quick :) Thank you very much!

However, this change doesn't seem to be covered by test examples (no _new_ `hasItem.seeAlso` in d3ec19c). [990366624790206441](https://lobid.org/marcxml/990366624790206441) would be such an example.

TobiasNx:
Now with updated libraryCode map and the suggested testfile.

blackwinter (approved):
:+1:

dr0i:
Will be deployed next Monday.

PR #1919, opened 2023-10-02
Restrain max size of log file to 100MB

dr0i:
This should prevent log files with size > 60GB.

- update logback to 1.3.11
- reformat


web/conf/logback.xml
```diff
@@ -33,9 +29,6 @@
     <appender-ref ref="STDOUT" />
   </appender>
 
-  <logger name="play" level="INFO" />
-  <logger name="application" level="INFO" />
-
```
fsteeg, line 38: I don't understand why this removes the loggers. Shouldn't the loggers use the new rolling appender somehow?

dr0i: I think we don't use any of these loggers. We just use  the root logger. Or did I miss any use?

dr0i: Ack. Reintroduced with [92f57a0](https://github.com/hbz/lobid-resources/pull/1919/commits/92f57a08eb52184a4f93f097f7765077fb13a243).

dr0i:
As discussed offline with @fsteeg I will merge this.

Issue #1058, opened 2020-02-27
Enrich with RVK based on Culturegraph


dr0i:
From our appointment on 13th February 2020. @hagbeck's workflow (which we shall implement at lobid):

- [x] filter to get all Culturegraph(CG)-clusters with hbz holdings  (https://github.com/hbz/lobid-resources/pull/1921)
- [x] build concordance: hbzID to RVK-Notations (https://github.com/hbz/lobid-resources/pull/1921)
- [x] query this concordance when indexing fulldumps and updates (https://github.com/hbz/lobid-resources/pull/2024)
- [x] [Write Blogpost](https://github.com/hbz/lobid-blog/issues/100)
- [x] update lookup table once a month based on new data from https://data.dnb.de/culturegraph/

See https://katalog.ub.tu-dortmund.de/taxonomy/tree for Dortmund's enrichment.

[Example in culturegraph, see field `084`](https://hub.culturegraph.org/resource/(AT-OBV)990072708180203331/marcxml).

Download CG data at : https://data.dnb.de/culturegraph/ , atm `aggregate_20240507.marcxml.gz`

TobiasNx:
@dr0i should we implement this for ALMA too?

dr0i:
Yeah I definitely have this on my mind ! :)

TobiasNx:
@dr0i I added an shortend approach for this in fix:
https://github.com/metafacture/metafacture-examples/pull/8

dr0i:
Pinging  @hagbeck and @blackwinter : we are considering to go on with this issue. We also consider to provide labels (#1835). Do you actually would use these labels?

blackwinter:
Yes, we would certainly like to use the label. We're [doing it already in IntrOX](https://github.com/hbz/limetrans/blob/d0c15f29f3e22137eb6dc9eb8e2d1dc593042be0/src/main/resources/transformation/marc/otherClassificationNumber.fix#L31) (by [converting the CSV to an LMDB](https://github.com/hbz/limetrans/blob/d0c15f29f3e22137eb6dc9eb8e2d1dc593042be0/bin/maps/rvk.sh)).

hagbeck:
Sorry, I've overseen your ping. Yes, we would like to use the labels, too.

dr0i:
So, we now can extract an RVK-almaMmsId concordance as `csv`. Resulting size is  ~300MB and has `6.658.601`entries - which sounds amazing :)
[...]

PR #1921, opened 2023-10-12
1058-enrichWithCulturegraphRvkWithFix

TobiasNx:
See #1058

I tried to update the draft from https://github.com/hbz/lobid-resources/tree/1058-enrichWithCulturegraphRvk to a version working with fix.
I merged the master into this branch.

In contrast to the morph approach:
- We do not need a separate filter step since the fix already can do this.

Also I do not know if we need:

https://github.com/hbz/lobid-resources/blob/d12b7a34a29ef21c282b0040d74cdb1390345129/src/main/java/org/lobid/resources/run/CulturegraphXmlFilterHbzToJson.java#L30-L45

The decode-marcXml mechanism correctly identified single records.


TobiasNx:
Somehow it does not find the fix file yet.

TobiasNx:
Additionally we would need zdbId as mapping parameter, because new zdb resources dont get hbzIds anymore.

dr0i:
> * `id` is always in the first column

a) As said in https://github.com/hbz/lobid-resources/pull/1921#issuecomment-2135449004 this should be done and makes only sense when creating a CSV. If it's not possible using _one_ FIX  could you provide a second one?
 
b) > all `id's` have their rows
Hm, should'nt that be possible with using triples somehow ? Anyway. Then we shoud make the first ID to the only ID (the only value for column one).

dr0i:
The build of the concordance just started, based on the 9,2 GB file  `aggregate_20240507.marcxml.gz` mentioned in https://github.com/hbz/lobid-resources/issues/1058#issue-572148389.
At a first glance - there are many doublette RVK entries, e.g.:
> "990191558260206441","HU 8423, HU 8424, HU 8423, HU 8424, HU 8424, HU 8423, HU 8424"

Can you prevent these doublettes via the FIX @TobiasNx ?

src/main/resources/rvk/cg-to-rvk-json.fix
TobiasNx, line 1: Why do we need two fixes?

dr0i: `json` doesn't make a record for every `id` - the _search_ is done by the search engine, in contrast to a csv where we nee a single unique key. The `json` is moe performant when using search engines, the `csv` is the only way to go when using tables.

src/main/resources/rvk/cg-to-rvk-json.fix
```diff
@@ -0,0 +1,29 @@
+set_array("rvk[]")
+
+do list(path: "084??", "var": "$i")
+    if any_match("$i.2", "rvk")
+        copy_field("$i.a","rvk[].$append")
+    end
+end
+
```
TobiasNx, line 8:
```suggestion
uniq("rvk[]")
```

dr0i:
I am going to merge this. 
Note that the branch name is not good, resp. that a new PR is going to be made where we not only generate the concordance but enrich our lobid-resources data with it.

dr0i (approved):

PR #1931, opened 2023-10-25
Rename biblioVinoId to schoeneNummer (RPB-37)

fsteeg:
See https://jira.hbz-nrw.de/browse/RPB-37

### DIFF
--- a/src/main/resources/alma/index-config.json
+++ b/src/main/resources/alma/index-config.json
@@ -391,7 +391,7 @@
           "analyzer": "id_analyzer",
           "type": "text"
         },
-        "biblioVinoId": {
+        "schoeneNummer": {
           "analyzer": "id_analyzer",
           "type": "text"
         },
--- a/web/app/controllers/resources/Queries.java
+++ b/web/app/controllers/resources/Queries.java
@@ -412,7 +412,7 @@ public class Queries {
 	public static class IdQuery extends AbstractIndexQuery {
 		@Override
 		public List<String> fields() {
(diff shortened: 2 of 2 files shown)


dr0i (changes requested):
Are there no tests? I would expect some jsonld with the new field name (could be here https://github.com/hbz/rpb/).
Also,  if you want to provide RDF the "schoeneNummer" needs to be defined in context.jsonld.

fsteeg:
We're only reusing the index config from RPB, no data is indexed into a lobid-resources index. Test data in RPB is here: https://github.com/hbz/rpb/commit/ff07a7c (was not in main when I openend this PR). RDF is not a use case here, and `biblioVinoId` (the old name) was not in the context either.

dr0i:
Can you explain why this is done in this lbid-resources repo and not in the rpb repo - the changes here don't have anything to do with lobid-resources IMO?

fsteeg:
> Can you explain why this is done in this lbid-resources repo and not in the rpb repo

We originally indexed RPB titles into the lobid-resources index (and might still do that in the future), so we're using the lobid-resources `index-config.json`. Also RPB uses a lobid-resources instance as its backend, hence the change in `IdQuery`. (And again, this is just the rename, this stuff is already in lobid-resources, see https://github.com/hbz/lobid-resources/pull/1436 and https://github.com/hbz/lobid-resources/pull/1460.)

dr0i (approved):

dr0i:
OK, thx for explaining.

PR #1948, opened 2023-12-01
Fix related resource with repeated subfield

maipet:
Use 1st occurrence of subfield w
Adjust regex for linking to related resources

### DIFF
--- a/src/main/resources/alma/fix/relatedRessourcesAndLinks.fix
+++ b/src/main/resources/alma/fix/relatedRessourcesAndLinks.fix
@@ -341,11 +341,15 @@ end
 # 776 - Additional Physical Form Entry (R) -  Subfield: $w (R), $t (NR), $i (R), $n (R), $x (NR), $z (R)
 
 set_array("related[]")
-do list(path:"775??|776??", "var":"$i")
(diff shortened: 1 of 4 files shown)


dr0i (approved):
+1 . Just a minor thing be discussed offlin is missing - the link to the issue further describing this workaround for there is one link to `related` missing.

TobiasNx:
I am not sure if `do list_as` is the correct mechanism here. Do we have a corresponding ticket?

TobiasNx:
> I am not sure if `do list_as` is the correct mechanism here. Do we have a corresponding ticket?
> 
> do list_as iterates over all 775??|776?? at the same time while do list does only one of these.

Okay it seems that  there seems to be a follow up merge, that has no PR  @maipet  do I see this correctly?

https://github.com/hbz/lobid-resources/commit/1911e7b4f2a696c5f208c292591f563fb6d469e2
Also there is no corresponding ticket for the issue.

I will open one, since the solution from 
https://github.com/hbz/lobid-resources/commit/1911e7b4f2a696c5f208c292591f563fb6d469e2 also quite impressive creates duplicate objects out if multiple ISBNs.

Issue #1952, opened 2023-12-28
Repeated subfield "d" in 100/700


maipet:
If subfield "d" (Dates associated with a name) in 100/700 is repeated, the fields should be skipped (according to MARC, the field ist NR)
e.g. MMSID 99371463467006441
```
<datafield tag="100" ind1="1" ind2=" ">
<subfield code="a">Sheffield, Alfred Dwight,</subfield>
<subfield code="d">1871-1961.,</subfield>
<subfield code="d">Sheffield, Alfred Dwight,</subfield>
<subfield code="d">1871-1961.,</subfield>
...
```

dr0i:
To be checked next Monday.

dr0i:
ping @maipet and @TobiasNx : you check and if ok close

TobiasNx:
Looks good

https://lobid.org/resources/99371463467006441.json

Closing.

PR #1954, opened 2024-01-04
Fix contribution transformation

maipet:
See #1952.

### DIFF
--- a/src/test/resources/alma-fix/99371463467006441.json
+++ b/src/test/resources/alma-fix/99371463467006441.json
@@ -1,0 +1,138 @@
+{
+  "@context" : "http://lobid.org/resources/context.jsonld",
+  "almaMmsId" : "99371463467006441",
+  "title" : "Training for group experience",
+  "alternativeTitle" : [ "Training for group experience" ],
+  "otherTitleInformation" : [ "A syllabus of materials from a laboratory course for group leaders given at Columbia University in 1927" ],
+  "publication" : [ {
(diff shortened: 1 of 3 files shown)


TobiasNx:
+1

dr0i:
Will be deployed next Monday.

src/test/resources/alma-fix/99371463467006441.json
```diff
@@ -0,0 +1,138 @@
...
+  "@context" : "http://lobid.org/resources/context.jsonld",
+  "almaMmsId" : "99371463467006441",
+  "title" : "Training for group experience",
+  "alternativeTitle" : [ "Training for group experience" ],
+  "otherTitleInformation" : [ "A syllabus of materials from a laboratory course for group leaders given at Columbia University in 1927" ],
+  "publication" : [ {
+    "startDate" : "1929",
+    "type" : [ "PublicationEvent" ],
+    "location" : [ "New York :" ],
+    "publishedBy" : [ "Inquiry" ]
```
dr0i, line 11: Strange - this build succesfully with GitHub actions first - but after merging it fails because it expects `[ "New York" ]` (i.e. without colon).

TobiasNx: Okay it seems that the commit was not okay, I did no good review here.

For an better approach we should add the new test file in a separate first commit, if there is none yet.
In the later commit we should add the code changes and include the changes in the test to highlight the reaction.

Issue #1956, opened 2024-01-12
Repeated subfield current library


maipet:
In field H52 it is permitted to repeat subfield b. Fix should be adjusted

dr0i:
Should be resolved with #1957 and will be deployed next Monday. Closing.

PR #1957, opened 2024-01-12
Adjust holdings for repeated subfield b (#1956)

maipet:
See #1956


TobiasNx:
@maipet there seems to be a follow up merge, that has no PR do I see this correctly?

https://github.com/hbz/lobid-resources/commit/1911e7b4f2a696c5f208c292591f563fb6d469e2

Also there is no corresponding ticket.

Issue #1835, opened 2023-07-26
RVK-Labels?


TobiasNx:
RVK has a csv which we could use for a lookup.
It is a littlebit messy, but we would need adjust it.
https://rvk.uni-regensburg.de/downloads/FachA-Z_2023_2.csv.gz

dr0i:
Feel free to contact me @maipet if you have any questions.

maipet:
@dr0i first attempt to add the rvk labels is available: label starts with "benennung" of the base notation. 
In some cases the lookup doesn't seem to work as expected, e.g. mmsId 991005935279706485 -> notation is invalid and should be deleted with the lookup.

dr0i:
Discussed offline: @maipet  will add the table into  https://github.com/hbz/lookup-tables/ , along with the python script (or Fix and Flux ) to https://github.com/hbz/lookup-tables/tree/master/docs to be able to reproduce the creation of the new `csv` (which was generated working on the `xml` file because the original downloadable `csv` is not very good).

dr0i:
See also [#1085.](https://github.com/hbz/lobid-resources/issues/1058)

dr0i:
See also this [comment where it is demanded to have labels](https://github.com/hbz/lobid-resources/issues/1058#issuecomment-2056900285).

maipet:
Reply from H.G. (15.05.2024): "... Auch wir würden erstmal nur die Notation verwenden. ..."

PR #1961, opened 2024-01-17
Add RVK labels incl. lookup tsv (#1835)

maipet:
#1835


dr0i (changes requested):
See https://github.com/hbz/lobid-resources/issues/1835#issuecomment-1898153338 .

PR #1967, opened 2024-02-09
Update Play app and GitHub action to Java 11 (#1965)

### LINKED ISSUE #1965: Switch to Java 11
We decided to bring up our java code to a higher, maintained version. Raising java 8 to Java 11 seems to be a good start.

As we want to have our libraries (see https://github.com/metafacture/metafacture-core/issues/518) also to version 11 it's mandatory for compiling to also have the play apps working with java 11.

fsteeg:
Will resolve #1965.


dr0i:
Will update the start scripts here too for when we would merge this alone it would break our deployment.

PR #1969, opened 2024-02-27
Replace Mapbox-based MakiMarkers with local default icon

### LINKED ISSUE #1966: Map markers don't load
Same as https://github.com/hbz/nwbib/issues/637

fsteeg:
Will resolve #1966, see e.g. http://stage.lobid.org/resources/990182586200206441

(And two other tweaks for things I noticed.)


dr0i (changes requested):
Code ok , but the marker cannot be loaded as only ./web/public/javascripts/images/marker-icon-2x.png exists but it wants ./web/public/javascripts/images/marker-icon.png .

fsteeg:
That's weird, I actually had to rename the file since it wanted `marker-icon-2x.png`, not `marker-icon.png`. And I thought it was working when I deployed earlier. Will look into it.

fsteeg:
Oh, I think I understand: my office screen is high-res, and I might have some scaling enabled on the OS level. So there, it requests the `2x` version, so it worked when I tested and deployed it earlier. Now, on my home setup, I get the same error as you. Will add the other icon.

fsteeg:
Deployed to test: http://test.lobid.org/resources/990182586200206441

(If you want, you could deploy to prod and stage if this is done and there is anything unusual to consider for deployment.)

dr0i (approved):

Issue #1977, opened 2024-03-27
Fix Error caused by 990166236770206441


TobiasNx:
```
2024-03-23 20:15:19 [WARN] org.metafacture.metafix.Metafix in Thread-8 - Error while executing Fix expression (at [file:/home/sol/git/lobid-resources-alma/web/target/universal/stage/conf/alma/alma.fix](file:///home/sol/git/lobid-resources-alma/web/target/universal/stage/conf/alma/alma.fix), line 33): include ("./fix/item.fix") -> Error while executing Fix expression (at [file:/home/sol/git/lobid-resources-alma/web/target/universal/stage/conf/alma/fix/macros.fix](file:///home/sol/git/lobid-resources-alma/web/target/universal/stage/conf/alma/fix/macros.fix), line 300): lookup("$i.@opacAlmaMmsId","isil2opac_almaMmsId", delete:"true")
```
=> `Caused by: java.lang.IllegalStateException: Expected String, got Array`

Reason There are two Element `H52` linking to the same `HOL`-Object in $8. 

```
  <datafield tag="H52" ind1="8" ind2=" ">
    <subfield code="b">P0001</subfield>
    <subfield code="c">UEO</subfield>
    <subfield code="8">22195808780006463</subfield>
  </datafield>
  <datafield tag="H52" ind1="8" ind2="2">
    <subfield code="z">ab Bd. 31 Einzelsignaturen</subfield>
    <subfield code="8">22195808780006463</subfield>
  </datafield>
```

Solution only one has `$b` so skip the linking if `$b` does not exist.

dr0i:
To be checked next Monday.

TobiasNx:
Error is gone. Closing

PR #1978, opened 2024-03-27
1977-duplicateH52

TobiasNx:
Fix reported error #1977


dr0i:
In your first commit you have only added the XML , not the resulted JSON. In the second commit you then added the JSON. Thus no diff is possible, which was your intention I assume.
I will add the JSON to the first commit and force push so we can better see what your Fix changes.

TobiasNx:
Nope! This is intentional, since the test breaks!!! The second commit resolves this.

dr0i:
Yeah I see - it's not even possible to generate that JSON ;)

dr0i:
There are conflicts - can you resolve these @TobiasNx ? I also wouldn't mind if you rebase and force push.

dr0i:
`Build and test with Play` fails. Have a look at the README (search for `play`) . If that is not sufficiently helpful we should talk together and you would bring the README to perfection :)

TobiasNx:
@dr0i fixed it.

dr0i (approved):

dr0i:
Good :)
Will be deployed next Monday. Closed.

Issue #1866, opened 2023-08-16
Add ALMA IZ to holding entry


TobiasNx:
When analyzing the records, due to the library codes in hasItem it is difficult to sort the records to the right holder. Sometimes not all institutions belonging to one ~mainholder~ALMA IZ have the same ISIL pattern.

~Therefore I suggest adding the main holder or parentOrganization to the holding entry.~

We could add to heldBy a https://schema.org/memberOf - property linking to the IZ and using the lobid-organisation `id` and add the `label` `IZ of [ISIL]`

TobiasNx:
@acka47 made a good suggestion. The IZ is no info about the holding institution but about the item. Therefore we should add an `inCollection`statement to the item.

dr0i:
Can you provide an example which points to the problem? All I see in the PR is a lot of redundany in all test files.

TobiasNx:
Now has the IZ as inCollection Statement:
```
 "hasItem": [
    {
      "label": "lobid Bestandsressource",
      "type": [
        "Item",
        "PhysikalischerTitel"
      ],
      "currentLibrary": "R5001",
      "currentLocation": "UNASSIGNED",
      "heldBy": {
        "isil": "DE-1116",
        "id": "http://lobid.org/organisations/DE-1116#!",
        "label": "Hochschule für Wirtschaft und Gesellschaft Ludwigshafen, Bibliothek"
      },
      "seeAlso": [
        "https://opac.hwg-lu.de/alipac/-/item-local?BASE=B-TIT&VID=HT020579803"
      ],
      "inCollection": [
        {
          "id": "http://lobid.org/organisations/DE-1116#!",
          "label": "Hochschule für Wirtschaft und Gesellschaft Ludwigshafen, Bibliothek"
        }
      ],
      "id": "http://lobid.org/items/990367731740206441:DE-1116:229499030007824#!"
    }
```

Also is searchable: http://lobid.org/resources/search?q=hasItem.inCollection.id%3A%22http%3A%2F%2Flobid.org%2Forganisations%2FDE-1116%23%21%22

PR #1982, opened 2024-04-04
Add IZ statement to holding item #1866

TobiasNx:
See #1866


TobiasNx:
@dr0i this should be indexed so we can search for it.

dr0i:
a) You want the `hasItem.inCollection.id` to be searchable I guess?
b) We should stick to one label. E.g. this is not good:
```json
    heldBy : {
      isil : DE-468,
      id : http://lobid.org/organisations/DE-468#!,
      label : Universitätsbibliothek Wuppertal
    },
    inCollection : [ { 
      id : http://lobid.org/organisations/DE-468#!,
      label : IZ of DE-468
    } ],
```
Also note the slight inconcruency by using  `heldBy.isil` while leaving this out for `inCollection`. I don't think this is crucial but maybe you just've missed it.

src/test/resources/alma-fix/990016782920206441.json
```diff
@@ -158,6 +178,10 @@
       "id" : "http://lobid.org/organisations/DE-62#!",
       "label" : "Stadtbibliothek Wuppertal"
     },
+    "inCollection" : [ {
+      "id" : "http://lobid.org/organisations/7816#!",
```
dr0i, line 182: This is no valid organisation.

dr0i (changes requested):

TobiasNx:
Good catch this seems to be no true IZ but a bridge. We should be able to identify a bridge like 49HBZ_BRIDGE_FIZ .

https://service-wiki.hbz-nrw.de/pages/viewpage.action?pageId=949911658#NZPublishingf%C3%BCrDiscoveryAnbindung:Lobid-Members

src/test/resources/alma-fix/990016782920206441.json
TobiasNx: I talked to @acka47 this is `Bibliotheken ohne zentrale Katalogisierung	49HBZ_BRIDGE_FIZ	7816` it has no ISIL an therefore no entry in lobid-organisations. I have introduced a dummy uri now: http://lobid.org/organisations/BRIDGE_FIZ-7816#! so we can identify this bridge.

TobiasNx:
@dr0i yes please set it to searchable.

Thanks for the hint with the isil, but we do not need it here. The incollection statement is just to identify the IZ and not the organisation itself.

dr0i (approved):

PR #1895, opened 2023-09-18
Add Spiegel as testfile

TobiasNx:
Testfile for #1830

PR #1997, opened 2024-05-16
Add  editorconfig-maven-plugin

TobiasNx:
Discussed with @dr0i to add check for editorconfig with https://github.com/ec4j/editorconfig-maven-plugin

I added the plugin with 1. commit to the pom.xml and 2. commit reformatted all files except the test xml and json files with `mvn editorconfig:format`

This does not seem to test automatically, perhaps that has to be configured. For now one has to run `mvn editorconfig:check`

Can you do this @dr0i


dr0i (changes requested):
Don't apply formatting to binary data. This may break things unexpectedly.

dr0i:
Have a look at  [47ba540](https://github.com/hbz/lobid-resources/pull/1997/commits/47ba540365cb8978f0f45beb937760ff52abd338) , there the  build already fails.

I've added the `mvn editorconfig:check` to GitHub action - this is what is done e.g. in [metafacture](https://github.com/metafacture/metafacture-core/blob/master/.github/workflows/build.yml#L14).

pom.xml
```diff
@@ -248,6 +248,35 @@
...
+							</goals>
+						</execution>
+					</executions>
+					<configuration>
+						<!-- See http://ec4j.github.io/editorconfig-maven-plugin/ for full configuration reference -->
+						<excludes>
+							<!-- Note that maven submodule directories and many non-source file patterns are excluded by default -->
+							<!-- see https://github.com/ec4j/editorconfig-linters/blob/master/editorconfig-lint-api/src/main/java/org/ec4j/lint/api/Constants.java#L37 -->
+							<!-- You can exclude further files from processing: -->
+							<exclude>src/main/**/*.whatever</exclude>
```
dr0i: This is just a template and should be adopted to our needs, e.g. ignoring binary files.

pom.xml
```diff
@@ -248,6 +248,35 @@
 						</execution>
 					</executions>
 				</plugin>
+				<plugin>
+					<groupId>org.ec4j.maven</groupId>
+					<artifactId>editorconfig-maven-plugin</artifactId>
+					<version>0.1.3						                        						                        						<!-- check the latest on http://search.maven.org/maven2/org/ec4j/maven/editorconfig-maven-plugin/ --></version>
+					<executions>
```
dr0i: Remove these comments - they are just there to begin with.

pom.xml
```diff
@@ -248,6 +248,35 @@
...
+							<id>check</id>
+							<phase>verify</phase>
+							<goals>
+								<goal>check</goal>
+							</goals>
+						</execution>
+					</executions>
+					<configuration>
+						<!-- See http://ec4j.github.io/editorconfig-maven-plugin/ for full configuration reference -->
+						<excludes>
```
dr0i: comments see above

pom.xml
```diff
@@ -248,6 +248,35 @@
...
+							<!-- Note that maven submodule directories and many non-source file patterns are excluded by default -->
+							<!-- see https://github.com/ec4j/editorconfig-linters/blob/master/editorconfig-lint-api/src/main/java/org/ec4j/lint/api/Constants.java#L37 -->
+							<!-- You can exclude further files from processing: -->
+							<exclude>src/main/**/*.whatever</exclude>
+							<exclude>src/test/resources/alma-fix/</exclude>
+						</excludes>
```
dr0i: Why excluding the test files? AFAIR we started this isssue here because of the different formatting of test files, originating from https://github.com/hbz/lobid-resources/pull/1895#issuecomment-2114169914.

dr0i (changes requested):

pom.xml
```diff
@@ -248,6 +248,35 @@
 						</execution>
 					</executions>
 				</plugin>
+				<plugin>
+					<groupId>org.ec4j.maven</groupId>
+					<artifactId>editorconfig-maven-plugin</artifactId>
```
dr0i: I still like `indent_size=2` esepcially for `XML` files. So adapt `.editorconfig` to either set `indent_size=2` for `[*]` or add `xml` to `[*.{fix,gradle,json,md,mwe2,sh,vim,xtext,yml}]
`.

pom.xml
TobiasNx: The testfiles are generated automatically. The pom.xml had the problem that triggered our initiative to use a editor config plugin.

dr0i: "The testfiles are generated automatically." Not necessarily. E.g. you open such an XML in your editor which reformats it automatically . I think that was the case in [66782bb](https://github.com/hbz/lobid-resources/pull/1895/commits/66782bb1c237790a12f159503152ff8ae6bae45e) (which doesnt' exist anymore - but it was there that I complained that diffs are unreadable because every line was changed by reformatting).

TobiasNx: It formats only when saving. I never save the test xml files.
In the old commit you complained because I changed the heap size in the `pom.xml` and also changed the formatting in the pom.xml at the same time. Because of that I added a separate commit https://github.com/hbz/lobid-resources/pull/1895/commits/214ad1aa20bab1d3a2e1be8f6304b037b02f0da3 before https://github.com/hbz/lobid-resources/pull/1895/commits/6ba89272653eda944fd4bf5476d0f111f2bc8c9c.

pom.xml
```diff
@@ -1,254 +1,283 @@
...
+              </goals>
+            </execution>
+          </executions>
+          <configuration>
+            <!-- See http://ec4j.github.io/editorconfig-maven-plugin/ for full configuration reference -->
+            <excludes>
+              <!-- Note that maven submodule directories and many non-source file patterns are excluded by default -->
+              <!-- see https://github.com/ec4j/editorconfig-linters/blob/master/editorconfig-lint-api/src/main/java/org/ec4j/lint/api/Constants.java#L37 -->
+              <!-- You can exclude further files from processing: -->
+              <exclude>src/main/**/*.whatever</exclude>
```
dr0i, line 270: This doesn't match anything. I think you took it from the example and should remove this also.

pom.xml
TobiasNx: Did that

pom.xml
TobiasNx: Deleted comment

pom.xml
TobiasNx: Deleted comment

pom.xml
TobiasNx: Deleted this

pom.xml
TobiasNx: Not sure if we still need to configurate this more properly. The template is deleted.

TobiasNx:
> Don't apply formatting to binary data. This may break things unexpectedly.

not sure how to do this. @dr0i can you help.

web/conf/context.jsonld
TobiasNx, line 1: It seems that the automated creation of this file does not add a trailing blank line. @dr0i should we exclude it from the format check?

.github/workflows/build.yml
```diff
@@ -33,7 +33,9 @@ jobs:
         ./gradlew publishToMavenLocal
         cd ..
     - name: Build with Maven
-      run: mvn install
+      run: |
+        mvn install
+        mvn editorconfig:check
```
TobiasNx, line 38: @dr0i This seems not only check the files in the repo but also the files downloaded or created by this action. This leads to failing GH-Action Build. Could you have a look.

dr0i:
Replaced by #2007. Closing.

Issue #1996, opened 2024-05-08
Redirect with HT-Nr deprecatedUri to detail page of one record when there are two hits? Instead redirect to results or delete deprecatedUri from CZ record.


TobiasNx:
Related to: https://github.com/hbz/nwbib/issues/611#issuecomment-2060901315

Due to the fact, that there are duplicate E-Book records of EZB publications because of CZ/Consor and NZ/ZDB we have faulty redirects. Only one record is linked not both matching records because both records have mapped HT-Nr. (`hbzId`) and (`deprecatedUri`). That created the confusion in: https://github.com/hbz/nwbib/issues/611#issuecomment-2060901315

We should 
**a)** either redirect to the result page instead of only redirecting to one detail page 
 OR 
**b)** we should delete the `deprecatedUri` link and do not map the `hbzId` from the CZ/Consors-Record, since the CZ-Consors-Record was never part of the Verbundkatalog and therefore should not have a HT-Nr. Especially because the ZDB publication do not receive any HT Numbers in ALMA anymore and we mapped the old HT-Numbers to be able to resolve the old URIs. So only the old Verbundkatalog/BZ/ZDB records have a redirect, a `deprecatedUri` and a `hbzId`

Writing this I prefer variant **b**. 

----------

e.g.

http://lobid.org/resources/HT015679368 redirects to https://lobid.org/resources/99370698631806441.
But there are two records with `deprecatedUrl: http://lobid.org/resources/HT015679368`
The other one is: https://lobid.org/resources/990171221570206441

CZ/Consors record: https://lobid.org/resources/99370698631806441. In variant b it would lose the `deprecatedUr`and the `hbzId`

```JSON
{
  "@context": "http://lobid.org/resources/context.jsonld",
  "almaMmsId": "99370698631806441",
  "urn": [
    "urn:nbn:de:hbz:6:2-69622"
  ],
  "oclcNumber": [
    "644257601"
  ],
  "dnbId": "990456757",
  "zdbId": "2451605-3",
  "deprecatedUri": "http://lobid.org/resources/HT015679368#!",
  "hbzId": "HT015679368",
  "title": "ABU-Info",
```

NZ/ZDB record: https://lobid.org/resources/990171221570206441

```JSON
{
  "@context": "http://lobid.org/resources/context.jsonld",
  "almaMmsId": "990171221570206441",
  "urn": [
    "urn:nbn:de:hbz:6:2-69622"
  ],
  "oclcNumber": [
    "984935842"
[...]

PR #1999, opened 2024-05-16
Only map HT Nr. for hbz NZ resources #1996

TobiasNx:
Related to #1996


blackwinter (approved):
These titles are _currently_ not included in the union catalogue anyway.

dr0i (approved):

Issue #2000, opened 2024-05-17
Sort elements in JSON


TobiasNx:
The `id`, `type`, `title` elements are not at the top.

For better presentation of the JSON Data
it would be nice to have the metadata in a specific order.

With fix this is not possible yet: https://github.com/metafacture/metafacture-fix/issues/61

e.g.:

http://lobid.org/resources/990208680360206441.json

```JSON
"@context": "http://lobid.org/resources/context.jsonld",
  "almaMmsId": "990208680360206441",
  "hbzId": "HT018708677",
  "deprecatedUri": "http://lobid.org/resources/HT018708677#!",
  "isbn": [
    "9783898999199",
    "389899919X"
  ],
  "oclcNumber": [
    "912956557"
  ],
  "title": "Ich lerne Ringen",
  "edition": [
    "2., überarb. Aufl"
  ],
  "publication": [
    {
      "startDate": "2015",
      "type": [
        "PublicationEvent"
      ],
      "location": [
        "Aachen"
      ],
      "publishedBy": [
        "Meyer & Meyer"
      ]
    }
  ],
```

should be something like:

```JSON
"@context": "http://lobid.org/resources/context.jsonld",
  "id": "http://lobid.org/resources/990208680360206441#!",
  "title": "Ich lerne Ringen",
   "type": [
    "BibliographicResource",
    "Book"
  ],
  "almaMmsId": "990208680360206441",
  "hbzId": "HT018708677",
  "deprecatedUri": "http://lobid.org/resources/HT018708677#!",
  "isbn": [
    "9783898999199",
    "389899919X"
  ],
  "oclcNumber": [
    "912956557"
  ],

  "edition": [
    "2., überarb. Aufl"
  ],
  "publication": [
    {
      "startDate": "2015",
      "type": [
        "PublicationEvent"
      ],
      "location": [
        "Aachen"
      ],
      "publishedBy": [
        "Meyer & Meyer"
      ]
    }
  ],
``

blackwinter:
> With fix this is not possible yet

It is possible if you emit those fields first. Other than that, no, not yet.

acka47:
Here is what I think should be the top fields:

- `@context`
- `id`
- `type`
- `medium`
- `title`

TobiasNx:
@dr0i could you help with: https://github.com/hbz/lobid-resources/pull/2003#issuecomment-2122492604

dr0i:
@TobiasNx I have done so, please have a look again.

[...]

PR #2003, opened 2024-05-17
Manipulate the order of the records #2000

TobiasNx:
See #2000


src/main/resources/alma/alma.fix
```diff
@@ -23,6 +23,12 @@ end
 
 add_field("@context","http://lobid.org/resources/context.jsonld")
 
+# Set empty elements to manipulate the order winthin the record.
```
dr0i, line 26:
```suggestion
# Set empty elements to manipulate the order within the record.
```

dr0i (approved):
besides the typo  +1

blackwinter:
What about the `id` field? Apparently, it's not moved to the top.

TobiasNx:
> What about the `id` field? Apparently, it's not moved to the top.

Why did I miss that?? 
I don't know why this does not work.
The `id` element is created here:

https://github.com/hbz/lobid-resources/blob/888a22fe10318230eb29cee87f736a806772f000/src/main/resources/alma/fix/identifiers.fix#L3

But somehow it is the last element to be passed on. And still is even when I add an empty ìd` element at the beginning.

blackwinter:
Oh, wait, maybe it's because `id` is a _special_ field in Metafix. I'll take a look next week.

blackwinter:
It's not in Metafix (I was confusing it with `_id`), but in lobid-resources: `JsonLdEtikett` always puts `id` last.

https://github.com/hbz/lobid-resources/blob/master/src/main/java/org/lobid/resources/EtikettJson.java#L80-L82

TobiasNx:
> It's not in Metafix (I was confusing it with `_id`), but in lobid-resources: `EtikettJson` always puts `id` last.
> 
> https://github.com/hbz/lobid-resources/blob/2c1d7fa315aee8cd37c0d76e1aa476006760ba42/src/main/java/org/lobid/resources/EtikettJson.java#L80-L82

@dr0i can you help here. Why does the EtikettJson put the id always last?

TobiasNx:
+1

Issue #1340, opened 2022-05-24
Make lobid-alma data valid against JSON schema


acka47:
https://gist.githubusercontent.com/TobiasNx/007a32d61457dc57e353c5f1cd97a5e0/raw/4e9f525f114c0ab06279d28b2c70854cc5c6cee8/validationError.txt

This is an list of the errors of the test data.

TobiasNx:
I spottet two errors in the validation process with the test data:

```
alma/(DE-605)TT050421649.json failed test
[
  {
    instancePath: '/describedBy/resultOf/endTime',
    schemaPath: 'describedBy.json/properties/resultOf/properties/endTime/pattern',
    keyword: 'pattern',
    params: {
      pattern: '(\\d{4})-(\\d{2})-(\\d{2})T(\\d{2}):(\\d{2}):(\\d{2})'
    },
    message: 'must match pattern "(\\d{4})-(\\d{2})-(\\d{2})T(\\d{2}):(\\d{2}):(\\d{2})"'
  }
]
alma/(OCoLC)945571548.json failed test
[
  {
    instancePath: '/contribution/0',
    schemaPath: 'contribution.json/items/required',
    keyword: 'required',
    params: { missingProperty: 'role' },
    message: "must have required property 'role'"
  }
]
```

TobiasNx:
@dr0i `/describedBy/resultOf/endTime` is created correctly later in the process (when indexing?) in the transformation process it self there is only "dummi" as value added. Therefore it breaks. Any way that we still could validate these?

dr0i:
That's a "feature" as the test files, at whatever date created, comparable. It might be worth to think about using a valid dummy pattern, e.g. 0000-00-00T00:00:00 . Or you could expand the validator to allow "dummy".

TobiasNx:
> 0000-00-00T00:00:00

+1 for that

dr0i:
Should be fine now. Closing.

TobiasNx:
Again I run this with the updates from #1344:

```
$ bash ./validateJsonTestFiles.sh  
Testing version: draft
strict mode: "items" is 1-tuple, but minItems or maxItems/additionalItems are not specified or different at path "type.json"
alma/(CKB)5280000000199164.json failed test
[
  {
    instancePath: '/hasItem/0/type/0',
    schemaPath: 'hasItem.json/items/properties/type/items/const',
    keyword: 'const',
    params: { allowedValue: 'Item' },
[...]

PR #2027, opened 2024-06-06
Schema: Adjust hasItem: loose constant value for type property #1340

maipet:
Schema only accepted `Item` as `type` in `hasItem` but we also have additional type classes:
`PhysicalObject`, `DigitalDocument`, `NurTitel`, `PhysikalischerTitel`

Since #1699 is not solved yet @maipet suggested to loosen up the type specification. (Written by @TobiasNx )

### DIFF
--- a/src/test/resources/schemas/hasItem.json
+++ b/src/test/resources/schemas/hasItem.json
@@ -15,7 +15,7 @@
             "type": {
                 "title": "Type",
                 "type": "array",
-                "items": {
+                "contains": {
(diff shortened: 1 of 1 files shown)


TobiasNx (approved):
+0/1 But we should adjust it again, if we find a solution for the properties: #1699

acka47 (approved):
+1 Looks good, assuming we really set `type` `Item` for every object in the `hasItem` array and add other optional types in addition.

Issue #2034, opened 2024-07-23
Adjust Regex for URN pattern in field 024


maipet:
cf. https://lobid.org/resources/990198125850206441
Reported by E.S. via email on 2024-07-23: Currently, the fulltext links from MARC-field 856 with http://**www.mdz**-nbn-resolving.de/ are not displayed / transformed in lobid. 
The reason seems to be processing the field 856 - URN (cf. https://github.com/hbz/lobid-resources/blob/master/src/main/resources/alma/fix/identifiers.fix#L21-L23), the regex pattern doesn't match "www.mdz"

TobiasNx:
This is fixed now.

PR #2035, opened 2024-07-24
adjust urn regex

maipet:
#2034


fsteeg (commented):
Updated the web app tests for the new test data to fix the build.

Looking at the code I'm wondering if the match-regex shouldn't be the same as the replace-regex, to make sure it is actually replaced if it matches? But will assign @TobiasNx for proper functional review.

TobiasNx:
Functionally this looks good. Next time combine the changes in the  Fix and the changes in the testfiles in one commit.

Otherwise is this hint from @fsteeg a must? Is there a scneario where there is a urn without http?

> Looking at the code I'm wondering if the match-regex shouldn't be the same as the replace-regex, to make sure it is actually replaced if it matches?

Otherwise +1

TobiasNx:
We talked about it off board, lets try to create the urn and the resolver links differently:

Fetch all urn from the links AND copy all urn links that are marked Resolver-System in $x without formatting except doi links.

TobiasNx:
@maipet I added some commits to better handle the URNs and Resolver Links. The links in 865 with Resolver System were already used but were skipped if they had a `nbn` refrence in them. I specified the regex pattern a little bit more and decoupled the URN links in 865 from the URN in 024 and I only use the URN from 024 as fallback for links.

Additionally I have changed the URL for nbn links to org instead of de since that seems to be the standard.

TobiasNx:
This pr is replaced by #2043

Issue #2033, opened 2024-07-22
add additional relations


hagbeck:
We have a new case of related works in our data. MARC field 505 sometimes contains information about another part of the current title. Subfield g describes what it is (a note) and subfield 6 contains the corresponding HBZ ID if the title exists in the network.
It would be nice if this information also appeared in lobid-resources. Gladly with the matching Alma-MMS-ID ;-)

Part 1: https://lobid.org/marcxml/99374022974006441
Part 2: https://lobid.org/marcxml/99374024871506441

TobiasNx:
I can create a workable link with the deprecated URI based on the hbzId but I cannot create the ALMA-MMS-ID out of nothing.

Will create a draft and you can have a look if this suits your needs.

TobiasNx:
Should be online next week

TobiasNx:
@hagbeck good like this?

https://lobid.org/resources/99374022974006441.json
https://lobid.org/resources/99374024871506441.json

hagbeck:
Looks good. Thanks!

PR #2038, opened 2024-07-31
Add related property for 505 with $6 hbzId.

TobiasNx:
See #2033

@hagbeck would this meet your demands?


hagbeck:
Looks good. 
Of course, it would be nice if the relation had a type. It would also be interesting to see what else we get with this feature.

TobiasNx:
> Looks good. Of course, it would be nice if the relation had a type. It would also be interesting to see what else we get with this feature.

Talked to @hagbeck offboard 505 (https://www.loc.gov/marc/bibliographic/bd505.html) is not a specific content, it CAN beside other thins refer to resources that are part of the records resource. We are not able to add a separate type, but the additional note in $g as provided as related.note.

acka47 (approved):
+1

PR #2040, opened 2024-08-05
Schema: Adjust spatial schema to validate RPB spatial info too.

TobiasNx:
Adjust spatial schema to validate RPB spatial info too. 

Related to #1340


acka47 (approved):
+1

maipet (approved):

Issue #2045, opened 2024-08-07
Skip 084 notations with z


TobiasNx:
```
z
    Other
    [used when a source code has not yet been assigned to a classification scheme]
```
https://www.loc.gov/standards/sourcelist/classification.html

We should skip these, since we cannot assign a specific notation system to them.

acka47:
Resolved with #2046. Closing.

PR #2046, opened 2024-08-07
Improve some subject mappings

TobiasNx:
See #2045 and #2047

We have some notation/concept systems that were not mapped properly.
If they are `z` they are not defined but sorted under "Other",
`sswd`  on the other hand is the GND Systematik as SKOS System.

Not sure if we should also use the Labels for the GND Systematik.

This will also fix some validation errors from #1340


acka47 (approved):
⁺1

Issue #1848, opened 2023-08-01
Missing Dewey Classification.


TobiasNx:
was: http://aleph.lobid.org/resources/HT007216566#!
![grafik](https://github.com/hbz/lobid-resources/assets/61879957/ac8fa0be-a596-451a-bd45-9fd6acfcfc16)

is: http://lobid.org/resources/990056283990206441#!
![grafik](https://github.com/hbz/lobid-resources/assets/61879957/f71b1b5b-02fb-453d-8c73-78d4ed86ea8c)

It seems that this is due to the classification change, sdnb is not mapped as Dewey.
I am not sure why, but sdnb in 084 seems not to be the same as dewey classification in 082.

TobiasNx:
I have done a first change giving labels to DDC Sachgruppen of DNB and ZDB, we could additionally add all DDC-Sachgruppen to  the Dewey Subjects as we handled it before in ALEPH. What do you think Adrian.

TobiasNx:
I added DNB Sachgruppen and DDC Sachgruppen der ZDB. I also talked to @acka47 to add all of them also as "normal" DDC. 

@acka47 I am not very pleased with the id that I used for DNB Sachgruppen and DDC Sachgruppen der DDC. Do you have better suggestions?

TobiasNx:
There is now sswd = > seems to be a GND Notation.
Otherwise looks good closing: http://aleph.lobid.org/resources/HT007216566#

PR #2049, opened 2024-08-08
Final adjustments and corrections for schema validation

TobiasNx:
See #1340

This PR:

- Adds missing id for "DDC-Sachgruppen der ZDB"
- Adds a variation in the skosConcept schema to either reqiure id AND label or jus label or notation without id.
- Fix a formal error in the json schema, instead of tuple validation, the items element has to be an object instead of an array, otherwise it would only check the first element.
- Adds a missing value to the type schema.


src/main/resources/alma/fix/subjects.fix
```diff
@@ -296,8 +296,10 @@ do list(path:"084??", "var":"$i")
             if any_match("$j","\\d{3}(\\.\\d{1,3})?|[BKS]")
               if any_equal("$i.q","DE-600")
                 add_field("subject[].$last.source.label","DDC-Sachgruppen der ZDB")
+                add_field("subject[].$last.source.id","https://zeitschriftendatenbank.de/fileadmin/user_upload/ZDB/pdf/zdbformat/5080.pdf")
```
acka47, line 299: Interesting choice of `id` for this source. While I like the pragmatism, this sets a new precedent linking to the description of a controlled vocabulary in a documentation PDF. As I have no better idea, I guess we can leave it for now.

acka47: On the other hand, we should probably better leave out the `id` field and add the documentation link in another field, e.g. `seeAlso`.

acka47 (changes requested):
The change to `skosConcept.json` looks good but we have to discuss the `id` for DDC-Sachgruppen der ZDB.

src/main/resources/alma/fix/subjects.fix
TobiasNx: I already had a bad feeling a while ago about id: https://github.com/hbz/lobid-resources/issues/1848#issuecomment-1707846009

TobiasNx: But we decided off board to go on for a lack of a better id.

acka47: Ok, then let's go. +1

acka47 (approved):

Issue #2062, opened 2024-08-27
faulty language ids


maipet:
The way the language ID is created is error-prone, it allows any value out of the Marc data. Perhaps the values should first be validate against ISO?
e.g. 
http://id.loc.gov/vocabulary/iso639-2/deutsch -> https://lobid.org/resources/990126426530206441.json
http://id.loc.gov/vocabulary/iso639-2/engl. -> https://lobid.org/resources/990173607780206441.json
http://id.loc.gov/vocabulary/iso639-2/ger024 -> https://lobid.org/resources/99373983156706441.json
...

TobiasNx:
The errors seem to result from: https://github.com/hbz/lobid-resources/blob/ddcd72b0c136572f00d9c6f97dddcff7780b932b/src/main/resources/alma/fix/otherFields.fix#L7

There some non standardized language values can sneak in.

TobiasNx:
https://www.alma-dach.org/alma-marc/bibliographic/041/041.html

If indicator 2 of 041 is " " empty, the marc codes should be used. But it seems they are not always as @maipet examples show.

We should introduce an conditional to test if the value has more than 3 letters and therefore meets the pattern. Unfortenatly we do not have a morph like list filter mechanism yet.

dr0i:
Will be deployed next Monday.

PR #2072, opened 2024-09-13
Skip not language info that are not three letters long

TobiasNx:
See #2062

@maipet could you have a look. 

You also had a list of more not matching values. Could you check if you can add any additional mapping?


TobiasNx:
@maipet I adjusted the transformation as you suggested at the office.

maipet (approved):

TobiasNx:
@dr0i could you merge if okay.

dr0i (approved):

dr0i:
Will be deloyed next Monday. Closed.

PR #2075, opened 2024-09-25
2074-introxOpacLinks

### LINKED ISSUE #2074: Add adjustment for Introx opac links
In Introx only CZ/NZ resources that are not part of NZ/Verbundkatalog have OPAC-Links based on ALMA MMS ID, all others have ZDB or HBZ identifier as base.

TobiasNx:
Resolves #2074. Needed for https://github.com/hbz/lookup-tables/pull/56

Only CZ resources work with alma mms id links in introx, NZ/Verbund resources need hbzId or zdbId as base.


blackwinter (changes requested):
I'm afraid we had this discussion before, but why so complicated? Why not just walk through the priority list (hbz ID, ZDB ID, MMS ID, ISSN) and use the first one found?

Also, note that I said "found"; it may be possible for a library not to have OPAC links of all types defined. Due to this uncertainty, I would not feel comfortable with the proposed solution.

TobiasNx:
> I'm afraid we had this discussion before, but why so complicated? Why not just walk through the priority list (hbz ID, ZDB ID, MMS ID, ISSN) and use the first one found?
i would be okay with this if the lookup tables are cleaned up and those libraries that just use alma mms id have no hbzId opac link and no zdbId alma link in the lookup table.


> 
> Also, note that I said "found"; it may be possible for a library not to have OPAC links of all types defined. Due to this uncertainty, I would not feel comfortable with the proposed solution.

I think we skip those without any matching opac links. Any suggestion how to change the mapping for those?

TobiasNx:
> > i would be okay with this if the lookup tables are cleaned up and those libraries that just use alma mms id have no hbzId opac link and no zdbId alma link in the lookup table.
> 
> Of course, there should be no non-functional links. Are you aware of any?

Not sure about non functional links, but I checked if only digibib libraries have almaMmsIds, hbzId and zdbId links they do. In my opinion we can change the order then.

TobiasNx:
I redid this PR and opened a new PR #2076

TobiasNx:
Replaced by #2076

PR #2077, opened 2024-09-25
Skip all records with only suppressed holdings/items/portfolios.

blackwinter:
Mimics similar behaviour in DigiBib IntrOX (Limetrans; relevant Jira issues: DOX-1354, DOX-1568, DOX-1622).

If you don't want to exclude these records altogether, we'd have to apply this change to the `inCollection` statement for "DigiBib hbz Verbundkatalog" instead.

### DIFF
--- a/src/main/resources/alma/fix/item.fix
+++ b/src/main/resources/alma/fix/item.fix
@@ -3,7 +3,10 @@ set_array("@ITM-H") # Helper element for creating Holding from HOL ("Physikalisc
 do list(path:"ITM  ", "var": "$i")
   paste("$i.supressedLocation", "$i.M", "~+", "$i.x",join_char:"")
   lookup("$i.supressedLocation","suppressedLocations",delete:"true")
-  unless exists("$i.supressedLocation") # Test if location is suppressed with mapping provided by the libraries.
+  if exists("$i.supressedLocation") # Test if location is suppressed with mapping provided by the libraries.
+    add_field("@suppressed", "true")
+  else
+    add_field("@unsuppressed", "true")
     set_hash( "hasItem[].$append")
     add_field("hasItem[].$last.label", "lobid Bestandsressource")
     set_array("hasItem[].$last.type[]", "Item","PhysicalObject")
@@ -31,7 +34,10 @@ do list(path: "HOL  ", "var": "$i")
         if exists("$H52.b")
           paste("$H52.supressedLocation", "$i.M", "~+", "$H52.c",join_char:"")
           lookup("$H52.supressedLocation","suppressedLocations",delete:"true")
-          unless exists("$H52.supressedLocation") # Test if location is suppressed with mapping provided by the libraries.
+          if exists("$H52.supressedLocation") # Test if location is suppressed with mapping provided by the libraries.
+            add_field("@suppressed", "true")
+          else
+            add_field("@unsuppressed", "true")
             set_hash( "hasItem[].$append")
             add_field("hasItem[].$last.label", "lobid Bestandsressource")
             set_array("hasItem[].$last.type[]", "Item","PhysikalischerTitel")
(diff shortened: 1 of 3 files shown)


TobiasNx:
Concerning the suppression I will get feedback. to you what we prefer.

I would have preferred two commits because one cannot relate the changes in the tests to the changes in the code so easily. I was confused that the reject mechanism resulted in the deletion of a single holding object. But then I realized that this was due to the new "suppression" mechanism for POR that was also commited.

blackwinter:
> Concerning the suppression I will get feedback. to you what we prefer.

Okay, thanks.

> I would have preferred two commits because one cannot relate the changes in the tests to the changes in the code so easily.

Which changes exactly would you have separated?

src/main/resources/alma/fix/item.fix
```diff
@@ -55,44 +61,49 @@ do list(path: "HOL  ", "var": "$i")
...
-    call_macro("holdingLibrary+opacLink", mmsIdField:"$i.a",libraryCodeField:"$i.g")
-  end
-  # entity for every POR  .A
-  if exists ("$i.A")
-    copy_field("$i.D", "$i.@electronicLocator")
-    replace_all("$i.@electronicLocator","https://eu04.alma.exlibrisgroup.com/view/uresolver/49HBZ_NETWORK","")
-    copy_field("$i.d", "$i.@sublocation")
-    replace_all("$i.@sublocation","https://hbz-network.userservices.exlibrisgroup.com/view/uresolver/49HBZ_NETWORK","")
-    do list(path:"$i.A", "var": "$j")
+  if none_equal("$i.b", "Not Available")
```
TobiasNx, line 64: This conditional introduces something new. A suppression mechanism that was not there before.

blackwinter: Neither were the other two. But you would prefer one commit with the holding/item changes (no test coverage) and one with the portfolio changes?

blackwinter: Ah, I think I understand. Sorry for being dense so early in the morning ;) You mean the suppression of the portfolio, not the suppression of the record; holdings and items were suppressed before, but the portfolio wasn't.

blackwinter:
So there are two decisions to be made:

1. Should unavailable portfolios be suppressed?
2. Should records with only suppressed holdings/items/portfolios be suppressed altogether, or only the "DigiBib hbz Verbundkatalog" collection?

I'll adjust the commits once you've made your decision.

acka47:
My first thoughts on this (after not much reflection): It makes sense to skip those records but we should make sure to not exclude records that we actually want in lobid. E.g. in NWBib, there are 52k resources that are not articles and do not have any holding information: [`inCollection.id:"http://lobid.org/resources/HT014176012#!" AND NOT _exists_:hasItem AND NOT type:Article`](https://lobid.org/resources/search?q=inCollection.id%3A%22http%3A%2F%2Flobid.org%2Fresources%2FHT014176012%23%21%22+AND+NOT+_exists_%3AhasItem+AND+NOT+type%3AArticle)

blackwinter:
> do not have any holding information

Just to be clear: The proposed change only applies to records that only have _suppressed_ holdings; records without _any_ holdings aren't skipped (neither are records with both suppressed and unsuppressed holdings).

TobiasNx:
How do we ensure that these are not suppressed:
https://lobid.org/resources/search?q=hasItem.type%3A%22DigitalDocument%22+AND+%28hasItem.type%3A%22NurTitel%22+OR+hasItem.type%3A%22PhysikalischerTitel%22%29

And

https://lobid.org/resources/search?q=hasItem.type%3A%22PhysicalObject%22+AND+%28hasItem.type%3A%22NurTitel%22+OR+hasItem.type%3A%22PhysikalischerTitel%22%29

Because the supression marker is only set with ITM, H52 (PhysicalObject) and POR (DigitalDocument) , but not for the other types of holdings (NurTitel and PhysikalischerTitel)

blackwinter:
> Because the supression marker is only set for ITM, H52 (PhysicalObject) and POR (DigitalDocument) , but not for the other types of holdings (NurTitel and PhysikalischerTitel)

You mean, e.g., what if a record has a suppressed `H52` and a _different_ `HOL` without an (unsuppressed) `H52`? That certainly complicates things.

I guess the safest option would be to limit the change to the "DigiBib hbz Verbundkatalog" collection (question 2 above). This way you don't have to suffer any unforeseen consequences.

blackwinter:
See #2078 for alternate proposal.

blackwinter:
The discussion might not have been totally in vain, but we've decided not to go forward with this change (for the time being). Sorry for the noise, everyone.

(The examples which @eschackmann identified didn't match the use case we previously saw in IntrOX and weren't covered by the conditions implemented here.)

PR #2078, opened 2024-09-26
Exclude records with only suppressed holdings/items/portfolios from "DigiBib hbz Verbundkatalog" collection.

blackwinter:
Mimics similar behaviour in DigiBib IntrOX (Limetrans; relevant Jira issues: DOX-1354, DOX-1568, DOX-1622).

See #2077 for alternate proposal (skip these records altogether).


blackwinter:
The question remains: Do we need tests for these cases?

blackwinter:
We've decided not to go forward with this change (for the time being). Sorry for the noise, everyone.

PR #2080, opened 2024-10-07
Support auto-complete suggestions created from multiple fields

fsteeg:
Like in lobid-gnd: configure fields to use in the `format` param (e.g. `json:title,contribution`), or use `format=json:suggest` for default fields.

See https://jira.hbz-nrw.de/browse/RPB-141

Deployed to test: https://test.lobid.org/resources/api#auto-complete

### DIFF
--- a/web/app/controllers/resources/Application.java
+++ b/web/app/controllers/resources/Application.java
@@ -14,6 +14,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


acka47:
This looks already very good. There is one place where we should probably improve it: When resources have multiple contributing agents, the list of names should be cut after the first person and an "et al." added (like in the lobid HTML result lists).

Example: https://test.lobid.org/resources/search?q=%22Enhancing%20space%20resilience%20through%20non-materiel%20means%20%22&format=json%3Asuggest

acka47 (changes requested):
See my comment

fsteeg:
Hm, but isn't that example a case where we actually need all authors to select the correct entry?

acka47:
Maybe we should just wait what RPB editors say.

fsteeg:
RPB editors are done with review, re-requesting review from @acka47 (request for @dr0i is still open).

https://jira.hbz-nrw.de/browse/RPB-141

acka47 (approved):
+1

dr0i:

Issue #2082, opened 2024-10-15
Map genre/form label if GND-ID is available in 655 but no label


maipet:
Map the label from https://github.com/hbz/lobid-resources/blob/5c4dc57859a49592c1f519fb2057f6c25ea288f9/src/main/resources/alma/maps/formangabe.tsv if GND-ID exists but label $a is empty.
At the moment the subfield $a must exist if a GND-ID is given
E.g. 991030115479706476

dr0i:
Deployed next Monday. To be checked then.

TobiasNx:
@maipet https://lobid.org/resources/991030115479706476.json is that what you expected?

maipet:
looks good. Closing

PR #2028, opened 2024-06-06
Change publishedBy for secondaryPublicationEvent to array #2011

TobiasNx:
Related to #2011

secondary publication event had an error in the transformation, since it mapped `publishedBy` as string instead of the needed array.

@maipet this fixes the error for invalid `publishedBy`-statement. #1340

PR #2090, opened 2024-11-05
lookup formangabe

maipet:
#2082


TobiasNx (changes requested):
Fixes look good. But I would change the commits a bit so that you can better see the changes that your code introduces.

Add the test records from https://github.com/hbz/lobid-resources/pull/2090/commits/1fed75909c1c68eba6f61089f261e31d9cd6bf5e in an commit before adding the changes in the code.
In a second commit add the changes in the fixes as well as the changes that result when running: `~/git/lobid-resources$ mvn clean install -DskipTests=false -DgenerateTestData=true`. You can also fix the formatting in this commit.

In a third commit  add http://lobid.org/resources/99374153235806441 as test for fixing the columns  of the formangaben lookup.

In a fourth commit add your changes to the fix from: https://github.com/hbz/lobid-resources/pull/2090/commits/bb77c450fbe05119e8b0213f0f1c62f957972d8d  as well as the changes that result when running: `~/git/lobid-resources$ mvn clean install -DskipTests=false -DgenerateTestData=true`. 

In a last commit update the web app tests or you could add the updates of the web app tests when introducing a new test. (Which I usually forget)

TobiasNx:
> @TobiasNx Thx for your suggestions. As the JSON was not created due to the incorrect transformation, it is not possible to see the changes between two results. I have therefore edited this in one step. Where does the example http://lobid.org/resources/99374153235806441 come from? Sorry, but I cannot see any relation to #2082.

The example came up in context of #2028 and was provided by you having a wrong publication date. But the JSON Schema Tests og the github action failed due to the mapped natureOfContent.id being wrong: https://github.com/hbz/lobid-resources/actions/runs/11680572050/job/32523855291#step:8:187

It will be fixed by your commit: https://github.com/hbz/lobid-resources/pull/2090/commits/bb77c450fbe05119e8b0213f0f1c62f957972d8d Therefore it would be a good test file for this commit.

maipet:
replaced by #2092

PR #2096, opened 2024-11-14
2095-noOverwritingHbzId

TobiasNx:
This PR introduces a conditional to stop overwriting existing `hbzId`.

Context:

There are old aleph zdb records that have zdb id informations
 in their MAB data in `026-1` but not in `025z1`.

Due to this in context of the migration to ALMA their HT number was not replaced by the ZDB ID. Also  in the old aleph lobid mapping these records had no `zdbId`

e.g. see: 
ALEPH: http://lobid.org/hbz01/HT019286510 // https://aleph.lobid.org/resources/HT018628300
ALMA: https://lobid.org/marcxml/990217879290206441 // https://lobid.org/resources/990217879290206441

---

To ensure that old links still work we added an enrichment of all records with zdbId that are part of the NZ based on the latest data from aleph data before the switch to ALMA as master.

Also we with ALMA we started to catch the ZDB ID from `035` not only when they have the prefix `(DE-600)` but also if `(DE-599)ZDB`.

This resulted in overwriting existing hbzId in the records with mapped ones. We want to adjust this mechanism so that only records WITHOUT `zdbId`are enriched with their old `hbzId`


dr0i:
Build fails.

dr0i (approved):

TobiasNx:
I unassigne Adrian as reviewer, since he should only see that we introduced this change. Due to the discussion at metadaten.community.

TobiasNx:
@dr0i can you merge and deploy since the automatic deployment is already through?

dr0i:
Manually deployed. Data should be updated next Monday.

Issue #2098, opened 2024-11-20
Improve supressed location mapping to check for libraryCodeSubfield and locationCodeSubfield


TobiasNx:
Currently we only check for the institution and for the location but do not check for the specific library within this institution. This can suppress location that should not be suppressed due to non uniq location codes. Only in combination with the library code it working.

@blackwinter is creating a new mapping https://github.com/hbz/lookup-tables/pull/60 

That will replace the old mechanism.

blackwinter:
Did you mean to link this to #2097?

TobiasNx:
> Did you mean to link this to #2097?

yes

TobiasNx:
This is already in practice

PR #2100, opened 2024-11-22
Update mapping file for suppressed locations #2098

TobiasNx:
@blackwinter

I updated the test vocab as you commited in context of https://github.com/hbz/lookup-tables/pull/60

Is the result as expected?


blackwinter (approved):
Yes, everything looks correct. Not all changes are due to hbz/lookup-tables#60, though. You haven't updated the mapping in a while ;)

Issue #2114, opened 2024-12-19
Change `describedBy.object.dateCreated` from `MNG` to 008/00-05


TobiasNx:
@maipet discovered that we are not using the real date of cataloguing in our transformation for  `describedBy.object.dateCreated`  but the date of indexing in ALMA which is provided in `MGN` which is the date of migration i am not mistaken.

I suggest we change  `describedBy.object.dateCreated`  to 008/00-05.

But I do not know what the idea was behind choosing MGN instead.

[`describedBy.object.dateCreated`  was based on MAB 002 which was mapped to Marc 008/00-05.](https://service-wiki.hbz-nrw.de/pages/viewpage.action?pageId=483000321)

> 002a
> 002b
> 
> ->	
> 
> 008/00-05
> 	
> -> 
> 
> Datum der Ersterfassung
> Datum Fremddatenübernahme wird nach MARC 21 nur übernommen, wenn 002a nicht vorhanden
> (Rückkonvertierung 002b → 002a)

TobiasNx:
Context seems to be: https://github.com/hbz/lobid-resources/issues/1095

TobiasNx:
One cannot answer the RDA query anymore: https://hyp.is/0igroCTlEeeDn2vimOYVRg/lobid.org/resources/api Also the documentation would need some updates.

dr0i:
Will be deployed next Monday.

TobiasNx:
There seems to be a significant number of cataloguing errors with 008. The index reported a couple of parser exceptions:

e.g.

```LOG
2025-01-02 04:36:47 [WARN] org.lobid.resources.ElasticsearchIndexer in Thread-8 - Bulk insert failed: failure in bulk execution:
[1275]: index [resources-alma-fix-20241228-1154], type [resource], id [99374515437806441], message [MapperParsingException[failed to parse [describedBy.resultOf.object.dateCreated]]; nested: IllegalArgumentException[Cannot parse "2020-24-08": Value 24 for monthOfYear must be in the range [1,12]];]
2025-01-02 04:36:53 [WARN] org.lobid.resources.ElasticsearchIndexer in Thread-11 - Bulk insert failed: failure in bulk execution:
[...]

PR #2115, opened 2024-12-19
Use initial cataloguing date instead of alma indexing date #2114

TobiasNx:
MNG only states the data when the record was indexed into ALMA not initially catalogued. When migrating from ALEPH to ALMA the information from MARC 002 (fromer `describedBy.object.dateCreated`) was mapped to MARC 008/00-05. This commit reintroduce the initital cataluging date again and uses the alma indexing date only as fallback.

See: https://github.com/hbz/lobid-resources/issues/2114


acka47 (approved):
+1

maipet (approved):

dr0i:
Will be deployed next Monday.

PR #2118, opened 2025-01-06
Add workaround if cataloguing dates in 008 that are too long

TobiasNx:
As stated in https://github.com/hbz/lobid-resources/issues/2114#issuecomment-2572433239 some issues popped up with cataloguing dates that are too long consisting of 8 instead of 6 digits. 

Since 008/06 should not be an digit I have introduced a workaround if the date is to long.
@maipet can you have a look if this workaround makes sense?

Thanks.


maipet (approved):

TobiasNx:
@dr0i this should fix the mapper parser exception. The error was due to errors in cataloguing. Some records provided 8 instead of 6 digits [which is not valid with regard to 008 marc documentation. ](https://www.loc.gov/marc/bibliographic/bd008.html)

PR #2139, opened 2025-01-24
Adjust eclipse configuration

TobiasNx:
@dr0i could you have a look.

I updated the eclispe configuration since they still were set to Java 1.8


dr0i:
As I don't use `eclipse` but @fsteeg does I assign the latter.

fsteeg:
Since these config files are very dependent on the local setup (Eclipse version etc.) I think we should rather delete them in the repo and add them to `.gitignore`.

Issue #2141, opened 2025-02-10
Validate leap year dates, Feb 29


TobiasNx:
```
2025-02-09 08:18:13 [WARN] org.lobid.resources.ElasticsearchIndexer in Thread-10 - Bulk insert failed: failure in bulk execution:
[1835]: index [resources-alma-fix-20250208-1345], type [resource], id [990129250080206441], message [MapperParsingException[failed to parse [describedBy.resultOf.object.dateCreated]]; nested: IllegalArgumentException[Cannot parse "1997-02-29": Value 29 for dayOfMonth must be in the range [1,28]];]
```

dr0i:
That's a leap year. Only way to find these is to use a library. No sane regex would be sufficient.

TobiasNx:
1996 was a leap year, 1997 was not. I would only introduce a simple regex tester.

dr0i:
Thx. Will be deployed next Monday.

TobiasNx:
```json
"object": {
        "id": ["https://lobid.org/marcxml/990129250080206441"](https://lobid.org/marcxml/990129250080206441),
        "dateCreated": "1997-02-28",
        "dateModified": "2021-04-08",
```

looks good. closing

PR #2142, opened 2025-02-10
Validate leapYear #2141

TobiasNx:
See #2141

ElasticSearch throws `MapperParsingException` if leap year date 02-29 is not correct. I introduced a simple regex to test this part of this answer:

https://stackoverflow.com/a/8648129

Tests for range 1800-2099 and changes to date to 02-28 if leap year date is incorrect.


dr0i (changes requested):
Build is failing. Besides this +1.

PR #2144, opened 2025-02-11
WIP_testDE_Sol1_Mapping

TobiasNx:
Related to https://github.com/hbz/lobid-extra-holdings/issues/2


TobiasNx:
@dr0i could you have a look why `mvn clean install -DskipTests=false -DgenerateTestData=true` is brought to its knees because of the new lookup?

PR #2146, opened 2025-02-11
2145-De-Sol1-Mapping

TobiasNx:
This PR introduces a lookup for DE-Sol1 Holdings created in context of https://github.com/hbz/lobid-extra-holdings/issues/2

It uses a map record id in column 1 and holdings as value of json array in column 2. The mapped json data is deserialized in the lobid transformation and added to `hasItem`

Replaces #2144


TobiasNx:
@dr0i could you fix the actions?

src/test/resources/alma-fix/990110486750206441.json
```diff
@@ -0,0 +1,245 @@
...
+    "inCollection" : [ {
+      "id" : "http://lobid.org/organisations/DE-Sol1#!",
+      "label" : "Stadtarchiv Solingen, Bibliothek"
+    } ],
+    "heldBy" : {
+      "isil" : "DE-Sol1",
+      "id" : "http://lobid.org/organisations/DE-Sol1#!",
+      "label" : "DE-Sol1#!"
+    },
+    "label" : "Stadtarchiv Solingen, Bibliothek",
```
acka47, line 198: I think, we should use another label here. AFAICS, we are currently using in lobid-resources "lobid bestandsressource" in the string at this position. We could use the same here of – if we want to mark the difference of the source data (Aleph/Strapi instead of Alma) also in the `label`– something "DE-Sol1-Bestandsressource" or "Exemplar von Stadtarchiv Solingen, Bibliothek"  o.ä.

acka47: Or was this label intended to be in `hasItem.heldBy.label` where we now have "DE-Sol1#!`? This would make sense..

acka47 (changes requested):

src/test/resources/alma-fix/990110486750206441.json
TobiasNx: I fixed this by fixing the fix: https://github.com/hbz/lobid-extra-holdings/commit/7de5d028f176c3cab8811acb5183652b41b0aea7 and uploading the newly generated upload file: https://github.com/hbz/lobid-extra-holdings/commit/579842349ac69d53feb7b197da2607b13c7c735b

acka47 (approved):
+1

dr0i (approved):

PR #2147, opened 2025-02-12
Avoid duplicating MARC 500 (General Note).

blackwinter:
If the field contains a citation ("In:"), the content gets added to both `note` and `bibliographicCitation`.

Ex.: 990110402310206441


TobiasNx:
Since this is an `bibliographicCitation` is an element introduced by @acka47 I would rather let him do the review.

This was the old morph transformation:
https://github.com/hbz/lobid-resources/blob/7f982c13a25f9d6a4ef1a2c81a3749cf1cdaed36/src/main/resources/morph-hbz01-to-lobid.xml#L7032

Issue #2160, opened 2025-03-04
Map labels  for GND-Subject-Classifications in 084


TobiasNx:
https://d-nb.info/standards/vocab/gnd/gnd-sc.html#19.1d

https://lobid.org/resources/990107385450206441.json

dr0i:
Check after tomorrow https://github.com/hbz/lobid-resources/pull/2162#issuecomment-2714692914.

dr0i:
Hm, the `ttl` was there but it doesn't seem to work, se e.g.:
https://lobid.org/resources/search?q=subject.notation%3A%2219.1d%22%20AND%20describedBy.resultOf.object.dateModified:2025-03-12&format=json

```
{
      "notation": "19.1d",
      "type": [
        "Concept"
      ],
      "source": {
        "label": "GND-Systematik",
        "id": ["http://d-nb.info/standards/vocab/gnd/gnd-sc"](http://d-nb.info/standards/vocab/gnd/gnd-sc)
      }
    },
```
I've forgot to rebuild because only then the resources are copied to the proper path, i.e. `./web/target/universal/stage/conf/alma/maps/gnd-sc.ttl` and/or `./target/classes/alma/maps/gnd-sc.ttl`.

Check again next Monday.

TobiasNx:
It seems to work now:

lobid.org/resources/search?q=subject.notation%3A%2219.1d%22%20AND%20describedBy.resultOf.object.dateModified:2025-03-12&format=json

```
{
          "notation": "19.1d",
          "id": ["https://d-nb.info/standards/vocab/gnd/gnd-sc#19.1d"](https://d-nb.info/standards/vocab/gnd/gnd-sc#19.1d),
          "label": "Reise",
          "source": {
            "label": "GND-Systematik",
            "id": ["http://d-nb.info/standards/vocab/gnd/gnd-sc"](http://d-nb.info/standards/vocab/gnd/gnd-sc)
          },
          "type": [
            "Concept"
          ]
        }
```

PR #2162, opened 2025-03-05
Introduce mapping for GND Subject Categories labels #2160

TobiasNx:
@dr0i this would need a cronjob to fetch https://d-nb.info/standards/vocab/gnd/gnd-sc.ttl

Related #2160


TobiasNx:
@dr0i what should I do?

dr0i:
I need a location ehre to download the file to. A directory called `test` seems not the proper one wrt production.

TobiasNx:
> I need a location where to download the file to. A directory called `test` seems not the proper one wrt production.

https://github.com/hbz/lobid-resources/pull/2162/files#diff-8057d2f6f68f3245d1e3843c6323177281a5f79498da99191ef40c016dcf019bR145 This should be the location. I think that most of the cronjob place them here.

dr0i:
Maybe this PR is not in the scope of what note here, but:
I must admit, I don't understand why `./src/test/resources/alma/maps` differs so much from `./src/main/resources/alma/maps`. I mean, we use the same `fix`s for testing ETL but why using different maps?
Then,  a map like `hbzId2zdbId.tsv.gz` (in `test`) consists of 16 MB (!) - is this necessary?
And `isil2opac_isbn.tsv` consists of 0 byte - does this makes sense? Then, I didn't found `isil2opac_isbn.tsv` in `main` but at the server they are incorporated via GitSubmodules  in `./lookup-tables/data/opacLinks/isil2opac_isbn.tsv`.
To conclude (at least we don't change our workflow):
- maps that are only present via other repos , incorporated via GitSubmodules, may be copied (as smaller test version, e.g. only some lines instead of 16 MB) to the `test` directory
- other maps , residing in `main`, could be reused by the tests. This seems to be the case already, as there are no duplictaes in `test` and `main` (in contradiction to what you propose in https://github.com/hbz/lobid-resources/pull/2162#issuecomment-2705812820 )

dr0i (changes requested):
Could we reuse the map ? Could we move it from `test` to `main` ?

TobiasNx:
> Maybe this PR is not in the scope of what I note here, but: I must admit, I don't understand why `./src/test/resources/alma/maps` differs so much from `./src/main/resources/alma/maps`. I mean, we use the same `fix`s for testing ETL but why using different maps? Then, a map like `hbzId2zdbId.tsv.gz` (in `test`) consists of 16 MB (!) - is this necessary? And `isil2opac_isbn.tsv` consists of 0 byte - does this makes sense? Then, I didn't found `isil2opac_isbn.tsv` in `main` but at the server they are incorporated via GitSubmodules in `./lookup-tables/data/opacLinks/isil2opac_isbn.tsv`. To conclude (at least we don't change our workflow):
> 
>     * maps that are only present via other repos , incorporated via GitSubmodules, may be copied (as smaller test version, e.g. only some lines instead of 16 MB) to the `test` directory
> 
>     * other maps , residing in `main`, could be reused by the tests. This seems to be the case already, as there are no duplictaes in `test` and `main` (in contradiction to what you propose in [Introduce mapping for GND Subject Categories labels #2160 #2162 (comment)](https://github.com/hbz/lobid-resources/pull/2162#issuecomment-2705812820) )

I think this would be related to:
https://github.com/hbz/lobid-resources/issues/2164 

I think that you are missing a third and a fourth scenario. 
3) Maps that are collected via cronjobs because they update reguarily and therefore create diffs to files. They are copied into `./maps`, a static version for testing resides in `test`. (I have no overview concerning the existing cronjobs. Could be the lobid-organisations label mapping.)
4) Maps that are not public, also reside in maps, and only testsamples are provided in `test`. (e.g. DDC and RVK mapping)

----
Concerning `hbzId2zdbId.tsv.gz`. Yes this is indeed necessary, we would loose all old hbzIds of zdb publication, if we drop this mapping. Unforetunately it is big.

TobiasNx:
@dr0i could you setup the cronjob for the file [d-nb.info/standards/vocab/gnd/gnd-sc.ttl](https://d-nb.info/standards/vocab/gnd/gnd-sc.ttl) saving it to `src/main/resources/alma/maps/`

dr0i:
I've scheduled a cron job, every 2th Wednesday at `00:04 am` the ttl is wget'ed. Although the comments suggests a relative stable version:
```
    rdfs:comment """
            Changes from 2021-02-09 to 2024-08-06
                        Add skos:notation
                        Add skos:narrower for first three top-levels
                        Remove owl:Ontology as type
                        Add skos:hasTopConcept
            Changes from 2019-10-15 to 2021-02-08
                        Add english labels
                        Changes from 2017-09-12 to 2019-10-15
                        All d-nb.info-URIs changed from http to https
                """@en ;
```
this don't hurts because the file is only `135K` sized.

dr0i:
To be checked after tomorrow (a) file retrieved and b) successfully lookuped).

dr0i:
I've forgotten to rebuild because only then the resources are copied to the proper path, i.e. `./web/target/universal/stage/conf/alma/maps/gnd-sc.ttl` and/or `./target/classes/alma/maps/gnd-sc.ttl`.

PR #2176, opened 2025-06-02
DuplicateVariableInDoListInDoList

TobiasNx:
Came up in RPB-288. See https://github.com/hbz/rpb/pull/151

@dr0i this fixes some basic issues with the transformation e.g. not creating a link when no $$w or no match Prefix in $$w is given. Also it adjusts the duplicate use of the variable $j within the do list bind.

The test result stays the same.


dr0i (approved):

dr0i:
Will be deployed next Monday.

PR #2177, opened 2025-06-03
Update config to index the top-level `note` field

fsteeg:
Requirement from https://jira.hbz-nrw.de/browse/RPB-284

### DIFF
--- a/src/main/resources/alma/index-config.json
+++ b/src/main/resources/alma/index-config.json
@@ -1005,7 +1005,6 @@
           ]
         },
         "note": {
-          "index": false,
           "type": "text"
         },
         "ismn": {


dr0i:
I don't like it as e.g. 
1.`supplement` is a word often used within the `note` field without the resource having to do anything with "medicine"
2. the index would be bloated

Of course none of my concerns alone must result in a :-1:  so let's see what @blackwinter is thinking about it. 
(if it's a no-go for digibib we may configure this field as not part of a default query (that _is_ possible I think?) or something along the line)

dr0i (approved):

dr0i:
(I am just waiting with merging if RPB customer consider to explicitly enrich their subjects instead of adding the "dirty" note field to their default query @fsteeg (I've asked that in [the](https://jira.hbz-nrw.de/browse/RPB-284)  ).

fsteeg:
We currently don't have any sort of branch or fork of the lobid-resources instance used for RPB. Only a separate instance using a separate index, but created with the same config as lobid-resources. This keeps having a common index in the future a realistic option.

Issue #2179, opened 2025-06-11
Add additional mapping for rpb subjects and spatial since cataloguing changed


TobiasNx:
See: RPB-289 (https://jira.hbz-nrw.de/browse/RPB-289?focusedId=334423&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-334423)

RPB subjects and spatial notations used to be mapped like this:

See: https://lobid.org/marcxml/99371152051606441

```
<datafield tag="084" ind1=" " ind2=" ">
<subfield code="a">rpbr_99_o13207076_</subfield>
<subfield code="2">rpb</subfield>
</datafield>
<datafield tag="084" ind1=" " ind2=" ">
<subfield code="a">rpbr_10_</subfield>
<subfield code="2">rpb</subfield>
</datafield>
```

and now LBZ follows the mapping of NWBib like this:

```
<datafield tag="084" ind1=" " ind2=" ">
<subfield code="a">Verbandsgemeindebeschreibung</subfield>
<subfield code="0">http://purl.org/lobid/rpb#n102042</subfield>
<subfield code="2">rpb</subfield>
</datafield>
<datafield tag="084" ind1=" " ind2=" ">
<subfield code="a">Prüm, Verbandsgemeinde</subfield>
<subfield code="0">https://rpb.lobid.org/spatial#n23206</subfield>
<subfield code="2">rpb</subfield>
</datafield>
```

`$$a` now holds the subject instead of the notation and `$$0` hold the URI. The notation is not given anymore.

This does not really conform to LoC Documentation of Marc: https://www.loc.gov/marc/bibliographic/bd084.html But seems to be the convention choosen by hbz with regard to NWBib.

This change only applies to all new cataloguing and the old records will probably not be updated. Therefore we have to support both ways when mapping RPB `subjects` and `spatial` info.

It is still unclear if LBZ Notation will change too.

PR #2180, opened 2025-06-11
2179-additionalRpbSpatialSubjectsMapping

TobiasNx:
See #2179 and RPB-289

This PR adds an additional new mapping for RPB spatial and subjects since the new way of cataloguing follows the pattern of NWBib but the old versions are still kept. Therefore we have to keep the mapping for both ways.

Functional Review by @acka47


dr0i (approved):

acka47:
The test output looks fine. +1

Issue #2186, opened 2025-06-30
Add 591 info (Redaktionelle Bemerkung) to lobid


TobiasNx:
In context of NWBib we were asked to add the info from 591 to our lobid data in context of https://github.com/hbz/nwbib/issues/658

This could be temporarily.

@acka47 suggested to add the info as note to the holdings of the Münster as note.

If I remember correctly BT tiles do not have any or only some holdigs not even the ones of Münster, since they are duplicates from the NZ and the goal is to add holding to these titles.

Therefore I suggest to use another element:

e.g.: https://schema.org/disambiguatingDescription

acka47:
> If I remember correctly BT tiles do not have any or only some holdigs not even the ones of Münster, since they are duplicates from the NZ

You are right, so a note on the `Item` is not a practical approach.



> Therefore I suggest to use another element:
> 
> e.g.: https://schema.org/disambiguatingDescription

+1 And we should probably remove the data as soon as this deduplication "project" is over but will be able to re-use the property in future dedup efforts.

TobiasNx:
Deployed on friday. @acka47 you could check this on monday

PR #2187, opened 2025-06-30
Add disambiguatingDescription for deduplication purposes

TobiasNx:
See #2186 and https://github.com/hbz/nwbib/issues/658

I mapped 591 $$a to disambiguatingDescription.

I also added a specific test afterwards to see if it holds the intended info because 591 does appear more often BUT tests do not show the NWBib specific refrence that holds the callNumber of a NZ publication.


dr0i (changes requested):
It's not added to the json-ld context.

acka47 (changes requested):
See my comment

dr0i (approved):

src/main/resources/alma/index-config.json
```diff
@@ -1123,6 +1123,12 @@
             "q.all"
           ]
         },
+        "disambiguatingDescription": {
+          "type": "text",
+          "copy_to": [
+            "q.all"
```
dr0i, line 1129: Are you certain to copy this to the Digibib query? See https://github.com/hbz/lobid-resources/pull/2177#pullrequestreview-2892919288 .

dr0i (changes requested):

src/main/resources/alma/index-config.json
blackwinter: Is this new field somewhow relevant for the union catalogue?

dr0i (approved):

PR #2188, opened 2025-07-03
Use result from sru harvesting of lobid-extra-holdings for DE-Sol1

TobiasNx:
See https://github.com/hbz/lobid-extra-holdings/pull/10
also see https://github.com/hbz/lobid-extra-holdings/issues/3

This utilizes the current result from harvesting DE-SOL1 ZDB holdings records from SRU besides the mapping of the old aleph files as lookup map.

We would then exclude all results from DE-Sol1 bridge, which is frozen, and use the more current mapping until we use a cronjob an make the mapping more dynamic.


TobiasNx:
I updated the files after https://github.com/hbz/lobid-extra-holdings/pull/10#issuecomment-3051537325 and fixed the error hinted by @dr0i .

@dr0i and @acka47  could you have another look

acka47 (approved):
+1

dr0i (approved):

PR #2190, opened 2025-07-08
2173 invalid rvk

TobiasNx:
See #2173

I found the RVK regex here: https://www.wikidata.org/wiki/Property:P1150

Since the broken RVK labels come from the enrichment, I need test data from the RVK Dump. @dr0i could you fetch me the marcXml from RVK aggregations of my two test files?

The marc records would be added to the `<collection>` tag in src/test/resources/cg/aggregate_20240507_example.marcxml


TobiasNx:
I added the old rvk aggregation examples again, that were deleted.

Broken RVK Lables now are kicked out in the workflow generating the culturegraph RVK lookup as well in the lobid transformation if broken rvk are provided there.

TobiasNx:
@hagbeck would this work for you?

hagbeck (approved):
Looks fine :+1:

dr0i (approved):

Issue #2182, opened 2025-06-20
No 912 / 962 in ZDB-OERSI


UBmakla:
Hi,

it seems that the fields 912 or 962 for the library code are missing in the ZDB-OERSI data on lobid.
Could you maybe check on that?

Thanks a lot and best regards,
Manuela

acka47:
As discussed in [today's meeting](https://metadaten.community/t/oersi-und-bibliothekarische-rechercheangebote-in-nrw-juli-2025/809), due to some Alma idiosynchracies, hbz-internal product sigels with "ZDB" prefix are common. As the sigel is thus fix, we can proceed with the implementation.

acka47:
Looks good, around 90k resources now have an OERSI `inCollection` statement: https://lobid.org/resources/search?q=inCollection.id%3A%22https%3A%2F%2Foersi.org%2F%22

Closing.

PR #2193, opened 2025-07-09
2182-oersiCollection

TobiasNx:
Add pseudo zdb collection of hbz to inCollection as OERSI.

See #2182


acka47 (changes requested):
As discussed offline, I'd prefer using `https://oersi.org` as `inCollection.id`.

TobiasNx:
@acka47 could you have another look. I have set a specific collection name for OERSI

PR #2208, opened 2025-09-03
Use lookup from DE-Sol1 holdings including strapi results instead of …

TobiasNx:
…aleph dump

See: https://github.com/hbz/lobid-extra-holdings/issues/5


dr0i:
Will be deployed next Monday.

Issue #2102, opened 2024-11-27
Add URIs for RVK


TobiasNx:
https://www.wikidata.org/wiki/Property:P1150

It seems that RVK has URIs:

http://rvk.uni-regensburg.de/nt/$1

[syntax clarification](https://www.wikidata.org/wiki/Property:P2916)
replace spaces with underscore! (English)

We should add them.

TobiasNx:
I add the RVK labels following @nichtich instructions here: https://metadaten.community/t/rvk-und-uris-fuer-die-notationen/537/4

TobiasNx:
@acka47 could you have a look at the PR.

PR #2210, opened 2025-09-03
Add uri 's for rvk subjects #2102

TobiasNx:
See #2102


src/test/resources/alma-fix/990050000600206441.json
```diff
@@ -97,35 +97,45 @@
     "source" : {
       "label" : "RVK (Regensburger Verbundklassifikation)",
       "id" : "https://d-nb.info/gnd/4449787-8"
-    }
+    },
+    "id" : "http://rvk.uni-regensburg.de/nt/SK%20110",
+    "label" : "SK 110"
```
dr0i, line 102: 1. are those the proper URIs (I ask because they don't yield any result)
2. internally there URLs are redirected to their SSL pendant so why not use https directly?

PR #2215, opened 2025-09-23
Illustrate transformation error

TobiasNx:
Tests: https://github.com/metafacture/metafacture-core/pull/706

There seems to be some kind of error with regard to adding subjects.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,17 +15,17 @@
         <dependency>
             <groupId>org.metafacture</groupId>
             <artifactId>metafacture-io</artifactId>
-            <version>7.0.0</version>
+            <version>feature-moveFixCommandEnumsToIndividualClasses-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.metafacture</groupId>
             <artifactId>metafacture-files</artifactId>
-            <version>7.0.0</version>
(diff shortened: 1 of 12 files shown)


blackwinter:
@fsteeg: Can you maybe give it a try?

`INDEX_SUBPATH_PATTERN` only occurs in `FixPath.matches(String)`, which only occurs in `FixPath.to(value, int)`, which only occurs in `Record.transform(String, UnaryOperator)`. So I would expect any "transform" on an array wildcard (such as `upcase("array.*.value")`) to fail if there are more than 10 elements in the array. But everything I've tried so far always passes (except when some intermediate array element is missing the `value`, but that's a different issue).

So I thought, since you're the one who originally wrote `FixPath.to()` (introduced the `matches()` check in metafacture/metafacture-core@64d8d98), that you may have some insight that I'm missing.

fsteeg:
Something like this seems to cover the issue:

```java
@Test
public void shouldMatchTwoDigitIndexInPath() {
    final String wildcardPath = "field.*.field";
    final String concretePath = "field.10.field";
    final Value value = new Value("").withPathSet(concretePath);
    final FixPath result = new FixPath(wildcardPath).to(value, 15);
    // `value` has a `concretePath` matching the `wildcardPath`, so
    // `result` should be that path, not a path using the passed index:
    Assertions.assertEquals(concretePath, result.toString());
}
```

Issue #2217, opened 2025-09-26
`008` is not used as fallback for `publication.startDate`


aquast:
Recently at FRL we use the lobid element `dateCreated` for extracting the publication year. Unfortunately (for us) the dateCreated value comes from MARC datafield 264 c where you can find a lot of date formats allowed. 
E. g.  https://lobid.org/marcxml/99375707820206441 and https://lobid.org/resources/99375707820206441?format=json

Extracting the Year of publication from this is rather complex and not very robust. 

Since controlfield 008 2 has a controlled 4 digits format for the year, we suggest to provide that information in a new lobid element publicationYear or year Issued. 

Thank you in advance

lschruff:
Hi @acka47 ,

If I understand you right regarding https://github.com/hbz/lobid-resources/issues/2217, we just need to change the "|" at pos 6 of 008 to "s" and  as soon as the lobid record is updated, the startDate should have been created from field 008. Correct?

That would be an easy fix!

Kind regards,
Lara

lschruff:
Thanks so much!

We have about 100 records with a pipe at pos 6 for FRL. I think this was not translated correctly from Aleph to Alma, so it is good, that we found them is way. I will also remind our cataloguers to use the alma template for frl records, where pos 6 is already set at the right value.

aquast:
Sounds as the issue is solved 👍 . 

Thank you all. 

Should I close the issue?

lschruff:
Yes, @TobiasNx . Looks great and also works for FRL 👍

lschruff:
@TobiasNx Hope you can see this comment, although I do not have the permission to re-open the ticket. I discovered two records for which "publication.startDate" has not been formed, despite your changes. I updated both records manually in Alma on 29 September and checked them on lobid ever since.

Could you please have a look at the metadata? Maybe there is more I need to change in Alma to make it work? 
- HT030593412, 99373314450106441
- HT021473613, 99371608114406441

Thanks!

TobiasNx:
I think this is because they have no publication info is assigned so that we can map it to `publication` and  `publication.startDate`:

We map based on:
`264 Ind2=1` -> [1 - Publication](https://www.loc.gov/marc/bibliographic/bd264.html)

 
> * HT030593412, [99373314450106441](https://lobid.org/resources/99373314450106441)

```
<datafield tag="264" ind1=" " ind2="0">
<subfield code="a">Karlsruhe</subfield>
[...]

PR #2218, opened 2025-09-26
Improve fallback for pubdates #2217

TobiasNx:
Se #2217

Adds more types for mapping the pub date in 008 as fallback


TobiasNx:
@fsteeg could you help with regard to the test, something seems off.

acka47 (approved):
+1

Issue #2156, opened 2025-02-18
Weird publication years


TobiasNx:
https://lobid.org/resources/search?q=inCollection.id%3A%22http%3A%2F%2Flobid.org%2Forganisations%2FDE-655%23%21%22+AND+NOT+_exists_%3Asubject&sort=newest

TobiasNx:
@acka47 could you review my related PR.

TobiasNx:
Dates look better, less unrealisitc dates. Therefore closing. We could reopen this with regard to the still open resources with future publication dates:

https://lobid.org/resources/search?sort=newest

PR #2220, opened 2025-09-29
Do not use unrealistic 008 fallback date #2156

TobiasNx:
See #2156 

9999 is a placeholder for publication dates that are still active continues resources. 

Because this date should be deleted as start date.

https://www.loc.gov/marc/bibliographic/bd008a.html

### DIFF
--- a/web/test/tests/IndexIntegrationTest.java
+++ b/web/test/tests/IndexIntegrationTest.java
@@ -33,7 +33,7 @@ public class IndexIntegrationTest extends LocalIndexSetup {
 	public static Collection<Object[]> data() {
 		// @formatter:off
 		return queries(new Object[][] {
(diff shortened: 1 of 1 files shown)


acka47:
I am confused. The start date should be clear for a conintuing resource. I understand the LoC doc you linked to as if the `endDate` should contain "9999", right? 

> Continuing resource currently published
>
> Currently published is defined as an item for which an issue has been received within the last three years.
> 
> 008/07-10 contain the beginning date of publication; 008/11-14 contain the characters 9999.


Isn't the current transformation to `startDate` incorrect then?

acka47 (approved):
Thanks for the explanation, @TobiasNx , I now get the problem and understand the solution. +1

dr0i:
Will be deployed next monday.

Issue #2199, opened 2025-07-28
Add Structured Ownership and Custodial History (MARC 361)


TobiasNx:
See #2198 by @blackwinter 

> Initial groundwork for implementing the new MARC field `361` (Structured Ownership and Custodial History) which is required for the union catalogue (DigiBib; cf. [Jira ticket](https://jira.hbz-nrw.de/browse/DOX-2295)).
> 
> Tentatively modeled after the [MARC format specification](https://www.loc.gov/marc/bibliographic/bd361.html) (see also [Service Wiki: Format](https://service-wiki.hbz-nrw.de/spaces/ALF/pages/1159528455/Vollst%C3%A4ndiges+Feldformat+361+in+Bearbeitung)); handing off to @acka47 / @TobiasNX for proper modeling in accordance with Lobid requirements.
> 
> The search fields are still being discussed (see [Service Wiki: Survey](https://service-wiki.hbz-nrw.de/spaces/ALF/pages/1178271876/Verbundkatalog+und+lobid+-+Anzeige+und+Funktionalitäten+betreffend+361)) and might have to be changed once the survey has concluded.
> 
> Examples:
> 
> <details><summary>HT005003652 (990081807500206441 - https://lobid.org/resources/990081807500206441)</summary>
> 
> ```xml
> <datafield tag="361" ind1=" " ind2=" ">
>   <subfield code="5">DE-51</subfield>
>   <subfield code="s">G 1506a-2</subfield>
>   <subfield code="o">Vorbesitz</subfield>
>   <subfield code="a">Weichert, August</subfield>
>   <subfield code="f">Exlibris</subfield>
>   <subfield code="z">Angabe auf dem Exlibris-Etikett: Ex Bibliotheca Augusti Weichert</subfield>
> </datafield>
> ```
> </details>
> 
> <details><summary>HT005003646 (990081807490206441 - https://lobid.org/resources/990081807490206441)</summary>
> 
> ```xml
> <datafield tag="361" ind1=" " ind2=" ">
>   <subfield code="5">DE-51</subfield>
>   <subfield code="s">G 1506a-1</subfield>
>   <subfield code="o">Vorbesitz</subfield>
>   <subfield code="a">Weichert, August</subfield>
>   <subfield code="f">Exlibris</subfield>
>   <subfield code="z">Angabe auf dem Exlibris-Etikett: Ex Bibliotheca Augusti Weichert</subfield>
> </datafield>
> ```
> </details>
> 
[...]

PR #2232, opened 2025-10-23
Draft transformation for Structured Ownership and Custodial History

TobiasNx:
See #2199

I have drafted two ideas how to model it:

1. Either we create an own property which I here call `hasItemWithProvenanceInformation`:  each object collects the custodial history for a specific publication that is identified by the holding institution `361 $5` and the `361 $s` (https://github.com/hbz/lobid-resources/pull/2232/commits/5e60d7c13db9cd180817b0735c5891058db9c4c4)

2.
	a) Or we move them in `hasItem` currently without identifying the real holding. The object would differ from the others that they have the property  `custodialHistory` (https://github.com/hbz/lobid-resources/pull/2232/commits/226b0cfc25969bae6d674c4b28f38b41f9570f05)

	b) Optionally I could try to identify the correct item by comparing signature AND callNumber if not found create a new item object. (https://github.com/hbz/lobid-resources/pull/2232/commits/2b100b4036d4f142c038d487c109b58a05eff2ae)

----

I would prefer the last version.

Unfortunately none of the example provides `$y - Identifier of the copy described (NR)` which would allow to identify the specific item.

----

At a first quick look it seems that bibframes `custodialHistory` is not usable since it only specifies for a literal. (http://id.loc.gov/ontologies/bibframe/custodialHistory)

I would introduce several new properties:
`hasItemWithProvenanceInformation`, `owner`, `typeOfOwnership`, `proof`, `dateOfOwnership`

### DIFF
--- a/web/test/tests/IndexIntegrationTest.java
+++ b/web/test/tests/IndexIntegrationTest.java
@@ -98,7 +98,7 @@ public class IndexIntegrationTest extends LocalIndexSetup {
 			{ "describedBy.resultOf.object.dateCreated:\"2023-03-22\"", /*->*/ 1},
 			{ "describedBy.resultOf.object.dateModified:\"2023-07-30\"", /*->*/ 3},
 			{ "describedBy.resultOf.object.sourceOrganization.id:\"http\\://lobid.org/organisations/DE-5#\\!\"", /*->*/ 5},
-			{ "describedBy.resultOf.object.modifiedBy.id:\"http\\://lobid.org/organisations/DE-6#\\!\"", /*->*/ 19 },
+			{ "describedBy.resultOf.object.modifiedBy.id:\"http\\://lobid.org/organisations/DE-6#\\!\"", /*->*/ 20 },
 			{ "\"Reader-friendly\"", /*->*/ 1},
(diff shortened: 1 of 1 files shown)


TobiasNx:
@acka47 I found in RDA the following property that relates the item and not "just" the resource to a custodial history, is this a better match?
https://www.rdaregistry.info/Elements/i/#P40026

Also could you have another look with regard to the modelling:

```YAML
custodialHistory:
  - type: Event
    typeOfOwnership:
      - Maps the repetable 361 $o Type of ownership and custodial history information
    proof:
      - Maps the repeatable 361 $f Ownership and custodial history evidence term
    note:
      - maps the repeatble 361  $z - Public note (R)
    contribution:
      - agent:
          id:
          name:
    startdate: formated date $k
    dateStatement: unformated date statement $l
    seeAlso:
      - Link as given in $u
```

Only custodialHistory from rda and two new custom lobid properties `typeOfOwnership` and `prof` would be introduced.

TobiasNx:
@acka47 could you have a last look at this before @blackwinter can review this? I also updated the context as well as the schema.

PR #2240, opened 2025-11-10
Add search field `q.provenance`.

blackwinter:
Complements #2232 by adding a search field for the relevant data (see #2199).

Feel free to drop the `index: false` directives for fields that should be searchable in lobid-resources as well.


acka47:
I think, we should at least support a `_exists_:custodialHistory` query  so that people can fetch all the records with this information and process them locally.

Issue #2231, opened 2025-10-22
Should `subject.componentList.type=work` have `dateOfBirth` or `dateOfDeath`?


TobiasNx:
Compare: http://lobid.org/resources/990070639680206441

```
{
      "type": [
        "ComplexSubject"
      ],
      "label": "Arndt, Johann: Vom wahren Christentum",
      "componentList": [
        {
          "type": [
            "Work"
          ],
          "label": "Arndt, Johann: Vom wahren Christentum",
          "source": {
            "label": "Gemeinsame Normdatei (GND)",
            "id": "https://d-nb.info/gnd/7749153-1"
          },
          "id": "https://d-nb.info/gnd/4246767-6",
          "gndIdentifier": "4246767-6",
          "dateOfBirth": "1555",
          "dateOfDeath": "1621",
          "altLabel": [
            "Arndt, Johann: Vier Bücher vom wahren Christentum",
            "Arndt, Johann: Vier Bücher von wahrem Christentum",
            "Arndt, Johann: 4 Bücher von wahrem Christentum",
            "Arndt, Johann: Vier Bücher von wahrem Christenthumb",
            "Arndt, Johann: 4 Bücher von wahrem Christenthumb",
            "Arndt, Johann: Von wahrem Christentum",
            "Arndt, Johann: Von wahrem Christenthumb/ Heilsamer Busse/ Hertzlicher Rewe und Leid uber die Sünde und wahrem Glauben",
            "Arndt, Johann: Von wahrem Christenthumb",
            "Arndt, Johann: Von wahrem Christenthumb/ Heilsamer Busse"
          ]
        }
      ]
    }
```

Also see: <http://lobid.org/resources/search?q=NOT+subject.componentList.type%3APerson+AND+subject.componentList.dateOfBirth%3A*>

@acka47 what do you think? Compare: https://lobid.org/gnd/4246767-6 and https://portal.dnb.de/opac.htm?method=simpleSearch&cqlMode=true&query=nid%3D4246767-6

PR #2241, opened 2025-11-10
Adjust dates for conferences #2231

TobiasNx:
See #2231 

This PR maps the `startDates` and `endDates` of `Conferences` in `contribution` and `subject`.


acka47:
> But I also could just drop the dates for concerences since they are already part of the label.

+1 for dropping the dates for conferences. If someone asks at any point, we can still add the information.

TobiasNx:
I adjusted the PR and now delete the dates and use a macro for all dates.

acka47 (approved):
+1

Issue #2229, opened 2025-10-21
Add marker or statement for periodicals that are still in active publication


TobiasNx:
Add http://id.loc.gov/ontologies/bibframe/status based on 008/6:

    c - Continuing resource currently published -> http://id.loc.gov/vocabulary/mstatus/current
    d - Continuing resource ceased publication -> http://id.loc.gov/vocabulary/mstatus/ceased
	u - Continuing resource status unknown -> http://id.loc.gov/vocabulary/mstatus/u

Not sure if to map it as toplevel or as part of the first object of publication.

PR #2245, opened 2025-11-13
Add publication.status #2229

TobiasNx:
Resolves: #2229 

This PR introduces a status property for publication that states if a continues resources is current, ceased or unknown mapped on the [basis of 008/06](https://www.loc.gov/marc/bibliographic/bd008a.html)

If okay info schema and context need to be updated.


acka47 (approved):
+1

TobiasNx:
I updated the context and the schema @acka47 could you have another look?

PR #2258, opened 2025-11-28
Dont use list binds for inCollection mappings where not necessary

TobiasNx:
@blackwinter this is not urgent, but does this make sense to improve the performance a little bit?


blackwinter (approved):
My guess is that it should be more performant, yes. But you'd have to verify with your specific workflow and data.

Are you positive that all the conditionals are functioning correctly with arrays/repeated fields - in both positions, field and subfield? I don't recall at the moment, but I may have stumbled upon such an issue in the past.

Also, note that your change would only add a single `inCollection` statement in each case where previously it could have potentially added multiple identical ones (e.g., if there are multiple EZB links). This seems to be a desirable outcome, just wanted to call it out.

Issue #2259, opened 2025-11-28
Update schema


TobiasNx:
TODO: Check the schema, update wrong descriptions, add more requirements, check if WEBAPP has requirements that break the GUI as https://github.com/hbz/lobid-resources/pull/2257 did

@dr0i is there a way to test the current dump without including the test in the transformation?

dr0i:
There is no JSON-LD dump, if that's what you want. There is only an XML dump.

TobiasNx:
PR #2265 is ready for review

PR #2265, opened 2025-12-03
2259-updateSchema

TobiasNx:
Resolves the updating part of #2259 

@acka47 Could you have a look?

Also I have four questions:

- [x] What is the correct way for the `"pattern"` for URIs  sometimes the slashes are escaped sometimes not. (Solution see: https://github.com/hbz/lobid-resources/pull/2265#issuecomment-3665357257)
- [x] Should `related` always have a label? Then we need to fix the transformation too. This currently breaks the validation.

Improvements maybe done in another PR:
- [ ] With regard to hasItem we had the problem of the missing `id` that broke the WebApp with regard to the detail page, do we have any other properties that I have to remember? Can I identify the requirements somehow from the 
- [ ] Should we add `title` and `descriptions` for ALL properties?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# README
+# lobid-resources
 
 ## About
 
@@ -41,10 +41,17 @@ Build lobid-resources:
 Build the web application:
 
 - `cd web`
-- Then follow the script in `web/monit_restart.sh`.
+- `sbt clean`
+- `sbt stage`
+- `./target/universal/stage/bin/lobid-resources-web -no-version-check`
 
 See the `.github/workflows/build.yml` file for details on the CI config
 used by Github Actions.
+
+To run the tests:
+
+- `cd web`
+- `sbt test`
 
 ## Eclipse setup
 
@@ -60,12 +67,6 @@ in Eclipse: `Preferences` \> `General` \> `Workspace` \>
 
 Copy `web/conf/resources.conf_template` to `conf/resources.conf` and
 configure that file to your need.
-
-Use `"start 8000"` to run in production background mode on port 8000
-(hit Ctrl+D to exit logs). To restart a production instance running in
-the background, you can use the included `restart.sh` script (configured
-to use port 8000). For more information, see the [Play
-documentation](https://playframework.com/documentation/2.4.x/Home).
 
 ## Example of getting the data
 
--- a/src/main/resources/alma/alma.fix
+++ b/src/main/resources/alma/alma.fix
@@ -58,6 +58,12 @@ include ("./fix/mediumAndType.fix")
 # mediumAndType.fix is needed ahead of contribution.fix because of dependencies
 include ("./fix/contribution.fix")
 
+do list(path:"alternateGraphicRepresentation[]","var":"$i")
+  unless any_match("$i.script.id","https://unicode.org/iso15924/iso15924.txt#.+")
+    add_field("$i.script.id","https://unicode.org/iso15924/iso15924.txt#Zzzz")
+    add_field("$i.script.label","Kein Schriftcode vergeben")
+  end
+end
 vacuum()
 retain(
   "@context",
--- a/src/test/resources/alma-fix/99374608440706441.json
+++ b/src/test/resources/alma-fix/99374608440706441.json
@@ -22,8 +22,8 @@
     }
   }, {
     "script" : {
-      "id" : "https://unicode.org/iso15924/iso15924.txt#",
-      "label" : "iso15924.txt#"
+      "id" : "https://unicode.org/iso15924/iso15924.txt#Zzzz",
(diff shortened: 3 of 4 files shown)


src/test/resources/schemas/describedBy.json
```diff
@@ -131,7 +133,7 @@
               "additionalProperties": false,
               "properties": {
                 "id": {
-                  "title": "lobid-organisations URI",
+                  "title": "lobid-organisations URI or external URI",
```
acka47: Just curious: Where are non-lobid-organisations URIs used here and for `provider`and `modifiedBy`?

src/test/resources/schemas/description.json
```diff
@@ -17,6 +17,9 @@
         "description": "A string indicating the type of description",
         "type": "string"
       }
-    }
+    },
+    "required": [
+      "label"
```
acka47: Are there cases where the is only a `label` and no `id`? Should there be?

src/test/resources/schemas/inCollection.json
```diff
@@ -21,6 +21,10 @@
       "label": {
         "type": "string"
       }
-    }
+    },
+    "required": [
+      "id",
+      "label"
```
acka47: Didn't we discuss this recently. I think, we should either have `"type": "Collection"` as const for all `inCollection` statements or remove the type statement everywhere for sake of consistency.

src/test/resources/schemas/medium.json
```diff
@@ -47,6 +47,10 @@
           "Sonstige"
         ]
       }
```
acka47: Wasn't there a way to associate corresponding URIs and strings with one another in the JSON Schema? However, it probably isn't important given how we do the transformation.

src/test/resources/schemas/accessRights.json
```diff
@@ -1,16 +1,11 @@
 {
   "$id": "accessRights.json",
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "title": "conditionsOfAccess",
-  "description": "Description if resource is open access.",
+  "title": "Open Access Marker",
+  "description": "This is a marker if resource is open access. It does not state if a resource is not due to missing definitive info in the MARC data.",
```
acka47: Better:

"This is a marker when the resource is open access. It does not state when a resource is not open access due to missing definitive info in the MARC data."

src/test/resources/schemas/alternateGraphicRepresentation.json
```diff
@@ -2,25 +2,33 @@
...
   "$schema": "http://json-schema.org/draft-07/schema#",
   "title": "Record information in other scripture",
+  "description": "The property provides certain properties in additional scriptures.",
   "type": "array",
   "items": {
     "type": "object",
     "additionalProperties": false,
     "properties": {
       "script": {
+        "description": "The additionally provided scripture",
```
acka47: "The scripture in which the additional information is provided"

TobiasNx:
> @acka47 could you have a look at the open questions and make a suggestions what we should add as default label for related, if we always want a label.
> 
> > ```
> > * [ ]  Should `related` always have a label? Then we need to fix the transformation too. This currently breaks the validation.
> > ```
> > 

In order to check for an `label` property in an `related` object if an `id` exists, I updated the version of the json-schemas and used `dependentRequired`

I should be done then! @acka47 could you check my changes?

src/test/resources/schemas/accessRights.json
```diff
@@ -1,16 +1,11 @@
 {
   "$id": "accessRights.json",
-  "$schema": "http://json-schema.org/draft-07/schema#",
-  "title": "conditionsOfAccess",
-  "description": "Description if resource is open access.",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "Open Access Marker",
+  "description": "This is a marker if resource is open access. It does not state if a resource is not due to missing definitive info in the MARC data.",
```
TobiasNx:
```suggestion
  "description": "This is a marker when the resource is open access. It does not state when a resource is not open access due to missing definitive info in the MARC data.",
```

src/test/resources/schemas/alternateGraphicRepresentation.json
TobiasNx: Will use your suggestion.

src/test/resources/schemas/description.json
TobiasNx: You are correct, adjusting the schema.

TobiasNx:
@acka47 I implemented your suggestions and updated the schema version in order to implement some suggestions. Could you have another look?

acka47:
This is looking good. Generally, we should think about adding invalid files in the future against which the schema is also tested.

PR #2266, opened 2025-12-04
Get rid of the etikett maker #2130

### LINKED ISSUE #2130: Replace the fallback mechanism for labels of the ettiket maker
> For now we agreed to only delete the lookup mechanism of the ettiket maker.
> 
> Next steps would be, replace the context creation and the label fallback mechanism.
> 
> For now closing. 

 _Originally posted by @TobiasNx in [#2106](https://github.com/hbz/lobid-resources/issues/2106#issuecomment-2598366837)_

TobiasNx:
Resolves #2130

With regard to improving our schema and introduce a validation check for our data I would suggest to finally get rid of the etikett maker. Default labels should be added by the fix not by a "unnoticed" backround mechanism.


dr0i:
Hm, I don't like how you phrase EtikettMaker as 'unnoticed" backround mechanism.' and cannot see how this improves our data. It's more the other way round: EtikettMaker not only checks (that every `id` has its `label`) but also heals that data (we introduced EtikettMaker back in the days as a customer complained that the json was not parsable for him because he expected a `label` for every `id`).
It's not clear for me if every `id` is guaranteed to have a `label` now. Should we not wait until you will come up with something as a substitution before merging?

PR #2275, opened 2025-12-16
Fix `alternateGraphicRepresentation.publication` mapping #2274

TobiasNx:
Resolves #2274

PR fixes mapping for broken `alternateGraphicRepresentation.publication` mapping.

The example that @blackwinter found has a cataloguing error. @acka47 Should I filter it out or keep it in?

Issue #2274, opened 2025-12-15
Missing `record` subfield in `alternateGraphicRepresentation`.


blackwinter:
In the transformation of alternate graphic representations for publication fields, the intermediate `record` level may be missing in some cases. Ex. [99374608440706441](https://lobid.org/resources/99374608440706441.json):

```json
  "alternateGraphicRepresentation" : [ {
    "script" : {
      "id" : "https://unicode.org/iso15924/iso15924.txt#Jpan",
      "label" : "Japanisch (Kanji, Hiragana und Katakana)"
    },
    "record" : {
      "title" : "村田喜代子",
      "contribution" : [ {
        "agent" : {
          "label" : "村田, 喜代子"
        }
      } ]
    }
  }, {
    "script" : {
      "id" : "https://unicode.org/iso15924/iso15924.txt#",
      "label" : "iso15924.txt#"
    },
    "publication" : [ {
      "location" : [ "東京" ],
      "publishedBy" : [ "新潮社" ]
    } ]
  } ]
```

The latter should actually be:

```json
  {
    "script" : {
      "id" : "https://unicode.org/iso15924/iso15924.txt#",
      "label" : "iso15924.txt#"
    },
    "record": {
      "publication" : [ {
        "location" : [ "東京" ],
        "publishedBy" : [ "新潮社" ]
      } ]
    }
  }
```

This mapping was introduced in 398771d and currently lives in [`macros.fix`](https://github.com/hbz/lobid-resources/blob/8e0f3584b8699e60b09379bf3fe92c291a2cd13f/src/main/resources/alma/fix/macros.fix#L532-L537).

blackwinter:
But this is still wrong, isn't it?

```perl
add_array("alternateGraphicRepresentation[].$last.publication[]")
add_hash( "alternateGraphicRepresentation[].$last.publication[].$append")
add_array("alternateGraphicRepresentation[].$last.publication[].$last.location[]")
copy_field("$880.a", "alternateGraphicRepresentation[].$last.publication[].$last.location[].$append")
add_array("alternateGraphicRepresentation[].$last.publication[].$last.publishedBy[]")
copy_field("$880.b", "alternateGraphicRepresentation[].$last.publication[].$last.publishedBy[].$append")
```

It should be:

```perl
add_array("alternateGraphicRepresentation[].$last.record.publication[]")
add_hash( "alternateGraphicRepresentation[].$last.record.publication[].$append")
[...]

PR #2276, opened 2026-01-05
Remove notations that have no mapped source #2274

TobiasNx:
Fixes https://github.com/hbz/lobid-resources/pull/2275#issuecomment-3661971617


acka47 (approved):
+1

Issue #2248, opened 2025-11-14
Identify Open Access publications


blackwinter:
We have received a [request](https://service-wiki.hbz-nrw.de/pages/diffpagesbyversion.action?pageId=908591140&selectedPageVersions=1065&selectedPageVersions=1064) to mark Open Access publications in the union catalogue. In order to enable this feature in DigiBib, we need to be able to identify such publications via the Lobid metadata.

We have previously implemented it in Limetrans based on [MARC 506](https://www.loc.gov/marc/bibliographic/bd506.html) (first indicator equals `0`; see hbz/limetrans@c5d7c80) and [MARC 856](https://www.loc.gov/marc/bibliographic/bd856.html) (subfield `$7` equals `0`; see hbz/limetrans@c6d7041), which may serve as an inspiration and clarify our requirements.

See also [MARC PROPOSAL NO. 2019-01: Designating Open Access and License Information for Remote Online Resources in the MARC 21 Formats](https://www.loc.gov/marc/mac/2019/2019-01.html).

blackwinter:
> We already have the licensed info from `540` mapped

Are you suggesting that we analyze the licenses and derive the OA status from those?

> With regard to 856 this only makes a statement about the link not about the resource.

This doesn't matter for us. As long as there's at least one OA link, we treat the resource as OA.

TobiasNx:
@acka47 do we want to be able to query for open access

@hagbeck this may be something interesting for your discovery.

PR #2277, opened 2026-01-05
Make accessRights searchable

dr0i:
See #2248.


dr0i:
Will be deployed next Monday.

PR #2283, opened 2026-01-06
Outcomment the metafacture snapshot build #2282

### LINKED ISSUE #2273: Do we need a metafacture-core installation?
I saw that the github action is installing a metafacture-core installation as well as the README demands a metafacture core installation. Is this really needed?

TobiasNx:
Resolves #2273 and #2282

Is outcommenting enough? @dr0i [ I follow your suggestion from here.](https://github.com/hbz/lobid-resources/issues/2273#issuecomment-3645658351)


dr0i (approved):
Yes, outcommenting is fine so we can quickly bring that back if we need it.

PR #2319, opened 2026-03-13
`isPartOf` adjustments

TobiasNx:
This PR adjusts the mapping for `isPartOf`

- adds fallback labels for `isPartOf` relations that have no `hasSuperordinate.id`
- skip creation of an `isPartOf` object for non-HBZ/ZDB-refrencing links if already a `isPartOf` Object exists that links to a lobid resource via (`hasSuperordinate.id`).


acka47 (changes requested):
As commented, some code comments might be confusing and should be adjusted. Otherwise it looks fine.

PR #2335, opened 2026-03-25
Add `inCollection`-object for nrwLastCopies

TobiasNx:
See: https://jira.hbz-nrw.de/browse/MDD-28


src/test/java/TestMinus.java
```diff
@@ -0,0 +1,299 @@
...
+                o.get().literal("type", "TEST");
+                o.get().endEntity();
+                o.get().endRecord();
+            }
+        );
+    }
+
+    @Test
+    public void shouldMinusValueInArray() {
+        MetafixTestHelpers.assertFix(streamReceiver, Arrays.asList(
```
TobiasNx: Not sure If I am able to import `MetafixTestHelpers` since it is not part of `scr/main/` but of `src/test/` from metafix.

src/test/resources/alma-fix/990207214230206441.json
```diff
@@ -234,6 +234,8 @@
     } ],
     "id" : "http://lobid.org/items/990207214230206441:DE-294-21:23106804930006471#!"
   } ],
+  "@holder" : [ "DE-Kn28", "DE-290", "DE-294" ],
```
TobiasNx: Here something weird is happening. The regex should not replace `"@holder" : [ "DE-Kn28", "DE-290", "DE-2941" ]`

When testing this on the playground I cannot recreate this behaviour....

https://metafacture.org/playground/?flux=inputFile%0A%7Copen-file%0A%7Cas-records%0A%7Cdecode-json%0A%7Cfix%28transformationFile%29%0A%7Cencode-yaml%0A%7Cprint%0A%3B&transformation=replace_all%28%22@holder%5B%5D.%2A%22%2C%22%28DE-\\d%2A%29-.%2B%22%2C%22%241%22%29%0A&data=%7B%0A++%22@holder%22+%3A+%5B+%22DE-Kn28%22%2C+%22DE-290%22%2C+%22DE-2941%22+%5D%0A%7D%0A

This changed with: https://github.com/hbz/lobid-resources/pull/2335/changes/188b13dd13043079fec1eae96e5b929a5fd98470

Perhaps @blackwinter encountered this behaviour too?

TobiasNx: Never mind, I must have mixed some commits up. Or I have to investigate this again... hmm weird

TobiasNx: Okay, the incoming values were: 

`   "@holder" : [ "DE-Kn28", "DE-290", "DE-294-21" `

TobiasNx: this seems okay.

TobiasNx: This might have been a bug from VS Code, when showing the diff.......

TobiasNx:
@acka47 Can you check my transformation idea:

- I collect the holder of physical objects,
- filter out your list of isil (https://jira.hbz-nrw.de/browse/MDD-28?focusedId=354176&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-354176)
- change institute libraries to their parents isil
- uniq the resulting list and count the number of holders

When:
- the number of holders is between 1-3,
- the resource is is a NZ and not only IZ or CZ.
- structured pub date from 008 checked if resources is 10 years or older
- but younger than 1851
- Resource is of `type` `Book`, `Map` or `PublishedScore`
- It has `bibliographicLevel.label` "Serial Component Part" or "Monograph/Item"

Then I add an `inCollection` "https://lobid.org/collections#hbzLastCopies"
I currently also add a `note` element. that states the holding institutions as a list. We might drop this.

PS: We should take into account that if resources have suppressed locations these are currently not recognized. We should talk to Verbundgruppe if we need to.

src/main/resources/alma/fix/item.fix
```diff
@@ -228,6 +228,23 @@ do list(path: "hasItem[]","var":"$i")
...
+  if any_contain("$i.type[]","PhysicalObject")
+    copy_field("$i.heldBy.isil","@holder[].$append") # Copy holder ISIL for listing and counting
+  end
+end
+
+replace_all("@holder[].*","(DE-\\d*)-.+","$1") # Subsume libraries to their parent library
+
+uniq("@holder[]")
+
+do list(path: "@holder[]","var":"$i") # Filter out all libraries that are not part of hbz
```
acka47:
```suggestion
do list(path: "@holder[]","var":"$i") # Filter out all libraries that are not in NRW
```

src/main/resources/alma/alma.fix
```diff
@@ -64,6 +64,26 @@ do list(path:"alternateGraphicRepresentation[]","var":"$i")
...
+# Add inCollection for
+# This tests for various fields from different outsourced fixes so that i cannot be placed in
+if exists("@inNZ") # only NZ resources no local IZ resources
+  if any_match("@numberOfHolders[]","[123]") # check if 1-3 publications exists in the Verbund over different libraries
+    if less_than("@standardPubDate","$[todaysYearMinus9]")  # check if resource is 10 years or older
+      if greater_than("@standardPubDate","1851") # check if resource is newer than 1851
+        if any_match("type[]","Book|Map|PublishedScore") # check if resource is of type Book, Map or Published Score
+          if any_match("bibliographicLevel.label","Serial component part|Monograph/Item") # Check for serials component parts or Monographs or Multivolum Book items
+            add_field("inCollection[].$append.id", "https://lobid.org/collections#hbzLastCopies")
+            add_field("inCollection[].$last.label", "hbz Last Copies")
```
acka47:
```suggestion
            add_field("inCollection[].$append.id", "https://lobid.org/collections#nrwLastCopies")
            add_field("inCollection[].$last.label", "nrwLastCopies")
```

acka47:
> I currently also add a `note` element. that states the holding institutions as a list. We might drop this.

We have the problem with the missing level of indirection here, as the `note` refers to the collection and not to the relation. Suggested solution: For now, we will leave this out and only `inCollection` statement. 

In a next step and if people ask about it, we can think about whether to add 1.) a specific top-level property with the ISILs of the last holding libraries or 2.) a note in the `hasItem` objects or something else.

acka47 (changes requested):
See comments.

src/test/java/TestMinus.java
TobiasNx: This is currently not supported by metafacture https://github.com/metafacture/metafacture-core/issues/407

TobiasNx:
@acka47 I adjusted the transformation as you sugested.

@dr0i I added a java class for the the counting `de.hbz.lobid.helper.Minus` due to the [missing portability of the metafacture test helpers](https://github.com/metafacture/metafacture-core/issues/407) I am not able to write a test for this.

acka47 (approved):
+1

TobiasNx:
The actions are not working anymore.

dr0i (approved):

dr0i:
> The actions are not working anymore.

What do you mean ?

TobiasNx:
@dr0i no test needed for the java class then?

and regarding the actions, on friday github seems to have been broken for a while

dr0i:
Re action: along the lines of https://github.com/hbz/lobid-resources/pull/2347
Re tests: no we don't need them neccessarily, as discussed offline, as this repo is no library and the hassle of going through adding tests may not worth the while.

PR #2343, opened 2026-04-27
Update Java Version to 17 in actions

TobiasNx:


TobiasNx:
This is not needed anymore since we updated Java to 21. Closing.

Issue #2346, opened 2026-04-27
Actions are deactivated due to yaml syntax error


TobiasNx:
https://github.com/hbz/lobid-resources/actions/runs/24841034033

```
Invalid workflow file: .github/workflows/build.yml#L5
You have an error in your yaml syntax on line 5
```

The commit https://github.com/hbz/lobid-resources/commit/a1b7b0404436cfc1998dbe3061545bee81d58efa has deleted the comment but it breaks the yaml syntax because of wrong indentation.

TobiasNx:
@dr0i we can close this, was solved with 61978730f99525ea2e229ea2bb0238f82c3981dd 

So closing. Next time we use the official way with PR and Review.

PR #2347, opened 2026-04-27
Revert "Fix indentation"

TobiasNx:
This reverts commit 61978730f99525ea2e229ea2bb0238f82c3981dd.

This quickfix had no documentation. In order to properly document the problem #2346.
 I suggest to revert this ~~PR~~ commit and redo the commits with proper documentation.


dr0i:
I still don't understand why you want to revert the commit, as it is a proper fix (although it's not sufficient to make GitHub actions run again). So we need some more fixing and thus commits and we still can reference in these commits the ones involved and this PR here. That should be sufficient to have 1. documentation and 2. room for discussing the problem.

PR #2349, opened 2026-04-27
Fix broken action #2246

TobiasNx:


TobiasNx:
@dr0i github copilot explaination is the jackson version incompatibility 2.12 and 2.14 in the pom. When I changed this it still was running into an error.

TobiasNx:
Replaced by #2351

Issue #2350, opened 2026-04-28
Use of Metafacture master-SNAPSHOT breaks project tests and web tests


TobiasNx:
I locally published the current Metafacture-Core master-SNAPSHOT to use it in lobid-resources. This seems to be an additional problem with regard to the failing actions #2348 . The current master of metafacture-core has multiple updates with regard to dependencies that create transitive problems.

```
$ mvn clean install -DskipTests=false -DgenerateTestData=true
[INFO] Scanning for projects...
[INFO] 
[INFO] ---------------------< org.lobid:lobid-resources >----------------------
[INFO] Building lobid-resources 1.0.2-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
Downloading from central: https://repo.maven.apache.org/maven2/commons-io/commons-io/2.21.0/commons-io-2.21.0.pom
Downloaded from central: https://repo.maven.apache.org/maven2/commons-io/commons-io/2.21.0/commons-io-2.21.0.pom (21 kB at 32 kB/s)
Downloading from central: https://repo.maven.apache.org/maven2/org/apache/commons/commons-parent/91/commons-parent-91.pom
Downloaded from central: https://repo.maven.apache.org/maven2/org/apache/commons/commons-parent/91/commons-parent-91.pom (79 kB at 488 kB/s)
Downloading from central: https://repo.maven.apache.org/maven2/org/apache/apache/35/apache-35.pom
Downloaded from central: https://repo.maven.apache.org/maven2/org/apache/apache/35/apache-35.pom (24 kB at 238 kB/s)
Downloading from central: https://repo.maven.apache.org/maven2/org/junit/junit-bom/5.13.4/junit-bom-5.13.4.pom
Downloaded from central: https://repo.maven.apache.org/maven2/org/junit/junit-bom/5.13.4/junit-bom-5.13.4.pom (5.7 kB at 62 kB/s)
Downloading from central: https://repo.maven.apache.org/maven2/org/apache/commons/commons-compress/1.28.0/commons-compress-1.28.0.pom
Downloaded from central: https://repo.maven.apache.org/maven2/org/apache/commons/commons-compress/1.28.0/commons-compress-1.28.0.pom (25 kB at 288 kB/s)
Downloading from central: https://repo.maven.apache.org/maven2/org/apache/commons/commons-parent/85/commons-parent-85.pom
[...]

Issue #2348, opened 2026-04-27
Action does not work with JAVA 11 anymore


TobiasNx:
After fixing #2346 it shows that the action fail due to the wrong Java Version.

TobiasNx:
@dr0i also it seems that the dependency updates from Metafacture-Core master (see: https://github.com/metafacture/metafacture-core/issues/426#issuecomment-4333902912) create problems due to transitive dependencies.

TobiasNx:
I tested around how to fix the issue and followed some suggestion from copilot explaining the failing actions but the errors are piling up: https://github.com/hbz/lobid-resources/pull/2349

TobiasNx:
This is resolved with #2351 and #2352

PR #2352, opened 2026-05-07
Fixes to make the current mf master runnable

TobiasNx:
See #2350 and #2348

This PR harmonizes the Jackson versions as well as enforces guice 6.0.0 as suggested by @fsteeg 

This would also fix the problems with the current metafacture master.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,8 +23,8 @@ jobs:
         cd ..
         git clone https://github.com/metafacture/metafacture-core.git
         cd metafacture-core
-        git checkout -b 7a64c159a2e2dfc7d1673c83699254ad148f4011
-        git reset --hard 7a64c159a2e2dfc7d1673c83699254ad148f4011
+        git checkout -b f2cf79da098258a2bc034da6bf7178f704fb5db4
+        git reset --hard f2cf79da098258a2bc034da6bf7178f704fb5db4
         ./gradlew publishToMavenLocal
         cd -
     - name: Build with Maven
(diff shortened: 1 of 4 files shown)


dr0i:
I don't get why to 55c3120fe044df36a50a3f1f8185458f90ee8fce is needed. Resetting to 5b0726f085ca16e001d2de46982d916f968068c2 makes it work all right or have I missed something?

TobiasNx:
Both commits are needed:
- https://github.com/hbz/lobid-resources/commit/5b0726f085ca16e001d2de46982d916f968068c2 to solve the issues with the deprecated method from `Jackson`
- https://github.com/hbz/lobid-resources/commit/5b0726f085ca16e001d2de46982d916f968068c2 is needed for the errors with `sbt` :  `NoClassDefFoundError: javax/inject/Provider`. The solution was described by @fsteeg [here](https://github.com/metafacture/metafacture-core/issues/426#issuecomment-4343827471)

I bump the metafacture version to the current commit to illustrate improvement and that we do not encounter the previous issues when building `mvn` or `sbt` and that it resolves our isues from #2350 

PS: The action breaks currently since github does not reach maven repositories.

dr0i:
> Resetting to https://github.com/hbz/lobid-resources/commit/5b0726f085ca16e001d2de46982d916f968068c2 makes it work all right or have I missed something?

Ah, now I understand why this worked: you hadn't set the mf-dependency to the current master.
When github actions builds successfully I will review.

TobiasNx:
Yes, I moved the PR not to the current master but the last commit of the master. Actions work now.

dr0i (approved):

dr0i:
Deployed. Closed.

PR #2356, opened 2026-05-20
Separate deployment from monitoring

Phu2:
Currently, we use Monit for both deployment and monitoring the process lifecycle including automatic restarts in one step, see the local `/etc/monit/conf.d/play-instances.rc` and https://github.com/hbz/lobid-resources/blob/master/web/monit_restart.sh plus https://github.com/hbz/lobid-resources/blob/master/web/restart.sh respectively.

This PR introduces an alternative setup in addition to the current approach. You can use one of the two options, not both at the same time.

- the app runs as an systemd service
- deployment is executed by a separate script
- Monit only watches the process and if failed will restart via systemctl


fsteeg:
Hm, so lobid-resources is a bit of a tricky special case when it comes to our Play apps, since the repo contains both the transformation (root folder, Maven build) and the Play app (web folder, SBT build). As far as I understand it, this PR is only about the Play app? Some question / thoughts to understand this better:

- When / where will the deploy.sh script be called?
- Should we move the deploy.sh script to `web/` to make its scope clearer?
- If we deploy to a clean location, `git reset` makes sense, but if we run in place, in the repo folder, we should probably `merge`, since in the old setup relies on local config changes.
- Could this interfere with the weekly deployment setup that exists for lobid-resources (transformation and Play app, but not sure about the details)?

TobiasNx:
Additionally one should consider that the transformation should only be automatically deployed once since changes during the week in the transformation might lead to incompatible and inconsistent data in the index which might lead to breaking lobid-resources web app. 
The deployment is currently timed that the full publishing from ALMA is the first transformation that uses the new changes and all consecutive updates are using the same transformation to ensure consistent data.

Currently manual deployment is used if changes are urgent and a full publishing transformation is also triggered by @dr0i and all current updates too.

Phu2:
We must also take these changes https://github.com/hbz/lobid-resources/pull/2401 into account.

Phu2:
> When / where will the deploy.sh script be called?

You can execute this manually if needed. Or it could be triggered by some other script or event. 

So far, deployment is executed by `monit_restart.sh`. It runs when the application is stopped, for whatever reason. I can't see a specific time for the deployment.

EDIT: as @TobiasNx stated above

> The deployment is currently timed that the full publishing from ALMA is the first transformation that uses the new changes and all consecutive updates are using the same transformation to ensure consistent data.

How exactly is the deployment triggered?

Phu2:
> * Should we move the deploy.sh script to `web/` to make its scope clearer?

Good idea.

web/deploy.sh
```diff
@@ -0,0 +1,110 @@
...
+BASE_DIR="${HOME}/git/${REPO}"
+WEB_DIR="$BASE_DIR/web"
+
+LOG_FILE="${WEB_DIR}/logs/deploy.log"
+
+SERVICE_NAME="${REPO}.service"
+
+# it is important to set the proper locale
+. $HOME/.locale
+export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64/
```
dr0i, line 26: This is outdated, we switched to Java21. With it some other things come, too.

web/deploy.sh
```diff
@@ -0,0 +1,110 @@
...
+
+# Ensure the script exits immediately if any command fails, and that all variables are defined before use.
+# The 'pipefail' option ensures that if any command in a pipeline fails, the entire pipeline is considered to have failed.
+set -euo pipefail
+
+##
+## Configuration
+##
+
+REPO=$1 # "lobid-resources-rpb-test"
```
dr0i, line 13: maybe rename the script "deploy.sh" to "deploy-lobid-resources-rpb-test.sh" if it's about lobid-resources-rpb-test? Otherwise one might be confused when it's coming to deploy lobid-resources ...

Phu2:
> * If we deploy to a clean location, `git reset` makes sense, but if we run in place, in the repo folder, we should probably `merge`, since in the old setup relies on local config changes.

Not sure about this. Is this a new requirement? I didn't change the deployment process, we've always `git reset` so far. However, i wondered about that, too. See my comment https://github.com/hbz/lobid-resources/pull/2401#issuecomment-5538279864

Phu2:
> * Could this interfere with the weekly deployment setup that exists for lobid-resources (transformation and Play app, but not sure about the details)?

That's beyond my understanding of the processes involved. @dr0i please help.

dr0i:
> * since in the old setup relies on local config changes.

it does not - as far as it concerns `resources.conf` (and afaik this is the only config one changes): in the repo exists only `resources.conf_template` which is then to be copied at the server to `resources.conf` - which is so never touched by git. It's very stable :)

web/deploy.sh
Phu2: Thanks, see my comment above https://github.com/hbz/lobid-resources/pull/2356#issuecomment-5541423500

web/deploy.sh
Phu2: Well, this script is meant to be generic. You can use it, but it's not required. As  i stated above:

> This PR introduces an alternative setup in addition to the current approach. You can use one of the two options, not both at the same time.

So yes, it might be confusing...

dr0i:
> So far, deployment is executed by monit_restart.sh. It runs when the application is stopped, for whatever reason. I can't see a specific time for the deployment.

Automatic Friday deployment is scheduled on our cron server, as usually.

dr0i:
> > * Could this interfere with the weekly deployment setup that exists for lobid-resources (transformation and Play app, but not sure about the details)?
> 
> That's beyond my understanding of the processes involved. @dr0i please help.

idk either. As @Phu2 stated, the new script is optional. afaik we won't use it when deploying lobid-resources. It's used for a "fork" or something. So I think we can just ignore it. That said - wouldnt' this be better put in an other place? It's a bit confusing ...

web/deploy.sh
dr0i: ok then could this comment just be deleted? Or am I overreacting here? I mean, this repo has nothing to with rpb and things should (and could?) be on the proper shelf...

PR #2358, opened 2026-06-02
Only remove duplicate tokens in the same position for DigiBib search fields.

blackwinter:
Fixes searching for repeated tokens in a search field (e.g., author "Liu, Liu"; reported in [DOX-2434](https://jira.hbz-nrw.de/browse/DOX-2434)) by only removing duplicate tokens that are in the same position.

The [`unique` token filter](https://www.elastic.co/guide/en/elasticsearch/reference/2.2/analysis-unique-tokenfilter.html) was originally meant to remove duplicate tokens resulting from preceding filters, but appears to be overzealous with regard to explicitly repeated tokens in the search query or the source document (such as in titles or author names). Instead, use the [filter option](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/analysis-unique-tokenfilter.html) `only_on_same_position: true` (the equivalent [`remove_duplicates` token filter](https://www.elastic.co/guide/en/elasticsearch/reference/6.4/analysis-remove-duplicates-tokenfilter.html) is only available since Elasticsearch 6.4).

Before:

```json
{
  "tokens": [
    {
      "token": "liu",
      "start_offset": 1,
      "end_offset": 4,
      "type": "<ALPHANUM>",
      "position": 0
    }
  ]
}
```

After:

```json
{
  "tokens": [
    {
      "token": "liu",
      "start_offset": 1,
      "end_offset": 4,
      "type": "<ALPHANUM>",
      "position": 0
    },
    {
      "token": "liu",
      "start_offset": 6,
      "end_offset": 9,
      "type": "<ALPHANUM>",
      "position": 1
    }
  ]
}
```

### DIFF
--- a/web/test/tests/IndexIntegrationTest.java
+++ b/web/test/tests/IndexIntegrationTest.java
@@ -151,6 +151,8 @@ public class IndexIntegrationTest extends LocalIndexSetup {
 			{ "q.all:HT072067630", /*->*/ 0 },
 			{ "q.all:(Erleben \\- Verstehen & Lernen)", /*->*/ 5 },
 			{ "q.all:(Lexicography \\: Selected Papers)", /*->*/ 1 },
+			{ "q.all:(Lernen Lernen)", /*->*/ 4 },
+			{ "q.all:\"Lernen Lernen\"", /*->*/ 4 },
 			{ "contribution.agent.label.digibib:Westfalen", /*->*/ 5 },
 			{ "contribution.agent.label.digibib:Westfälen", /*->*/ 5 },
 			{ "contribution.agent.label.digibib:Westfälisch", /*->*/ 5 },
--- a/web/test/tests/IndexIntegrationTest.java
+++ b/web/test/tests/IndexIntegrationTest.java
@@ -152,7 +152,7 @@ public class IndexIntegrationTest extends LocalIndexSetup {
 			{ "q.all:(Erleben \\- Verstehen & Lernen)", /*->*/ 5 },
 			{ "q.all:(Lexicography \\: Selected Papers)", /*->*/ 1 },
 			{ "q.all:(Lernen Lernen)", /*->*/ 4 },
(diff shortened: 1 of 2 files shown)


web/test/tests/IndexIntegrationTest.java
```diff
@@ -151,6 +151,8 @@ public static Collection<Object[]> data() {
 			{ "q.all:HT072067630", /*->*/ 0 },
 			{ "q.all:(Erleben \\- Verstehen & Lernen)", /*->*/ 5 },
 			{ "q.all:(Lexicography \\: Selected Papers)", /*->*/ 1 },
+			{ "q.all:(Lernen Lernen)", /*->*/ 4 },
```
blackwinter, line 154: I don't understand why this query still returns all four hits... 😕

dr0i: You only specified field `contribution.label` on index level in index-config.json

dr0i (approved):

web/test/tests/IndexIntegrationTest.java
blackwinter: Can you please elaborate? The search field `q.all` uses the DigiBib analyzers. Before this change, the above query would match four documents because it effectively searches for the single token "Lernen" only. After this change, it should search for both the repeated tokens "Lernen" and "Lernen" and only match a single document (`99375631628906441`: `"title" : "Lerntext und Aufgabe zum Thema Lernen: Lernen - Begriffsdefinition"`). It works for the phrase query (next line), but not this one.

dr0i: I was a bit quick - a hip-shot. I've now tested and tried a bit around, don't know also why your solution works fine with the phrase but not without it.

blackwinter: Okay, thanks. Then let's just wait and see what happens...

PR #2360, opened 2026-06-03
Use the new marc-relators map #2359

### LINKED ISSUE #2359: Use new lobid-vocabs marc-relator map
Related to https://github.com/hbz/lookup-tables/pull/65

@blackwinter has added a new central list for marc-relators. We should use that instead of our local one. 

I will also use this issue to test the created map by @blackwinter

TobiasNx:
Resolves #2359

Merge after https://github.com/hbz/lookup-tables/pull/65 is merged.


dr0i:
What I don't like about this process:
before we have tested maps that are uses in production. Now we will have a test map that's only used for tests (like in l'art-pour-l'art).

TobiasNx:
I do not understand your comment: For all dynamic maps that are collected from somewhere we have the static test map, which is a versin of the map, and we add them to the test folder. We did that in the same way when introducing new maps in previous PRs.

dr0i (approved):

dr0i:
I am just speaking about this PR and that I liked the way of doing the test before more than now.

Issue #2363, opened 2026-06-12
Wrong ISIL `DE-Kob 7` in 361 maps wrong and breaks UI


TobiasNx:
Broken UI: http://lobid.org/resources/99376129189706441#!

https://lobid.org/marcxml/99376129189706441

```
<datafield tag="361" ind1="1" ind2=" ">
<subfield code="a">Riemenschneider, Dieter</subfield>
<subfield code="f">Etikett</subfield>
<subfield code="l">2025</subfield>
<subfield code="s">MAG 54682</subfield>
<subfield code="u">
https://primo.uni-koblenz.de/discovery/collectionDiscovery?vid=49HBZ_UKO:VU1&collectionId=8159892120008057&lang=de
</subfield>
<subfield code="0">(DE-588)121481158</subfield>
<subfield code="5">DE-Kob 7</subfield>
</datafield>
```

http://lobid.org/resources/99376129189706441.json
```
"hasItem": [
    {
      "label": "lobid Bestandsressource",
      "type": [
        "Item",
        "PhysicalObject"
      ],
      "callNumber": "MAG 54682",
      "inventoryNumber": "K2025/3409",
      "serialNumber": "M37 8926",
      "currentLibrary": "RL001",
      "currentLocation": "MAGKO",
      "heldBy": {
        "isil": "DE-Kob7",
        "id": ["http://lobid.org/organisations/DE-Kob7#!"](http://lobid.org/organisations/DE-Kob7#!),
        "label": "Universitätsbibliothek Koblenz"
      },
      "seeAlso": [
       [ "https://primo.uni-koblenz.de/permalink/49HBZ_UKO/16pslu1/alma99376129189706441"](https://primo.uni-koblenz.de/permalink/49HBZ_UKO/16pslu1/alma99376129189706441)
      ],
      "inCollection": [
        {
          "id": ["http://lobid.org/organisations/DE-Kob7#!"](http://lobid.org/organisations/DE-Kob7#!),
          "label": "Universitätsbibliothek Koblenz",
          "type": [
            "Collection"
          ]
        }
      ],
      "id": ["http://lobid.org/items/99376129189706441:DE-Kob7:2355738760008057#!"](http://lobid.org/items/99376129189706441:DE-Kob7:2355738760008057#!)
    },
    {
      "type": [
        "Item"
      ],
      "heldBy": {
[...]

PR #2364, opened 2026-06-16
Improve ISIL regex #2363

TobiasNx:
It did not catch DE-Kob 7 as an error. The Regex is from wikidata.


TobiasNx:
@acka47 can you help, what is failing with my regex?

acka47:
Why not, for the concrete use case, just reject white spaces, e.g. like this: `^[A-Z]{2}-.*[^\\s]`?

TobiasNx:
I took the regex for ISIL from Wikidata: https://www.wikidata.org/wiki/Property:P791

I am not sure if the problem was `ajv` because with `jsonschema` the regex does not create any problems.

I adjusted the validation process, now it seems to work.

In order to illustrate I temporarily commit a new test file:  https://github.com/hbz/lobid-resources/actions/runs/27640190543/job/81737863069?pr=2364#step:8:15

This test commit was reverted.

acka47 (approved):
+1

acka47:
I tested the Wikidata regex again with https://regex101.com/ . The slash before the closing square brackets looks wrong. It does work without it, like `\D{1,4}\-[A-Za-z\-0-9:]{1,11}`

TobiasNx:
> I tested the Wikidata regex again with https://regex101.com/ . The slash before the closing square brackets looks wrong. It does work without it, like `\D{1,4}\-[A-Za-z\-0-9:]{1,11}`

the error complaint that `\D` is not allowed seems to be due to `ajv`. the `/` at the end of the bracket only shows one sign that is allowed in an ISIL. Moving from `ajv` to `jsonschema` solved the issue.

Also the Wikidata regex fits the examples in the wikipedia page: https://de.wikipedia.org/wiki/Bibliothekssigel

TobiasNx:
@dr0i do you want to have a look to since I adjusted the actions?

dr0i (approved):

PR #2371, opened 2026-06-18
Continue validation and count invalid records #2370

### LINKED ISSUE #2370: Log all validation errors from updates
We should add the option `--continue` so that all errors in the update dump are reported. Currently only one single error in the first record is reported.

Additionally it would be great if the subject would include the number of invalid records would be counted.

TobiasNx:
The validation should continue and count the number of invalid file.

### DIFF
--- a/web/scripts/getUpdatesDumpJsonl.sh
+++ b/web/scripts/getUpdatesDumpJsonl.sh
@@ -79,7 +79,7 @@ cd ~/git/lobid-resources/src/test/resources/schemas
 jsonschema validate resource.json ${DUMP_DIRECTORY}${UPDATES_FNAME} --continue > /tmp/jsonschemaValidationOutput.log 2>&1
 
 if [ -s /tmp/jsonschemaValidationOutput.log ]; then
-        NUMBER_OF_INVALID_RECORDS = $(grep -c "fail:" /tmp/jsonschemaValidationOutput.log)
+        NUMBER_OF_INVALID_RECORDS=$(grep -c "fail:" /tmp/jsonschemaValidationOutput.log)
         head -n5 /tmp/jsonschemaValidationOutput.log | mail -s "[sysad] [lobid-resources] Jsonschema validated with ${NUMBER_OF_INVALID_RECORDS} invalid records" "${MAIL_TO}" -a "From: ${MAIL_FROM}"
 fi
 rm /tmp/jsonschemaValidationOutput.log
--- a/web/scripts/getUpdatesDumpJsonl.sh
+++ b/web/scripts/getUpdatesDumpJsonl.sh
@@ -79,7 +79,7 @@ cd ~/git/lobid-resources/src/test/resources/schemas
 jsonschema validate resource.json ${DUMP_DIRECTORY}${UPDATES_FNAME} --continue > /tmp/jsonschemaValidationOutput.log 2>&1
 
 if [ -s /tmp/jsonschemaValidationOutput.log ]; then
-        NUMBER_OF_INVALID_RECORDS = $(grep -c "fail:")
+        NUMBER_OF_INVALID_RECORDS = $(grep -c "fail:" /tmp/jsonschemaValidationOutput.log)
         mail -s "[sysad] [lobid-resources] Jsonschema validated with ${NUMBER_OF_INVALID_RECORDS} invalid records" "${MAIL_TO}" -a "From: ${MAIL_FROM}" < "/tmp/jsonschemaValidationOutput.log"
 fi
 rm /tmp/jsonschemaValidationOutput.log
--- a/web/scripts/getUpdatesDumpJsonl.sh
+++ b/web/scripts/getUpdatesDumpJsonl.sh
@@ -80,6 +80,6 @@ jsonschema validate resource.json ${DUMP_DIRECTORY}${UPDATES_FNAME} --continue >
 
 if [ -s /tmp/jsonschemaValidationOutput.log ]; then
         NUMBER_OF_INVALID_RECORDS=$(grep -c "fail:" /tmp/jsonschemaValidationOutput.log)
-        head -n5 /tmp/jsonschemaValidationOutput.log | mail -s "[sysad] [lobid-resources] Jsonschema validated with ${NUMBER_OF_INVALID_RECORDS} invalid records" "${MAIL_TO}" -a "From: ${MAIL_FROM}"
+        head -n333 /tmp/jsonschemaValidationOutput.log | mail -s "[sysad] [lobid-resources] Jsonschema validated with ${NUMBER_OF_INVALID_RECORDS} invalid records" "${MAIL_TO}" -a "From: ${MAIL_FROM}"
 fi
 rm /tmp/jsonschemaValidationOutput.log
--- a/web/scripts/getUpdatesDumpJsonl.sh
+++ b/web/scripts/getUpdatesDumpJsonl.sh
@@ -80,6 +80,6 @@ jsonschema validate resource.json ${DUMP_DIRECTORY}${UPDATES_FNAME} --continue >
 
 if [ -s /tmp/jsonschemaValidationOutput.log ]; then
         NUMBER_OF_INVALID_RECORDS = $(grep -c "fail:" /tmp/jsonschemaValidationOutput.log)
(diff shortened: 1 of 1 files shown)


web/scripts/getUpdatesDumpJsonl.sh
```diff
@@ -76,9 +76,10 @@ EOF
 
 # validate schema
 cd ~/git/lobid-resources/src/test/resources/schemas
-jsonschema validate resource.json ${DUMP_DIRECTORY}${UPDATES_FNAME} > /tmp/jsonschemaValidationOutput.log 2>&1
+jsonschema validate resource.json ${DUMP_DIRECTORY}${UPDATES_FNAME} --continue > /tmp/jsonschemaValidationOutput.log 2>&1
 
 if [ -s /tmp/jsonschemaValidationOutput.log ]; then
-        mail -s "[sysad] [lobid-resources] Jsonschema validated with errors" "${MAIL_TO}" -a "From: ${MAIL_FROM}" < "/tmp/jsonschemaValidationOutput.log"
+        NUMBER_OF_INVALID_RECORDS = $(grep -c "fail:")
```
dr0i: this won't work for the logfile is missing

dr0i (changes requested):

web/scripts/getUpdatesDumpJsonl.sh
TobiasNx:
```suggestion
        NUMBER_OF_INVALID_RECORDS = $(grep -c "fail:" /tmp/jsonschemaValidationOutput.log)
```
you are right, like this?

dr0i: yep

web/scripts/getUpdatesDumpJsonl.sh
```diff
@@ -76,9 +76,10 @@ EOF
 
 # validate schema
 cd ~/git/lobid-resources/src/test/resources/schemas
-jsonschema validate resource.json ${DUMP_DIRECTORY}${UPDATES_FNAME} > /tmp/jsonschemaValidationOutput.log 2>&1
+jsonschema validate resource.json ${DUMP_DIRECTORY}${UPDATES_FNAME} --continue > /tmp/jsonschemaValidationOutput.log 2>&1
 
 if [ -s /tmp/jsonschemaValidationOutput.log ]; then
-        mail -s "[sysad] [lobid-resources] Jsonschema validated with errors" "${MAIL_TO}" -a "From: ${MAIL_FROM}" < "/tmp/jsonschemaValidationOutput.log"
+        NUMBER_OF_INVALID_RECORDS = $(grep -c "fail:" /tmp/jsonschemaValidationOutput.log)
+        mail -s "[sysad] [lobid-resources] Jsonschema validated with ${NUMBER_OF_INVALID_RECORDS} invalid records" "${MAIL_TO}" -a "From: ${MAIL_FROM}" < "/tmp/jsonschemaValidationOutput.log"
```
dr0i:
```suggestion
        mail -s "[sysad] [lobid-resources] Jsonschema validated with ${NUMBER_OF_INVALID_RECORDS} invalid records" "${MAIL_TO}" -a "From: ${MAIL_FROM}" < "head -n2 /tmp/jsonschemaValidationOutput.log"
```

dr0i: With my suggestion max. 2 docs are part of the mail. Otherwise, in a worst case, you could have a GB-sized mail.

TobiasNx: Should we up this to 5? But i am also okay with 2

dr0i: 5 is also good :+1:

TobiasNx:
```suggestion
        mail -s "[sysad] [lobid-resources] Jsonschema validated with ${NUMBER_OF_INVALID_RECORDS} invalid records" "${MAIL_TO}" -a "From: ${MAIL_FROM}" < "head -n5 /tmp/jsonschemaValidationOutput.log"
```

dr0i (approved):

web/scripts/getUpdatesDumpJsonl.sh
```diff
@@ -76,9 +76,10 @@ EOF
 
 # validate schema
 cd ~/git/lobid-resources/src/test/resources/schemas
-jsonschema validate resource.json ${DUMP_DIRECTORY}${UPDATES_FNAME} > /tmp/jsonschemaValidationOutput.log 2>&1
+jsonschema validate resource.json ${DUMP_DIRECTORY}${UPDATES_FNAME} --continue > /tmp/jsonschemaValidationOutput.log 2>&1
 
 if [ -s /tmp/jsonschemaValidationOutput.log ]; then
-        mail -s "[sysad] [lobid-resources] Jsonschema validated with errors" "${MAIL_TO}" -a "From: ${MAIL_FROM}" < "/tmp/jsonschemaValidationOutput.log"
+        NUMBER_OF_INVALID_RECORDS = $(grep -c "fail:" /tmp/jsonschemaValidationOutput.log)
```
blackwinter: Bash does not allow whitespace around `=`.

```suggestion
        NUMBER_OF_INVALID_RECORDS=$(grep -c "fail:" /tmp/jsonschemaValidationOutput.log)
```

web/scripts/getUpdatesDumpJsonl.sh
```diff
@@ -76,9 +76,10 @@ EOF
...
+        mail -s "[sysad] [lobid-resources] Jsonschema validated with ${NUMBER_OF_INVALID_RECORDS} invalid records" "${MAIL_TO}" -a "From: ${MAIL_FROM}" < "head -n5 /tmp/jsonschemaValidationOutput.log"
```
blackwinter: You're reading from a file named `head -n5 /tmp/jsonschemaValidationOutput.log` - which does not exist!

```suggestion
        head -n5 /tmp/jsonschemaValidationOutput.log | mail -s "[sysad] [lobid-resources] Jsonschema validated with ${NUMBER_OF_INVALID_RECORDS} invalid records" "${MAIL_TO}" -a "From: ${MAIL_FROM}"
```

dr0i: yeah, properly correct - always good to at least run one time the code

TobiasNx:
@dr0i do you want to have another look?

dr0i:
I would test and apply @blackwinter suggestions if it was my code in the first plac. Or am I mistaken?

TobiasNx:
@dr0i  thanks I missed the suggestion from @blackwinter but how can I test this locally?

dr0i:
Test it by adapting like:
```
UPDATES_FNAME=2026-06-15_to_2026-06-16_lobid-resources-updates.jsonl.gz

# validate schema
cd ~/git/lobid-resources/src/test/resources/schemas

jsonschema validate resource.json ${UPDATES_FNAME} --continue > /tmp/jsonschemaValidationOutput.log 2>&1

if [ -s /tmp/jsonschemaValidationOutput.log ]; then
        NUMBER_OF_INVALID_RECORDS=$(grep -c "fail:" /tmp/jsonschemaValidationOutput.log)
        echo $NUMBER_OF_INVALID_RECORDS
        head -n5 /tmp/jsonschemaValidationOutput.log
fi
```
I've used a dump which was reported as flawed (see mails) and downloaded it from https://lobid.org/download/dumps/lobid-resources/. 
(I've probably an old version of jsonschema locally because it outputs:
> /usr/bin/jsonschema:5: DeprecationWarning: The jsonschema CLI is deprecated and will be removed in a future version. Please use check-jsonschema instead, which can be installed from https://pypi.org/project/check-jsonschema/ )

Tested this then on server.
Remarkably, it found only 1 error (don't know if this is coincidence or just bailing out and thus not working as expected).
However, we may just deploy the script and see what happens in the days to come.

Note also that `head -n5` results in just a tiny bit of information - the ressource is not dumped in one line but is pretty printed, so we have "one field, one line". Thus I will increase this to 333 and, as usually, will adapt this when we see results.

dr0i:
Script found 14 invalid resources today - so counting works :)

TobiasNx:
Thats great!. But it seems that the mail only includes one not five errors, is that correct?

dr0i:
Yep, see[ comment above](https://github.com/hbz/lobid-resources/pull/2371#issuecomment-4832543429):
> Note also that head -n5 results in just a tiny bit of information - the ressource is not dumped in one line but is pretty printed, so we have "one field, one line". Thus I will increase this to 333 and, as usually, will adapt this when we see results.

If you want more, we could increase to 2k lines? This doesn't _guarantee_ 5 docs, though.

Issue #2365, opened 2026-06-17
Add other Verbund-Ids


TobiasNx:
In order to harvest Schlagwortfolgen from BSZ we need their record identifiers. So I suggest to add identifiers for all Verbundkataloge as well as GVI and ekz.

- BVB
- GBV
- hebis
- KOBV
- SWB
- VÖBB
- ekz
- ÖBV

PS: maybe elisa id

acka47:
I am not sure I understand the need for this issue. The agreed on plan is, in the next stept to harvest Schlagwortfolgen from *BVB* via the *DNB identifier* and then work on enriching Schlagwortfolgen via Culturegraph. In Culturegraph, we can already look up manifestation bundles with the hbz ID alone so there is no real need for other identifiers. So what do you intend to gain from this?

I assume, as these identifiers are already in some of the Alma records, you generally think it will be beneficial to have them for matching purposes. Is this right?

TobiasNx:
> I am not sure I understand the need for this issue. The agreed on plan is, in the next stept to harvest Schlagwortfolgen from _BVB_ via the _DNB identifier_ and then work on enriching Schlagwortfolgen via Culturegraph. In Culturegraph, we can already look up manifestation bundles with the hbz ID alone so there is no real need for other identifiers. So what do you intend to gain from this?
> 
> I assume, as these identifiers are already in some of the Alma records, you generally think it will be beneficial to have them for matching purposes. Is this right?

I asked that Verbund N., if I should move to BVB via `dnbId` she said that this is probably useless since they would have minimal additional Schlagwortfolgen. Instead we should query via the BVB specific identifier so that we can query the BVB like this: http://bvbr.bib-bvb.de:5661/bvb01sru?version=1.1&recordSchema=marcxml&operation=searchRetrieve&query=marcxml.idn=BV044913359&maximumRecords=1

If we add the BVB identifier the DNB script only would need minimal adjustments and could harvest Schlagwortfolgen.

TobiasNx:
I added a comment to https://jira.hbz-nrw.de/browse/MDD-29: https://jira.hbz-nrw.de/browse/MDD-29?focusedId=359576&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-359576

acka47:
[...]

PR #2375, opened 2026-06-29
Add new verbund ids to index #2365

TobiasNx:
@dr0i we need these ids for https://github.com/hbz/schlagwortfolgen-harvester/issues/1 
`hebisId` was already indexed, since it was introduced by @fsteeg in context of rpb.

### DIFF
--- a/src/main/resources/alma/index-config.json
+++ b/src/main/resources/alma/index-config.json
@@ -393,6 +393,30 @@
             "q.all"
           ]
         },
+        "bvbId": {
+          "analyzer": "id_analyzer",
+          "type": "text"
+        },
+        "bzsId": {
+          "analyzer": "id_analyzer",
+          "type": "text"
+        },
+        "gbvId": {
+          "analyzer": "id_analyzer",
+          "type": "text"
+        },
+        "k10PlusId": {
+          "analyzer": "id_analyzer",
+          "type": "text"
+        },
+        "kobvId": {
+          "analyzer": "id_analyzer",
+          "type": "text"
+        },
+        "obvId": {
+          "analyzer": "id_analyzer",
+          "type": "text"
+        },
         "rpbId": {
           "analyzer": "id_analyzer",
           "type": "text"


dr0i (changes requested):
Seems odd to me, see why e.g. at https://stackoverflow.com/questions/52845088/difference-between-keyword-and-text-in-elasticsearch ("text" vs "keyword").  I mean "id" feels like "keyword".

src/main/resources/alma/index-config.json
```diff
@@ -393,6 +393,30 @@
             "q.all"
           ]
         },
```
TobiasNx, line 395: @dr0i I followed the example here.

src/main/resources/alma/index-config.json
TobiasNx, line 431: @dr0i and here.

src/main/resources/alma/index-config.json
dr0i: Ok. Let's ask @blackwinter for advice.

TobiasNx:
@dr0i  I followed the examples in the index file [1](https://github.com/hbz/lobid-resources/pull/2375#discussion_r3492031107) and [2](https://github.com/hbz/lobid-resources/pull/2375#discussion_r3492034327). Also even if it is an id, the search is enabled e.g. with `hbzId` so that we can search for `HT*` and `almaMmsId` for `*6441` Similar actions are true for `hebisId` and `rpbId` if I am not mistaken (this could be answered by @fsteeg). So the exact match query of `keyword` is not enough.

src/main/resources/alma/index-config.json
blackwinter: What's the issue? The `type: "text"`? That's because you want lowercase IDs (via `id_analyzer`). `type: "keyword"` is not analyzed.

blackwinter:
> So the exact match query of `keyword` is not enough.

`keyword`s still support prefix and wildcard queries.

src/main/resources/alma/index-config.json
dr0i: ah , I see. HT-Nr. starting lowercase `ht`. Someone uses this ? I don't mind.

dr0i:
So question boils down to: "shall these IDs be case-sensitive or not" @fsteeg ?

src/main/resources/alma/index-config.json
dr0i: thx for quickly responding @blackwinter !

blackwinter: > Someone uses this ?

We're using the ID search fields (`almaMmsId`, `hbzId`, `rpbId`, `zdbId`) in DigiBib and since it has been lowercase from the "start" (07441a1; no idea where the index config was before), I don't think it should be changed now.

dr0i (approved):

src/main/resources/alma/index-config.json
dr0i: ok then, we leave it at that

dr0i:
We decided to leave it as `text` , see https://github.com/hbz/lobid-resources/pull/2375#discussion_r3492296809.

dr0i:
Deployed. Basedump-ETL started. To be reviewed tomorrow.

Issue #1365, opened 2022-08-08
ALMA-Fix: Clean up Fix files


TobiasNx:
Delete all unnecessary morph code that is commented out.

TobiasNx:
Done by https://github.com/hbz/lobid-resources/commit/ddc7a387b47596c962265690a2e7f2e6d893f4fb

PR #2376, opened 2026-06-30
Fix `bszId` indexing #1365

TobiasNx:
`bszId`  was not indexed due to a typo in the `index-config.json`


TobiasNx:
@dr0i should I add web tests in order to avoid such errors?

dr0i (approved):

dr0i:
Tests of course are nice, but as no code (an no ETL rule) is involved it's unlikey that we will run into an issue here in the future.

PR #2380, opened 2026-07-07
Exclude online and non NZ resources from nrwLastCopies-Collection MDD-28

TobiasNx:
https://jira.hbz-nrw.de/browse/MDD-28?focusedId=357576&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-357576

>  1. Bestände der IZ Bridges ignorieren
>   2. Solingen ignorieren
>   3. Online/Elektronische Ressourcen mit physische Beständen ignorieren
> 
> Außerdem mehr statistische Daten:
> 
> Wie ist der Anteil aller LastCopies an den potentiell markierten Daten?

This PR:
- excludes bridge holdings and DE-Sol1
- excludes online resources that are not hybrid, print as well.

@acka47 not sure how we could solve the analytics part. I could add additional collection as counter part for nrwLastCopies (nrwAllCopies)?


acka47 (approved):
Looks good to me. +1

acka47:
> not sure how we could solve the analytics part. I could add additional collection as counter part for nrwLastCopies (nrwAllCopies)?

We should not solve this through the addition of more fields. Can't you just increment a counter during the TL process? Otherwise, as Metafacture is also marketed as analytics tool it should be no problem to get the numbers in a Metafacture run separate from the production ETL processes.

TobiasNx:
I could run a analytics based on the fulldump yes.  Good idea!

PR #2381, opened 2026-07-09
Link collections to lobid query #2378

### LINKED ISSUE #2378: inCollection links in web-app should link to lobid-resources search and not to the URI
Image

e.g. link to https://lobid.org/collections#epflichtnrw

TobiasNx:
Resolves #2378

### DIFF
--- a/web/app/views/tags/result_doc.scala.html
+++ b/web/app/views/tags/result_doc.scala.html
@@ -177,6 +177,28 @@
   }}
 }
 
+@idInCollectionQuery(e: JsValue) = {
+  <a title='Nach weiteren Titeln der Kollektion &#39;@((e\"label").asOpt[String].getOrElse("--"))&#39; suchen' href='@resources.routes.Application.query(collection=((e\"id").asOpt[String].getOrElse("--")))'>@((e\"label").asOpt[String].getOrElse("--"))</a>
+}
+
+@inCollection(label: String, key: String) = {
+  @if((doc\key).asOpt[Seq[JsValue]].isDefined) {
+    @defining((doc\key).asOpt[Seq[JsValue]].getOrElse(Seq(doc\key))) { elems =>
+    <tr>
+      <td>@label</td>
+      <td>@idInCollectionQuery(elems.head.asOpt[JsValue].get)</td>
+    </tr>
+      @for(elem <- elems.tail; e <- elem.asOpt[JsValue]) {
+        <tr>
+          <td></td>
+          <td>@idInCollectionQuery(e)</td>
+        </tr>
+      }
+    }
+  }
+}
+
+
 @first(seq: Seq[JsValue]) = @{
   seq(0).asOpt[String].getOrElse("--")
 }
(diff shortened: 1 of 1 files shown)


dr0i:
Deployed to staging.

web/app/views/tags/result_doc.scala.html
```diff
@@ -177,6 +177,28 @@
   }}
 }
 
+@idInCollectionQuery(e: JsValue) = {
+  <a title='Nach weiteren Titeln der Kollektion &#39;@((e\"label").asOpt[String].getOrElse("--"))&#39; suchen' href='@resources.routes.Application.query(collection=((e\"id").asOpt[String].getOrElse("--")))'>@((e\"label").asOpt[String].getOrElse("--"))</a>
+}
+
+@inCollection(label: String, key: String) = {
+  @if((doc\key).asOpt[Seq[JsValue]].isDefined) {
+    @defining((doc\key).asOpt[Seq[JsValue]].getOrElse(Seq(doc\key))) { elems =>
```
fsteeg, line 186: Since using both `isDefined` and `getOrElse` on `(doc\key).asOpt[Seq[JsValue]]` is redundant, I think you could remove the `@if((doc\key).asOpt[Seq[JsValue]].isDefined)`.

web/app/views/tags/result_doc.scala.html
```diff
@@ -177,6 +177,28 @@
...
+    <tr>
+      <td>@label</td>
+      <td>@idInCollectionQuery(elems.head.asOpt[JsValue].get)</td>
```
fsteeg, line 189: Here, you're assuming `elems` is non-empty, i.e. if the `inCollection` array exists, but is empty, this should fail. To be safe, you could wrap all the `<tr>` creation in something like `@if(!elems.isEmpty)`.

fsteeg (approved):
+1, adding two inline comments (1 cleanup, 1 possible edge case, maybe you want to apply them).

web/app/views/tags/result_doc.scala.html
TobiasNx: As result of the transformation due to[ `vacuum` ](https://github.com/hbz/lobid-resources/blob/2378-inCollectionLinks/src/main/resources/alma/alma.fix#L92) there cannot be an empty array in the data. Therefore this seems to be not unnecessary.

web/app/views/tags/result_doc.scala.html
TobiasNx: The `@inCollection` is a reuse of:

https://github.com/hbz/lobid-resources/blob/138ad2e490546d6f415c747f7a5b95f970e02b4b/web/app/views/tags/result_doc.scala.html#L16-L31

There the steps are identical. Should this be adjusted in both places?

fsteeg: Hm, looking at it again removing the line would change the behavior (the whole block currently doesn't appear if not `isDefined`), the redundancy is actually in the `getOrElse(Seq(doc\key))` (which could be `get`), and the repeated `(doc\key).asOpt[Seq[JsValue]]`. But it's fine for me to just leave it as it is.

web/app/views/tags/result_doc.scala.html
fsteeg: Yes, I see, makes sense.

PR #2391, opened 2026-07-21
Introduce workaround for non gregorian startDates #2389

### LINKED ISSUE #2389: publicationDates wrong for non Gregorian calendar entries, they should use the standardized version from 008
https://lobid.org/marcxml/99371159339706441

https://lobid.org/resources/99371159339706441.json

```
  "publication": [
    {
      "dateStatement": "1422 n. H. = 2001 or 2002",
      "startDate": "1422",
      "type": [
        "PublicationEvent"
      ]
    }
  ],
```

should use the value from 

```
<controlfield tag="008">220201|2001####ir############|||#|#ara#c</controlfield>
``

TobiasNx:
Introduce a mechanism that tests if the structured date from `008` and the date in `publication.startDate` variate more than +-5 years. If so it uses the `008` element since it seems to be plausible that the intially mapped date is not gregorian and `008` is always gregorian.

This fallback is limited to non-continuous resources but issues, items and monographs.

https://github.com/hbz/lobid-resources/pull/2391/changes/f7d2c620a5a8b8bd0f20579c0b050676642f5952#diff-e46b0b6d2ce611a31b4d1d0625169d9e5cd0ac71b72c36abb5e0a0d241c06bf7


acka47 (approved):
+1

Issue #2390, opened 2026-07-21
Non-existing pubdates in records that have manufacture


TobiasNx:
> Also 008 fallback does not always work: https://lobid.org/resources/990004221170206441 

 _Originally posted by @TobiasNx in [#2389](https://github.com/hbz/lobid-resources/issues/2389#issuecomment-5034222725)_

This seems to be due to records that have no `publication`-property BUT have a `manufacture` property that was mapped from https://www.loc.gov/marc/bibliographic/bd264.html
`264 Ind2=3`

Unfortunately `manufacture` is not indexed currently.

https://lobid.org/resources/search?q=inCollection.id%3A%22https%3A%2F%2Flobid.org%2Fcollections%23nrwLastCopies%22+AND+NOT+_exists_%3Apublication.startDate+AND+NOT+%28inCollection.id%3A%22http%3A%2F%2Flobid.org%2Forganisations%2FDE-655%23%21%22+AND+publication.startDate%3A%5B1851+TO+2016%5D+AND+NOT+medium.id%3A%22http%3A%2F%2Frdaregistry.info%2FtermList%2FRDACarrierType%2F1018%22+AND+%28type%3ABook+OR+type%3AMap+OR+type%3APublishedScore%29+AND+bibliographicLevel.id%3A%28%22https%3A%2F%2Fwww.loc.gov%2Fmarc%2Fbibliographic%2Fbdleader.html%23Monograph_Item%22+OR+%22https%3A%2F%2Fwww.loc.gov%2Fmarc%2Fbibliographic%2Fbdleader.html%23Serial_component_part%22%29%29

TobiasNx:
To solve this issue we could add a new `publication`-element with only the `startDate`-property mapped from `008`

acka47:
As discussed today: 

* [ ] show `manufacture` in the GUI, see related NWBib issue https://github.com/hbz/nwbib/issues/627
* [x] index `manufacture`

TobiasNx:
The indexing of manufacture created some MapperParsingExceptions with regard to the pub dates:


```
2026-08-13 02:49:29 [WARN] org.lobid.resources.ElasticsearchIndexer in Thread-44 - Bulk insert failed: failure in bulk execution:
[...]

PR #2394, opened 2026-08-18
Fix manufacture dates that break indexing #2390

TobiasNx:
The mentioned issue with `manufacture` dates that breaks the index should be solved with this:
https://github.com/hbz/lobid-resources/issues/2390#issuecomment-5327245714

Related  #2390

### DIFF
--- a/src/main/resources/alma/fix/titleRelatedFields.fix
+++ b/src/main/resources/alma/fix/titleRelatedFields.fix
@@ -423,9 +423,13 @@ end
 # Clean manufacture
 
 do list(path: "manufacture[]","var":"$i")
+  replace_all("$i.startDate", ".*?([01]\\d{3}|20\\d{2}).*", "$1")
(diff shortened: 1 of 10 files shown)


acka47 (approved):
+1 To understand this correctly: The mapper parser exception appeared because of the uncleaned dates (e.g. `"startDate" : "© 2016"`) in `manufacture`, right?

TobiasNx:
Correct. ElasticSearch excpected `YYYY` for `startDate` and `endDate` in oder to allow sorting etc. If the record does not provide this, the single record is skipped and not included in the index.

dr0i (approved):
+1 principially. I just wonder if adding one new test resource wouldn't be enough?

TobiasNx:
The tests represent different versions of the malformed dates (`1961?`, `1903-1910`, `© 2016.` , `Anno domni. M..lxxxix. Finitus in die santi Leonardi. 1489.11.06.`) so that the fix works for different scenarios. A single test would only ensure that this particular scenario is fixed. The four tests help with an more encompassing approach.

PR #2401, opened 2026-09-03
Switch to ssh when pulling from GitHub

dr0i:
Pulling via HTTPS sometimes fails like this:
> fatal: could not read Username for 'https://github.com': No such device or address

resulting in trouble of restarting, see oi/lobid-resources-documentation/issues/9.

To (hopefully) [avoid this we use SSH deploy keys at GitHub](https://github.com/orgs/community/discussions/206581#discussioncomment-18269083). Check with "git remote -v" on your server if it's properly configured.

### DIFF
--- a/web/monit_restart.sh
+++ b/web/monit_restart.sh
@@ -36,7 +36,7 @@ ETL_TOKEN=$(cat scripts/.secrets/ETL_TOKEN)
 case $ACTION in
   start)
        cd ..
-       git fetch; git reset --hard origin/master; git submodule update --init --recursive --remote;
+       git fetch; git reset --hard ssh/master; git submodule update --init --recursive --remote;
        if [ ! -f lookup-tables/data/opacLinks/isil2opac_issn.tsv ]; then
           exit # see #2306
        fi
--- a/web/monit_restart.sh
+++ b/web/monit_restart.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -euo pipefail # See http://redsymbol.net/articles/unofficial-bash-strict-mode/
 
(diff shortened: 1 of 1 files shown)


TobiasNx:
Not sure how I should review this.

dr0i:
@TobiasNx You could execute the script being in the office and see if it runs ok.

Phu2:
I noticed two other things:

- there is no check if `RUNNING_PID` exists, i assume that the script will not work as expected if it is missing
- `git reset --hard` is executed and will delete local edits, not sure if that's desirable in any case

You may want to address these (in another issue)?

Phu2 (approved):

dr0i:
> there is no check if RUNNING_PID exists, i assume that the script will not work as expected if it is missing

uh, right: with the new "bailing out" mechanism further steps of coping with possible issues are no more reachable by the code. E.g. if a RUNNING_PID somehow was not updated and thus cannot be killed (as the PID no more exists) the force-removing of the RUNNING_PID will not be called. And thus the script will consecutively fail. 
Resolved in cb3ca3d266eb2a1b6e9cc738e3aa04bc029cd1b9.

> git reset --hard is executed and will delete local edits, not sure if that's desirable in any case

Admittedly, it has one ugly sideeffect - when testing the script one has to comment out this line first.
If you insist on changing this (analog to $todoUrlToForgejoIssueIcouldntFindSomethingWithRpb) we should do this in an own issue.

This script here was copied to e.g. https://developer.hbz-nrw.de/oi/rpb/src/branch/main/monit_restart.sh. We also may want to update these.

dr0i:
I leave you assigned here @TobiasNx so you can check it in office.
I've already merged it so we won't run into any problem GitHub may have.

